PR #4, opened 2017-01-06
Correct the bit lengths of various IR protocols

crankyoldgit:
Disable Panasonic codes as it is called wrong.
Store data in an unsigned variable, vs a signed on, otherwise it produces bad results for high numbered codes.
Mark protocols that are > 32 bits as they won't work with values large than an unsigned long (32 bits).


crankyoldgit:
The protocols left at '36' are in fact protocols with 36bit datas. The ones I changed to 32bits or less are the correct sizes for those protocols. You could nobble the 36bit protocols down to 32bits but either way, the results won't be fully correct.

You probably want to parse an unsigned long long (64bit) for your value obtained by mqtt. You could then also split up that value for the correct invocation of the panasonic procedure.

Personally, I'm only using the NEC & Sony protocols, which I wanted fixed.

PR #8, opened 2017-01-09
Use strtoul() instead of custom code.

crankyoldgit:
A standard function exists to convert c-style strings to unsigned longs.
https://linux.die.net/man/3/strtoul


1technophile:
Thanks for the proposal

PR #12, opened 2017-01-20
Control a fan via wiring gpio pins to remote

rickybrent:
I'm not sure this is worth approving, but I had a fun time hacking together support for controlling a very-much non-433mhz fan remote control kit I purchased. :) 

Thanks for putting this project online with such easy instructions; I started using it last week for my various power outlets (byebye standby and of course the Etekcity ones everyone uses) and it worked beautifully!


1technophile:
can be usefull for people wishing to have a device controled by the GPIO, I merge it and will put it on a separate folder, thanks for participating

PR #58, opened 2017-06-07
Addon lightsensor

HannesDi:
i hope all of you understand my bad english ;-)

sorry for the chaos and unstructured work but this is the first contribution
for a project in my life... i started 3 month ago with the arduino stuff
and have no clue i have understand the git system correct how to merge code..

So here we go....

https://github.com/HannesDi/OpenMQTTGateway/tree/addon-lightsensor

I have opened the Door on OpenMQTTGateway for "i2c Bus Modules/Sensors"
but this has required to complete restructure the Pin Connections
on the NodeMCU

Why?
2 Reasons

First:
As most of us know and and 1technophile has allready written
we run out of usable pins for different boards.
So i have made some research and found out that many pins
could switched its function...
like d0 - gpio16 and d5 - gpio14
d0 could not be used for receive because it has no interrupt as i have understand
so why not use it for Sending IR ? ;-)
the d5 is free and usable for other stuff.

Second:
To Use I2C there a 2 possible ways...

complete re-invent the i2c software for the Board
use the Wire.h that exists and work great...
But Wire.h has a small problem.. it explicit wants D1 and D2 on the ESP8266
and so i have freeded and switched the other pins..
Ended up in this:

D0 - GPIO16 # as Send Pin for IR (because IR need no interrupt)
D1 - GPIO5 # as SCL (I2C Bus) for BH1750 Light Sensor
D2 - GPIO4 # as SDA (I2C Bus) for BH1750 Light Sensor
D3 - GPIO00 # as Receive Pin for 433mhz
D5 - GPIO14 # as Receive Pin for DHT22
D8 - GPIO15 # as Send Pin for 433mhz

And all is Working ;-)

There is a another benefit for using this...
I await the next 2 Days a another i2c bus module the "BME280" a "I2C or SPI Temperature, Humidity and Pressure Sensor"

for all they dont know what this means...
the I2C bus can handle up to 8 modules on 2 wires in parallel (no more worries about free pins)
and... for thoose they have allready a DHT11 or DHT22 you can unplug it then and
we have again pin D5 free for other things ;-)

i hope you an first of all 1technophile like my work

Best Regards Synologix / HannesDi


1technophile:
I have updated the master .travis.yml added DHT libraries could you include these changes into your branch (not able to do it as this file is not part of the modified ones in your branch).
I took a look to the code and it is OK. Please remove the commented traces using serial.print Thanks for the work!

Regarding the pin do you have explored all the possibilities? (SD2, SD3,SD1 or others) I had some bad experiences with D0 and if we change all the pins setup we will impact all the users.

HannesDi:
Light Sensor Addon is Obsolete now because replaced by I2C Modules Addon 
I start a new Pull Request-> So... Close

PR #61, opened 2017-06-10
I2C Sensors (Replaces my Obsolete Lightsensor Module)

HannesDi:
So after i have finished the 2nd I2C Module in this Case a Multisensor for Temp, Humidity, Pressure and Altitude, my old Contrib Light Sensor is Obsolete.

This Contrib brings both: Light Sensor and Weather Sensor.

**1technophile / Florian has written last:**
_I have updated the master .travis.yml added DHT libraries could you include these changes into your branch (not able to do it as this file is not part of the modified ones in your branch).
I took a look to the code and it is OK. Please remove the commented traces using serial.print Thanks for the work!

Regarding the pin do you have explored all the possibilities? (SD2, SD3,SD1 or others) I had some bad experiences with D0 and if we change all the pins setup we will impact all the users._

- mofified .travis.yml
- serial.print is removed

About your Question for **SD2, SD3,SD1 or others**, yes i have explored all of them on my ESP8266
i have explored a long time to find a solution and endet up with this as working concept.

We have discussed this 1 or 2 weeks ago (there was my name synologix) talked about the hanging ESP8266 when the DHT is connected on Powerup etc....

I understand that you have some bad experiences with D0 (you are not alone) but part of my research included whats the Reason for the many user Complaints about D0 is.
D0 has very often Problems as INPUT / Reading Pin. So i used it for IR Output and this looks like working good.

And YES you are right with changeing all this Pins we impact all users they have allready soldered a PCB.... For new Users it makes no difference. Nevertheless i have thinked about this tonight and have created a (from my point) clever solution: 
/*-------------DEFINE YOUR WIRING TYPE BELOW----------------*/
// Choose between "I2C_Wiring" OR "Classic_Wiring"

you can see this in code review.... i think this could make all happy ;-)

But if everything else fails... I think when users using the I2C-Bus they use the way better Bosch BME280 instead the DHT11 or DHT22 and so... we can Regain the D5 (no more used for the DHT) and Plug the IR Send Pin from D0 back to D5

Also i have moved the DHT Pin config from zsensor to userconfig
DHT dht(DHT_RECEIVER_PIN,DHT22); // Defined in User_config.h

All other ideas you can find by reviewing and hopefully merging the project

Best Regards
Hannes (Synologix)


1technophile:
Hello Hans-Juergen,

Here is the only issue found :-):
regarding ZsensorBH1750 line 108
```
    if (isnan(Lux)) {
      trc(F("Failed to read from LightSensor BH1750!"));
```

The program doesn't seem to pass here even if there is no BH1750 connected (my case) in consequence it publish 0 values to MQTT:
```
home/LIGHTtoMQTT/lux 0
home/LIGHTtoMQTT/ftcd 0.00
home/LIGHTtoMQTT/wattsm2 0.0000 
```

In my point of view we should let classic wiring per default (as long as the documentation and schematic is not updated) and comment all the sensors define, most of the time people begin to use the gateway with RF and/or IR maybe BLE I don't wish to make it more complex per default . What do you think?
We could add a dedicated wiki section dedicated to the sensors and how activate them.

HannesDi:
Module detection is fixed now ;-)
classic for default is no problem. 
a another idea from me was, due it is needed to use the i2c wiring when user want to use the BH1750 or BME280 you can autoswitch to I2C Wiring when one or both modules are enabled.

something like:

#ifdef ZsensorBH1750 or/and  #ifdef ZsensorBME280  
  #define DHT_RECEIVER_PIN 14 //on nodeMCU this is D5 GPIO14
  #define IR_RECEIVER_PIN 2 // put 2 = D4 on nodemcu, 2 = D2 on arduino
.
.
else
  #define DHT_RECEIVER_PIN 0 //on nodeMCU this is D3 GPIO0
  #define IR_RECEIVER_PIN 2 // put 2 = D4 on nodemcu, 2 = D2 on arduino

and give a hint to the user...
// #define ZsensorBH1750 // Read notes in FAQ
// #define ZsensorBME280 // Read notes in FAQ
// Please Note: 
// If you want activate/use this Modules it is required to change the Wiring
// More Informations about this you will find under: https://github.com/1technophile/OpenMQTTGateway/wiki/Using-I2C-Modules

1technophile:
THanks for this PR, as I have added the modification proposed a new PR has been created based on this one . It is now integrated. I close this one in consequence

PR #78, opened 2017-07-29
Formatting and spelling suggestions

Landrash:
Some minor spelling errors.
Some formatting errors where capitalisation was a bit of.
Reordered some of the information.
Removed example at the end.
Added supported devices to first list from second list.

I'll submit one with only the formatting and spelling if you prefer the original layout.


1technophile:
No it is ok like that thanks  !

Issue #95, opened 2017-08-26
Analog pin (A0 on Nodemcu) support


PetricaM:
Hi,

Could you add support for connecting analog sensors on the A0 pin on the Nodemcu?

An analog sensor would read the value of the measured variable (luminescence, magnetic field, sound, etc) and publish it to an MQTT topic as a number between 0 and 1023. 

This could be published immediately as value changed or/and do a new read after certain number of seconds.

The home automation software would then read it and interpret it in specific way: for LDR sensors it can be used to create automation for lights or for a magnetic hall sensor as an open door sensor which sends both open and close states.

Thank you

PR #106, opened 2017-09-13
ADC pin reading

1technophile:
#95


1technophile:
@petricam ADC reading is available in this branch, if you have time I would be interested by your feedback.

PetricaM:
Hi

I'll get home during weekend and will try it.

Thanks

1technophile:
@PetricaM threshold added

PR #113, opened 2017-09-20
Makes RF2 sending and receiving work for ESP8266 with added support for group bit and dimming

larsenglund:
With code from https://github.com/1technophile/OpenMQTTGateway/pull/109
plus some other fixes.


larsenglund:
Just noticed that to make this work with Home Assistant the state of the switch has to be published via MQTT (RF2toMQTT) when a MQTTtoRF2 command is received..

larsenglund:
I also just noticed that the group bit is not supported, I'll probably add support for that..

larsenglund:
Another big problem is that the period varies on received codes so the topic to match on in Home Assistant will vary..

ZgatewayRF2.ino
```diff
@@ -117,8 +113,8 @@ void MQTTtoRF2(char * topicOri, char * datacallback) {
   
   int pos = topic.lastIndexOf(RF2codeKey);       
   if (pos != -1){
-    pos = pos + strlen(RF2codeKey);
-    valueCODE = (topic.substring(pos, topic.indexOf("/", pos))).toInt();
+    pos = pos + +strlen(RF2codeKey);
+    valueCODE = (topic.substring(pos,pos + 8)).toInt();
```
larsenglund: I see you added back the hard coded code length, this will make it not work with some remotes.

1technophile: indeed merge mistake, thanks for pointing this

larsenglund: Also my support for the group bit and the MQTT state updating (when receiving RF commands from MQTT) didn't make it through the merge..

1technophile: ok :-(, may I ask you to repush your modifications into this branch, some remarks
just replace *Serial.println(* by *trc(*
remove *int valueBIT  = 0;*
and comment ZgatewayRF2 for arduino as the sketch is too big for Uno

larsenglund: Sure, what would be the easiest way to go about going that, forking the current master branch and re-apply my changes and then make a new pull request? I'm still a newbie with git branching and pull requests.. I have also implemented support for dimming (verified to work with my Nexa WMR-250 dimmer)

1technophile: *forking the current master branch and re-apply my changes and then make a new pull request?*
I think it will be the easiest way
And we cancel this one, sorry for the rework...

larsenglund: Sure, no problem

1technophile:
Thanks!

PR #114, opened 2017-09-22
Moved dim level to MQTT payload for RF2

larsenglund:
Moved dim level from topic to payload to make dimming a bit more consistent with switching and to integrate better with Home Assistant.


1technophile:
thanks

PR #125, opened 2017-10-01
Show ip and mac after connect

ChiefZ:
When not using advanced network config, the ip is optained via DHCP, so display it after the connection is established. Output was 0.0.0.0 in this case before.
Also adding display of the mac (before connection) for mac address filter config in the wifi-ap.


1technophile:
Tested ok, thanks

```
NEUF_7AE0
OpenMQTTGateway mac: 
34:01:SZ:0B:CS:77
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
OpenMQTTGateway ip: 
192.168.1.51
WiFi ok
```

PR #147, opened 2017-12-03
Add support for presenting status with 3 LEDs

8666:
The idea is taken from https://www.mysensors.org/build/advanced_gateway

It is working for RF433 and IR module. Probably can be added to the other modules. RED will stay ON until MQTT connection is ready.

### DIFF
--- a/OpenMQTTGateway.ino
+++ b/OpenMQTTGateway.ino
@@ -80,6 +80,11 @@ PubSubClient client(mqtt_server, mqtt_port, callback, eClient);
 //MQTT last attemps reconnection date
 unsigned long lastReconnectAttempt = 0;
 
+//timers for LED indicators
+unsigned long timerG = 0;
+unsigned long timerY = 0;
+unsigned long timerR = 0;
+
 boolean reconnect() {
   // Loop until we're reconnected
   while (!client.connected()) {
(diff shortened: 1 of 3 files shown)


Landrash:
Some feedback. I haven't tried your code yet but just from a quick read trough.

Assuming a LEDs color is bad since that would assume that only that color could be used. Adding in in the documentation that the following colors have been used for testing is a decent workaround.
Instead of toggling a specific pin I would call a function that does that. 

Removing color and specific pins from the function makes it more scale-able if you want to separate function LEDs for different radios etc..

1technophile:
To have a valid continuous integration test you need to do the modification of user_config.h into 
https://github.com/1technophile/OpenMQTTGateway/blob/master/tests/Test_config.h also.
Indeed for the continuous integration user_config.h is replaced by test_config.h. 
test_config.h as much more gateway enabled for the test thant the default user_config.h

In your modification case we are talking about ledG... definitions that are missing into test_config.h

8666:
Looks fine now :)

1technophile:
Hello,

Thanks for the modifications, for my knowledge which board are you using ?

8666:
Mega 2560. The pins I used are on the bottom side (easy for wiring). There is wiring "diagram" included in the comment after decalrations

1technophile:
A last comment, we need to manage the multi board ability of OpenMQTTGateway, maybe we could add an 
```
#ifndef ESP8266
your pins definition
#endif
```
In this case we manage the arduino boards = if not def ESP8266
Due to the fact that esp8266 is lacking of pins, do you think you can find common pins compatible with arduino Uno and Mega and Maybe ESP32?

8666:
Well, 3 pins are needed. I do not know how many are free for each board...

Maybe you can run a configuration with two pins if you use the same LED for RX/TX and one for Errors. Or even single LED, if it is always on then it is an error, if it is on for short time then you have communication. You just define the same pin number.

What needs to be handled is when someone does not have one or all of the pins defined...then the LED commands should be skipped.

PR #150, opened 2017-12-28
Add support for TSL2561 based light intensity measurement

broekema:
I added light intensity measurement based on TSL2561 based modules, such as these from banggood: https://www.banggood.com/GY-2561-TSL2561-Luminosity-Sensor-Breakout-Infrared-Light-Sensor-Module-Integrating-Sensor-AL-p-1129550.html

Dependencies are on Adafruit_TSL2561 and Adafruit_Sensor, as stated in header.


broekema:
It seems that the Travis build fails on the Adafruit_TSL2561 dependency. I don't quite know how to fix that. Suggestions are welcome?

PR #151, opened 2017-12-30
globally expose repeat functionality as advanced function

broekema:
My RC5 based remote requires data to be send 3 times. The functionality for this is mostly implemented, so this is quite a small patch. Reading the code, I see no reason not to introduce the functionality for all remote types (with the exception of raw and GC), considering the support is available in the underlying library. If not set, this defaults to 0 (which sends the data once).

### DIFF
--- a/ZgatewayIR.ino
+++ b/ZgatewayIR.ino
@@ -198,27 +198,27 @@ void MQTTtoIR(char * topicOri, char * datacallback) {
   #ifdef IR_COOLIX
   if (strstr(topicOri, "IR_COOLIX") != NULL){
     if (valueBITS == 0) valueBITS = 24;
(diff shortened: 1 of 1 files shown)


broekema:
... available in the underlying library for ESP8266 apparently. Let me investigate a bit.

broekema:
I've committed an Arduino based implementation. This fixes what I believe are two bugs in the current mainline as well. The Arduino IR remote library only accepts two arguments for irsend.*(), therefore lines 229 and 265 in ZgatewayIR.ino must fail for non-ESP8266 builds. The tests pass these because all IR variants are commented out for Arduino by default.

1technophile:
Thanks for the PR and for pointing/correcting the arduino limitation

PR #154, opened 2018-01-09
Better sensors

broekema:
This branch has two goals and one bugfix:
- I noticed in on ESP32 that BME280 sensor readings quite often contained tailing garbage. This is due to several issues with buffer length in the various sensor code files. While incorrect on all platforms, this showed up on ESP32, probably due to a more modern base compiler. This is a classic C-style issue, which I've solved by moving to C++-style dynamically allocated strings. This may use more memory, so perhaps less suitable for arduino boards.
- There were also occasional spikes in the BME280 data. I've attempted to mitigate these by tweaking some of the sensor internal parameters (turned on the filter and significantly increase tStandby). At first glance the results are better.
- Finally, ZsensorADC did not compile on ESP32 for me on line 31. I believe this is ESP8266 specific, but I could be wrong.

### DIFF
--- a/OpenMQTTGateway.ino
+++ b/OpenMQTTGateway.ino
@@ -240,6 +240,9 @@ void setup()
   #ifdef ZsensorBH1750
     setupZsensorBH1750();
   #endif
+  #ifdef ZsensorTSL2561
+    setupZsensorTSL2561();
+  #endif
   #ifdef ZgatewayIR
     setupIR();
   #endif
@@ -371,6 +374,9 @@ void loop()
     #endif
     #ifdef ZsensorBH1750
       MeasureLightIntensity(); //Addon to measure Light Intensity with a BH1750
+    #endif
+    #ifdef ZsensorTSL2561
+      MeasureLightIntensityTSL2561();
     #endif
     #ifdef ZsensorDHT
       MeasureTempAndHum(); //Addon to measure the temperature with a DHT
--- a/ZsensorADC.ino
+++ b/ZsensorADC.ino
@@ -28,7 +28,7 @@
 */
 #ifdef ZsensorADC
 
-#if defined(ESP8266) || defined(ESP32)
+#if defined(ESP8266)
   ADC_MODE(ADC_TOUT);
 #endif
 
@@ -47,11 +47,9 @@ void MeasureADC(){
       trc(F("Failed to read from ADC !"));
     }else{
       if(val  >= persistedadc + ThresholdReadingADC || val  <= persistedadc - ThresholdReadingADC){
-        char value[4];
-        sprintf(value, "%d", val);
         trc(F("Sending analog value to MQTT"));
         trc(String(val));
-        client.publish(ADC,value);
+        client.publish(ADC,String(val).c_str());
         persistedadc = val;
        }
     }
--- a/ZsensorBH1750.ino
+++ b/ZsensorBH1750.ino
@@ -104,7 +104,7 @@ void MeasureLightIntensity()
       // Generate FtCd
       if(FtCd != persistedlf || bh1750_always){
         trc(F("Sending FtCd to MQTT"));
-        trc(String(ftcd));
+        trc(String(FtCd));
         client.publish(FTCD,String(FtCd).c_str());
       }else{
         trc(F("Same ftcd don't send it"));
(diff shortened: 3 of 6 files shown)


1technophile:
Thanks for proposing, my only concern is about the use of string for arduino UNO that is still supported.
Do we don't have a way to correct the issue without using String?

broekema:
Compare the travis build for Uno on my previous pull request:
Sketch uses 27530 bytes (85%) of program storage space. Maximum is 32256 bytes.
with the current one:
Sketch uses 28216 bytes (87%) of program storage space. Maximum is 32256 bytes.
I think that memory wise we'll be fine.

1technophile:
ZSensors are not tested during travis build. But if you don't find other solution to avoid string we will merge it like that.

broekema:
I don't think there is an easy way to do this with less use of memory that is guaranteed to work correctly. For information I just did a build for uno with four sensors enabled (DHT, BME, BH and TSL, everything else disabled) which resulted in slightly more memory used in the BetterSensors branch vs master:

- master : 31538 bytes (and 93% sram)
- BetterSensors: 32254 bytes (and 105% of sram)

I noticed that the TSL2561 code was missing, so I re-added it.

1technophile:
@broekema thanks for this test.
I understand that we get with these different sensors +12% of ram usage.

My level on C is not enough high on this kind of thing, I propose to ask to other people if they have ideas on how to correct this and try to keep the ram usage as low as possible. Are you ok with that?

broekema:
That's fine.

1technophile:
@crankyoldgit, @rickybrent, @ChiefZ 
Hello,
Could you help us on this subject; we would like to know your opinion about this modification:
https://github.com/1technophile/OpenMQTTGateway/pull/154/commits/8a1f7ed896eaa228848fb3ca19373c4cf214ae03

In term of code it looks cleaner to use string but it results in more memory use which can be an issue on the Uno. Do you have some advises to give about how to correct the point of @broekema ; "While incorrect on all platforms, this showed up on ESP32, probably due to a more modern base compiler."

crankyoldgit:
@1technophile In short, you've been lucky it's only been showing up on the ESP32. It's not a compiler issue per se that I can see. It could be the libraries for the ESP32 are pushing the output strings over the limit by more precision or something, that may have tickled the _bug(s)_ sooner, but the main issue is the existing poor string handling hygiene in the projects code, not the libraries or compilers fault.

If you can afford it, C++ strings are the way to go, but it sounds like you can't afford it _and_ maintain Uno support.

Good luck.

crankyoldgit:
Oh, just in case you didn't know how C's [null-terminated](https://en.wikipedia.org/wiki/Null-terminated_string) strings [work](http://www.cs.ecu.edu/karl/2530/spr17/Notes/C/String/nullterm.html).

1technophile:
@crankyoldgit thanks for all these infos !

broekema:
@crankyoldgit very well written review, thanks. I completely agree with your assessment that C++ style strings are the way to go.

In my opinion, while we can fix the immediate problems as suggested above by increasing the buffer sizes according to data sheet information, that is still error prone and is not guaranteed to be correct. Using snprintf instead of sprintf is a huge improvement. I prefer my original solution though, even if it does come at the cost of less Uno compatibility.

1technophile:
It is a good thread for having some inputs about C char array manipulations, thanks again @crankyoldgit 

If I sum up the discussion I see debate between these 2 choices:
- Use C ++ style strings for sensors
Less error prone compared to C char array
More memory needed
Easier to handle

- Use C functions like snprintf for sensors
Less memory needed preferable for Uno
Risk of errors due to fixed buffer size allocation
Speed is better compared to C++ style String

Added to that we can think about expanding the choice that will be done here to the whole project. But nevertheless I propose another approach which will enable to keep Uno compatibility at least for some base gateways. Why wanting to keep Uno compatibility; because Uno is most of the time the entry point for microcontrollers tinkering, and a lot of people have an Uno laying around.

What can be done is to use C++ strings for sensors and to optimize by using C style string for gateways as much as we can. To keep OMG introducable on an Uno. I will use the advices of @crankyoldgit for that.
Another possibility would be to differentiate by #define Uno code to other board codes but it would complexify the code much more. 

I'm going to merge this PR and [open an issue](https://github.com/1technophile/OpenMQTTGateway/issues/164) for the string handling improvment on the whole project

PR #168, opened 2018-01-28
Sensor classes

broekema:
This is an example for discussion based on the discussion on the better sensors pull request. 

The basis of this idea is to refactor the sensor code using a single Zsensor base class. All implementations extend this, but the MQTT publishing code is implemented in the base class. I've added both a base class and a new sensor (INA219). The base class currently implements 4 publish() methods for various types of data, but that can easily be extended.

Note that I've used C++ style defines to avoid multiple inclusion, and use those same defines to determine whether to run the sensor code. You now only need to include 'ZsensorINA219.h'. A ZsensorINA219 object will need to be created (line 103).


1technophile:
Thanks for proposing this, it's a new way of seeing sensor implementation that can be interesting also for gateways.
Could you add the library needed and the sensor into test_config.h to validate the compilation.

PR #203, opened 2018-05-10
Add Hostname to Wifi Connection

intractve:
Utilize ota_hostname variable from User_config.h to generate hostname for friendlier display on router's connected devices instead of ESP_xxxxxx  

Add hostname setting to both Manual and WifiManager setups


intractve:
The modifications are inside the if statements for ESP8266 or ESP32 only. Not sure why it's affecting Uno etc. in Travis CI

intractve:
Closing this PR to come up with a better more inclusive solution that passes all tests

Issue #228, opened 2018-07-08
Wiki error with the Esp IR schema 


xelfe:
It's about the Esp IR schema that we can see into the wiki [here](https://github.com/1technophile/OpenMQTTGateway/wiki/ESP8266-IR-Send-and-Receive).
The schema illustrate a TSOP38238 or a 1828T IR receiver. Both have ground on the middle leg of the receiver. The ground and the vcc have to be switch.
![Esp IR schema](https://raw.githubusercontent.com/1technophile/OpenMQTTGateway/master/img/OpenMQTTgateway_ESP8266_Addon_IR.png)

1technophile:
Thanks for pointing that, if you are interested don't hesitate to modify the [fritzing file ](https://github.com/1technophile/OpenMQTTGateway/blob/master/img/OpenMQTTgateway_ESP8266_Addon_IR.fzz) and submit a pull request.

PR #230, opened 2018-07-10
Fix for ESP8266 Ir schematic - issue #228

xelfe:
Fixing wiki issue #228


1technophile:
Thanks, would it be possible to update the [png file ](https://github.com/1technophile/OpenMQTTGateway/blob/master/img/OpenMQTTgateway_ESP8266_Addon_IR.png)also ?

xelfe:
@1technophile Of course ! By the way I found  a couple of other errors into the wiki. Should I create a issue before each pull request ?

1technophile:
Hello, No needs of an issue when there is a pull request. 
The goal is to track modifications attached to a release and a milestone, either a PR either an issue answer the need.
Thanks for the png !

PR #258, opened 2018-08-23
GPIOInput Third Time is a Charm!

QuagmireMan:
This will allow the user to see a closed circuit to ground.  This could be used for a button press, limit switch, float switch etc.

Hopefully I put the PR together correctly this time.  If I am ever in France I owe you a beer because of all the headache I put you through lol.


1technophile:
LGTM! Great and thanks!

PR #264, opened 2018-09-07
add MQTT publishing of service data from BLE devices

1technophile:
The goal is to enable users to see service data from devices not still integrated to OMG and propose the integrations by analysing servicedata seen on the MQTT broker


poroping:
@1technophile Thanks for implementing this, I've been running this on my esp32 and it's working fine. Will let you know if I run into any issues.

PR #265, opened 2018-09-08
Added decode for battery level and combined temperature/humidity

f-reiling:
Mi Jia temperature sensor sends combined message for temperature and humidity. With this modification it is possible to decode also this message type. Also a decoder for the battery status is added.


1technophile:
Thanks!

PR #266, opened 2018-09-08
Added actuator for RGB-LEDs via FASTLED lib.

joen:
This is all a litte dirty, but working.
You could theoretically use all RGB-LEDs, that are supported by fastled. Just change the configuration according to config_FASTLED.h

I am not quite happy with the effects, which could be done in a more intelligent manner.
The overall idea is to have a color value send to a topic that is used as the color the effect is "played" in and as the color that is set, when the effect is finished.


joen:
So i should fork the current version, integrate my changes an then issue a new pull request, right?

1technophile:
Yes it would be better

1technophile:
Hi @joen, do you think you would be able to do a new pull request ?

ZactuatorFASTLED.ino
```diff
@@ -0,0 +1,141 @@
...
+      leds[i] = number;
+    }
+    FastLED.show();
+  } else if (topic == subjectMQTTtoFASTLEDscan){
+    for (int j = 10; j > 0; j--) {
+      for (int i = 0 ; i < FASTLED_NUM_LEDS; i++ ) {
+        leds[i] = number;
+        leds[i - 1] = CRGB::Black;
+        FastLED.show();
+        delay(100);
```
Legion2: using delay will block the main loop.

madl3x:
How is this going? Any chance we get to see it released in 0.9.x?
Thanks

1technophile:
v0.9.2 I don't think, could be the next if you are interested in testing it?

BigDi:
I have a similar use case. I take the code from Joen and merge it into current dev branch. It will compile but I haven't tested it on my WEMOS.

1technophile:
This is great! Could you maybe submit a new updated PR?

BigDi:
I created a new pull request. Hope every thing is corrent. Is my first one.
I have some trouble while testing the code with my WEMOS D1 mini. I get it only working if I used pin D2. Other pins are not working.

joen:
closing, please refer to https://github.com/1technophile/OpenMQTTGateway/pull/433

PR #278, opened 2018-10-01
multi gpio input module

McGr3g0r:
This module uses up to 4 gpio pins and creates keycode from bits combination, there is 5th gpio pin used as latch signal - for example it can be used with ev1527 code receiver.

My OpenMQTTGateway in 3d printed case:

https://photos.app.goo.gl/unCdrHzDNx1J68cm6


1technophile:
@McGr3g0r Hello, thanks for this PR,
 would you be interested in adding a wiki page describing your setup and  another one about how to use it?

PR #298, opened 2018-11-02
Publish refactoring

1technophile:


1technophile:
factoring of the publishing function
integration of json publishing to MQTT
receiving and transcode json messages from mqtt
some corrections of conversion

PR #299, opened 2018-11-03
Flexible On Off, multiples pins, delays.. mosquitto_pub -t home/OpenM…

SenexCrenshaw:
Flexible On Off
multiples pins
delays

mosquitto_pub -t home/OpenMQTTGateway/commands/MQTTtoONOFF -m 2000,4<,publish after delay> 

Will turn pin 4 on for 2000 milliseconds. by default after the delay no mqtt message will be published unless set to 1

This is helpful if you only need the pin to go high for x milliseconds (fireplace control...) and then stop. Sending a pub after the delay is configurable. I want the fireplace switch to stay on high (on) even though the pin isnt. Works great with Home assistant.

I usually dont publish PRs so sorry about the extra formatting.

### DIFF
--- a/OpenMQTTGateway.ino
+++ b/OpenMQTTGateway.ino
@@ -5,11 +5,11 @@
    Send and receiving command by MQTT
 
   This program enables to:
- - receive MQTT data from a topic and send RF 433Mhz signal corresponding to the received MQTT data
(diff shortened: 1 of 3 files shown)


1technophile:
Closing the PR, I will retinegrate it from developement branch

PR #305, opened 2018-11-09
change RCSwitch to fork with more Protocols and better ESP32 support

animavitis:
hope this is better :)


1technophile:
As the RF gateway is the most used gateway of OMG how do you think we can assure other users this change will not generate regressions on their side?

animavitis:
To be honest we cant assure.
ive tested this repo with my esp32 & esp8266 devices and it works fine & stable (tested only with RF). 
Other things we can do is:
- static code analysis (Im not good enough developer , so it looks like it will be your job :) )
- deploy changes as next beta version and ask users for feedback.

animavitis:
i have mess in my repo, please remove this RP, i will create clean fork & recreate this RP (and will do same for mqtt discovery RP)

PR #311, opened 2018-11-14
Mqtt discovery

animavitis:
first version of home assistant mqtt discovery, it have omg status sensor & sensors for BME280


animavitis:
have messy master in my repo, will recreate fork and then create PR once again

PR #314, opened 2018-11-15
add mqtt discovery (compatibile with home assistant) with omg status …

animavitis:
- added omg status binary sensor
- added bme280 sensors
- anothers sensor can be easy added :)
- OMG device is visible in home assistant integrations  
- All discovered sensors are connected to OMG gateway in integrations

### DIFF
--- a/OpenMQTTGateway.ino
+++ b/OpenMQTTGateway.ino
@@ -124,6 +124,9 @@
 #ifdef ZsensorGPIOKeyCode
   #include "config_GPIOKeyCode.h"
 #endif
+#ifdef ZmqttDiscovery
+  #include "config_mqttDiscovery.h"
+#endif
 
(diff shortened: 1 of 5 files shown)


animavitis:
added for ESP32. Should i add it to other boards too?

1technophile:
Yes for esp8266 and arduino mega (__AVR_ATmega1280__)

animavitis:
done:)

1technophile:
It's progressing well :-)

1technophile:
Maybe we could remove the macro for each attribute and supposing that if there is a bme280 we expose all these?

animavitis:
as u wish:)

1technophile:
Do you have other optimizations planned or do I integrate it?

PR #323, opened 2018-12-06
Mqtt discovery for ha

1technophile:


1technophile:
Add MQTT sensors and gateways discovery for home assistant

PR #372, opened 2019-03-15
Add "repeat" field to RF JSON messages.

tjhowse:
Currently the MQTTto433 messages do not support a "repeat" key/value pair. This PR adds support for optionally specifying the number of times an RF message should be transmitted. The default RF_EMITTER_REPEAT remains in effect if "repeat" is not specified in the JSON message.

My use case is a ceiling light which toggles its state upon receipt of the RF signal.


1technophile:
Thanks for this PR!

PR #395, opened 2019-05-01
Introduced safeDelayMicroseconds() function to avoid wdt resets

pkozul:
In the rc-switch library, protocol 31 (Mertik Maxitrol G6R-H4T1) includes a guard time of 17,000 microseconds. The ON command needs to be transmitted over 80 times for the fireplace to actually switch on. These lengthy delays were causing frequent crashes of the ESP8266 (wdt reset). A new function safeDelayMicroseconds() has been added to replace the usage of the standard delayMicroseconds() function. This new function uses yield() within the delay loop, allowing the ESP to 'breathe' during these delays. After implementing this function and changing 3 lines of code to make use of this new function, no more crashes (wdt reset) occur. I have even tested using a repeat count of 200, and the transmission went fine.

Below is an example of how I am using this protocol within Home Assistant. As you can see, the ON command has a repeat count of 100, and is successfully able to switch on my Heat & Glo fireplace.

```
switch:
  - platform: mqtt
    name: Fireplace
    command_topic: "home/OpenMQTTGateway/commands/MQTTto433"
    payload_on: '{ "protocol": 31, "delay": 400, "length": 12, "repeat": 100, "value": 1267 }'
    payload_off: '{ "protocol": 31, "delay": 400, "length": 12, "repeat": 10, "value": 1271 }'
    optimistic: false
    retain: false
```

### DIFF
--- a/lib/rc-switch/RCSwitch.cpp
+++ b/lib/rc-switch/RCSwitch.cpp
@@ -504,6 +504,21 @@ void RCSwitch::sendTriState(const char* sCodeWord) {
 }
 
 /**
+ * @param duration   no. of microseconds to delay
+ */
+static inline void safeDelayMicroseconds(unsigned long duration) {
+#if defined(ESP8266) || defined(ESP32)
+  // uses yield() to avoid wdt reset
+  unsigned long start = micros();
+  while ((micros() - start) < duration) {
+    yield();
+  }
+#else
+  delayMicroseconds(duration);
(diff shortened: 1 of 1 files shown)


1technophile:
Thanks for the PR. Do your modification impact the other protocols ?

pkozul:
Will submit another pull request, with less impact to other protocols.

PR #396, opened 2019-05-02
Use safeDelayMicroseconds() for guard delay to avoid wdt resets

pkozul:
In the rc-switch library, protocol 31 (Mertik Maxitrol G6R-H4T1) includes a guard time of 17,000 microseconds. The ON command needs to be transmitted over 80 times for the fireplace to actually switch on. These lengthy delays were causing frequent crashes of the ESP8266 (wdt reset). A new function safeDelayMicroseconds() has been added to replace the usage of the standard delayMicroseconds() function. This new function uses yield() within the delay loop, allowing the ESP to 'breathe' during these delays. After implementing this function and changing the code to make sure the guard delay uses this new function, no more crashes (wdt reset) occur. I have even tested using a repeat count of 200, and the transmission went fine.

Below is an example of how I am using this protocol within Home Assistant. As you can see, the ON command has a repeat count of 100, and is successfully able to switch on my Heat & Glo fireplace.

switch:
  - platform: mqtt
    name: Fireplace
    command_topic: "home/OpenMQTTGateway/commands/MQTTto433"
    payload_on: '{ "protocol": 31, "delay": 400, "length": 12, "repeat": 100, "value": 1267 }'
    payload_off: '{ "protocol": 31, "delay": 400, "length": 12, "repeat": 10, "value": 1271 }'
    optimistic: false
    retain: false

Some additional info about delays:

**Timing and delays**

millis and micros return the number of milliseconds and microseconds elapsed after reset, respectively.

delay pauses the sketch for a given number of milliseconds and allows WiFi and TCP/IP tasks to run. delayMicrosecondspauses for a given number of microseconds.

Remember that there is a lot of code that needs to run on the chip besides the sketch when WiFi is connected. WiFi and TCP/IP libraries get a chance to handle any pending events each time the loop() function completes, OR when delay(...)is called. If you have a loop somewhere in your sketch that takes a lot of time (>50ms) without calling delay(), you might consider adding a call to delay function to keep the WiFi stack running smoothly.

There is also a yield() function which is equivalent to delay(0). The delayMicroseconds function, on the other hand, does not yield to other tasks, so using it for delays more than 20 milliseconds is not recommended.

### DIFF
--- a/lib/rc-switch/RCSwitch.cpp
+++ b/lib/rc-switch/RCSwitch.cpp
@@ -504,6 +504,26 @@ void RCSwitch::sendTriState(const char* sCodeWord) {
 }
 
 /**
+ * @param duration   no. of microseconds to delay
+ */
+static inline void safeDelayMicroseconds(unsigned long duration) {
+#if defined(ESP8266) || defined(ESP32)
+  if (duration > 10000) {
+    // if delay > 10 milliseconds, use yield() to avoid wdt reset
+    unsigned long start = micros();
+    while ((micros() - start) < duration) {
+      yield();
+    }
+  }
+  else {
+    delayMicroseconds(duration);
+  }
+#else
+  delayMicroseconds(duration);
+#endif
+}
+
+/**
  * @param sCodeWord   a binary code word consisting of the letter 0, 1
  */
 void RCSwitch::send(const char* sCodeWord) {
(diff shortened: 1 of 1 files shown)


1technophile:
Hello @animavitis, are you using protocols 11, 12 or 15 , they are affected by this PR. If yes it could be interesting if you could do some tests.
As you introduced the new rcswitch version with this extended list of protocols it could be interesting to have your feedback.

1technophile:
Thanks for the PR !

PR #403, opened 2019-05-16
Add is_defined filter to the value_template

kabturek:
Without the is_defined filter home assistant will reset the value to an empty string every time a topic is posted without the value template. For example when `home/OpenMQTTGateway/BTtoMQTT/--SNIP-- {"moi":2}` is posted - all other sensor with associated topic `home/OpenMQTTGateway/BTtoMQTT/--SNIP--` will get `""` as their value


1technophile:
That's great, I was thinking on developping some variable storage just for this issue but it would have been overkill.

Would it be possible to  add also "is_defined" to mi jia sensor auto discovery?

kabturek:
Added the filter mi jia too. (Also added device_class for the sensors that use it)

PR #426, opened 2019-07-22
[ci] disable telemetry

Rotzbua:
by default telemetry data is sent


1technophile:
hello,

I think telemetry can help improving platformio, do you have concern with telemetry proposed by platformio?

Legion2:
this project is public and open source, so no private informations can be leaked

Legion2:
Closing because no answer after three weeks

PR #438, opened 2019-08-05
Check bool value of simplePublishing

jan-hendrikscholz:
Hey @1technophile,
lots of respect for creating OMG! I just tried to switch over to the JSON formatted MQTT messages for my BLE devices and found out a strange issue. Even if simplePublishing was commented in the user_config, I was able to see the simple data my mqtt broker (mosquitto) and was able to confirm this via the serial monitor.

Instead of creating an issue and creating efforts on your side, I *tried* to write a fix.
Downside of this would be, that simplePublishing has to be defined as true or false in the user config.
But, to be honest, i'm not that deep in C-development. I learned C# years ago and you might have a better solution for this. So don't hesitate and decline my PR ;-) 
My test using my ESP32-devkit C is running fine.
Again, thank your for all your efforts!
Best regards
Jan-Hendrik

### DIFF
--- a/main/main.ino
+++ b/main/main.ino
@@ -295,7 +295,7 @@ void pub(char * topicori, JsonObject& data){
       pub(topic, JSONmessageBuffer);
     #endif
 
-    #ifdef simplePublishing
+    #if(simplePublishing == true)
       trc(F("Pub data per topic"));
       // Loop through all the key-value pairs in obj 
       for (JsonPair& p : data) {


Legion2:
I checked for definitions of `simplePublishing`, but it's only defined in the `User_config.h` and there it is commented out. So you maybe have somewhere `#define simplePublishing  false` in you code or defined it in the build options.

jan-hendrikscholz:
You got me... It's defined in the platformio.ini for com-esp which will be inherited to esp32dev-ble. Would you agree if I close this PR?

Legion2:
Yes you can close the PR. @1technophile should the platformio default settings be the same as in the User_config? Also currently it's not possible to disable settings that are enabled in the User_config by default.

1technophile:
The philosophy is the following; Json publishing and receiving are considered as the per default/minimal configuration and defined permanently into user_config.h.
For the boards that are having enough memory I enable through platformio also simple receiving. This is done in platformio.

jan-hendrikscholz:
@1technophile Thank you for the clarification. As this PR is totally nonsense, I'll close it.

1technophile:
You're welcome.

PR #444, opened 2019-08-26
Correct missing module string for DHT

mpember:
The DHT module was not being appended to the modules string reported in the SYStoMQTT messages.

I also noticed the code incorrectly references DHT functionality for the INA226 module.


1technophile:
Thanks for the fix !

Issue #464, opened 2019-09-29
Bad values for moisture - Mi Flora


leonols:
**Describe the bug**
Values fo moisture are bad, for other characteristics they are good.
I noticed that rest_data for moisture is 2 characters shorter

```
Creating BLE buffer
device detected
C47C8D65D539
BLErssi
-75
txPower
-59
BLE DISTANCE :
5.83
Pub json into:
home/home_presence/OpenMQTTGateway_ESP32_BLE
{"id":"c4:7c:8d:65:d5:39","name":"Flower care","rssi":-75,"distance":5.832737}
Get services data :
1
Pub json into:
home/OpenMQTTGateway_ESP32_BLE/BTtoMQTT/C47C8D65D539
{"id":"c4:7c:8d:65:d5:39","name":"Flower care","rssi":-75,"distance":5.832737,"servicedata":"712098002b39d5658d7cc40d0910028b01","servicedatauuid":"0000fe95-0000-1000-8000-00805f9b34fb"}
Processing BLE device data
mi flora data reading
Creating BLE buffer
rest_data
712098002b39d5658d7cc40d0910028b01
data_length
4
395.00
Pub json into:
home/OpenMQTTGateway_ESP32_BLE/BTtoMQTT/C47C8D65D539
{"fer":395}
Processing BLE device data
Creating BLE buffer
device detected
C47C8D65CA17
BLErssi
-73
txPower
-59
BLE DISTANCE :
4.76
Pub json into:
home/home_presence/OpenMQTTGateway_ESP32_BLE
{"id":"c4:7c:8d:65:ca:17","name":"Flower care","rssi":-73,"distance":4.756481}
Get services data :
1
Pub json into:
home/OpenMQTTGateway_ESP32_BLE/BTtoMQTT/C47C8D65CA17
{"id":"c4:7c:8d:65:ca:17","name":"Flower care","rssi":-73,"distance":4.756481,"servicedata":"712098009917ca658d7cc40d08100111","servicedatauuid":"0000fe95-0000-1000-8000-00805f9b34fb"}
Processing BLE device data
mi flora data reading
Creating BLE buffer
rest_data
712098009917ca658d7cc40d08100111
data_length
2
1.00
Pub json into:
home/OpenMQTTGateway_ESP32_BLE/BTtoMQTT/C47C8D65CA17
**{"moi":1}**
```

The correct moisture value for the above example is 17

i'm use:
OpenMQTTGateway version 0.9.2
[...]

PR #465, opened 2019-09-29
Fix Mi Flora moisture reading

kanttti:
This fixes a bug in single byte values reported in issue #464.
Bug also affected 2 byte and longer values, but in these values first 4 bits seem to be usually irrelevant.


1technophile:
Thanks for the PR!

PR #466, opened 2019-09-29
Add Unit of Measurement for MiFlora Fertility and Fix MiJia BT battery to percentage

kanttti:
Home assistant requires unit of measurement for data to display correctly in graphs.
Added unit "µS/cm" for MiFlora Fertility.

Also after fixing single byte issue (PR #465) noticed that MiJia Bluetooth sensor battery unit must be % instead of V. Esphome also uses % as MiJiaBT unit.

I don't have any Xiaomi LYWSD02 or ClearGrass sensor available so not able to check which battery value they are reporting.


1technophile:
Thanks!

PR #470, opened 2019-10-14
Added NTP support for ESP platform

johnrosshunt:
This patch (hopefully) fixes WiFi connection issues for the ESP platform.  OMG was getting disconnected from my Apple router every 15-20 minutes, or so.

I had a similar issue with a Linux laptop that wouldn't connect to WiFi until the time was corrected in the system BIOS.

As an experiment, I added NTP support to OMG to see if would make a difference. It seems to have fixed the problem for me.  Nearly 3 days uptime without a single disconnect.

The patch adds [simpleDSTadjust](https://github.com/neptune2/simpleDSTadjust) library by neptune2.


1technophile:
Hello,

Do you think you could take into account the comments?

PR #484, opened 2019-11-04
Add Livolo 433Mhz switch support

robertcsakany:
There is the initial version of Livolo switch support. Tested with Livolo clone Welaik variant with an ESP32 WROOM.

 mosquitto_pub -h mqtt.local -t 'home/Gw1/commands/MQTTtoLivolo' -m "{\"remoteId\" : \"13770\", \"keyCode\" : \"16\" }"

### DIFF
--- a/main/ZgatewayLivolo.ino
+++ b/main/ZgatewayLivolo.ino
@@ -33,27 +33,30 @@
 
 bool livoloIsHigh;
 
-void setupLivolo(){
+void setupLivolo()
+{
   trc(F("LIVOLO_EMITTER_PIN "));
   trc(String(LIVOLO_EMITTER_PIN));
   trc(F("ZgatewayLivolo setup done "));
 }
 
 
-void MQTTtoLivolo(char * topicOri, JsonObject& data) {
-
-  int result = 0;
-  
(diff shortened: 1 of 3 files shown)


1technophile:
Hello,

Thanks for the PR, I see that Livolo seems to be supported by ZgatewayPilight
https://wiki.pilight.org/livolo_switch_v7_0

Did you had a chance to check it?

robertcsakany:
@1technophile I haven't checked yet, I haven't know about it before.  I make a check if works.

robertcsakany:
@1technophile 
I've checked PILight and I found a code in the rewrite branch. But. My problem is with that version it does not handle the key codes right. It maps the different remote controller types, but not all variant and not with the good way. - so it makes button number and calculates the keycode.
It does not maps the 4 key token correctly, in my set, so for me is better to send the keyCode itself.

1technophile:
Thanks for the check. Maybe it could be interesting to submit the modification into the Pilight repository ? This way it will be taken into account for OMG and for Pilight.

If you want it on a more direct way. You could use rcswitch [fork](https://github.com/1technophile/rc-switch) as a support. Enabling to gather your work with all the other RF protocols.

PR #520, opened 2020-01-02
Replace fixed allocation of json buffer

1technophile:
by a dynamic one, and simplify trace function of json objects


1technophile:
keep it the same for arduino UNO due to its lower memory capacity

PR #532, opened 2020-01-16
Add support for weather stations

BlackSmith:
This PR adds support for these weather stations:
- Venus W174/W132
- Auriol H13726   (tested)
- Hama EWS 1500
- Meteoscan W155/W160

I tested this script on ESP-WROOM-32.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -44,6 +44,7 @@ extra_configs =
 ;default_envs = rf-wifi-gateway
 ;default_envs = nodemcuv2-rf2
 ;default_envs = nodemcuv2-pilight
+;default_envs = nodemcuv2-weatherstation
 ;default_envs = nodemcuv2-ir
 ;default_envs = nodemcuv2-ble
 ;default_envs = nodemcuv2-all
@@ -431,7 +432,7 @@ build_flags =
   '-DGateway_Name="OpenMQTTGateway_ESP8266_Pilight"'
 board_build.flash_mode = dout
 
-[env:nodemcuv2-rf3]
+[env:nodemcuv2-weatherstation]
 platform = ${com.esp8266_platform}
 board = nodemcuv2
 lib_deps =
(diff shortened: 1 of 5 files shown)


1technophile:
Thanks for this PR!

platformio.ini
```diff
@@ -33,6 +33,7 @@ extra_configs =
 ;default_envs = esp32dev-all
 ;default_envs = esp32dev-rf
 ;default_envs = esp32dev-pilight
+;default_envs = esp32dev-weatherstation
```
1technophile: If the code works with esp8266 and/or arduino would it be possible to add the corresponding environment definitions?

BlackSmith:
I added environment definitions for eps82266 and arduino. 
I have tested esp8266.
I tried to test Arduino UNO with mini W5100, but arduino can not connect to MQTT server (error rc=5 on arduino and "Socket error on client <unknown>, disconnecting." on MQTT server). 
Maybe hardware issue on my W5100. :(

1technophile:
Thanks, It could be an issue also with the arduino having not enough memory to run the program correctly.

BlackSmith:
OK, I removed Arduino configuration.

1technophile:
Thanks

Dattel:
Does that feature also work with the new CC1101 feature?

1technophile:
@orrpan any idea on @Dattel question?

orrpan:
@Dattel @1technophile no, the weathermodule uses its own RF functions (I think) not rc-switch/newswitch as RF/RF2. But I can look at it, I don’t have a weather station to test with

Dattel:
hmmm.... okay no worries... I was just a question.. 
I wired some RBX6 and build a omqttg with weatherplugin but it only gathers crap, since my sensors are talking a different protocol..... so currently no use for me :-(

PR #538, opened 2020-01-22
module for DS1820 temperature sensor(s)

lrswss:
- adds support for Dallas DS1820 1-wire temperature sensors
- requires OneWire and DallasTemperature libraries
- reads temperatures (non-blocking) from multiple sensors
- tested with ESP01 and Wemos D1 mini, the later along with IR/RF modules/peripherals


1technophile:
Thanks for the PR!

PR #569, opened 2020-03-17
Removed duplicated valueAsASubject option in User_config.h

Legion2:
The `valueAsASubject` was present in two places in the `User_config.h`.
Now it is only in the "MQTT general topics" section.


1technophile:
thanks for pointing this!

PR #577, opened 2020-04-05
Fixed Pilight can not send on RX pin

Legion2:
Set the pin mode of the `RF_EMITTER_PIN` in the `setupPilight`. Actually the pin mode is already set in the ESPiLight constructor, but if the `RF_EMITTER_PIN` is the RX pin the [pin mode is reset](https://github.com/esp8266/Arduino/blob/007e495e0d88eabc6dae31fad311df38f8914607/cores/esp8266/uart.cpp#L714) with the `Serial.end();` call in [main.ino](https://github.com/1technophile/OpenMQTTGateway/blob/development/main/main.ino#L622). So it must be set again.

### DIFF
--- a/main/ZgatewayPilight.ino
+++ b/main/ZgatewayPilight.ino
@@ -60,6 +60,7 @@ void setupPilight()
 #ifndef ZgatewayRF &&ZgatewayRF2 //receiving with Pilight is not compatible with ZgatewayRF or RF2 or RF315 as far as I can tell
   rf.setCallback(pilightCallback);
   rf.initReceiver(RF_RECEIVER_PIN);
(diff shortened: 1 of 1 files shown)


1technophile:
Thanks!

Legion2:
I think the default RX pin should not be used for any gateway. The pin is used for Serial input, which use this pin as INPUT. Using the pin as OUTPUT and connecting it to hardware result in:
1. Serial read not working
2. Serial signal interfere RF emitter signal when testing some sketch without this fix
3. Confusion for users because of 1. and 2.

1technophile:
I agree, this was done to maximize pin usage but users are accumulating a lot of modules on rare cases, and of course less often than using the default  RX pin.

Legion2:
Yes, but I think we should not offer the RX pin as the default to the user, we can use another pin as the default and only use RX if there is no other option because all other pins are used by other modules.

PR #586, opened 2020-04-25
Add M5 boards compatibility and low power mode

1technophile:
-M5StickC & M5Stack display support
-M5StickC & M5Stack log to display
-M5 boards battery infos publishing to MQTT
-Low power modes for ESP32 BLE scanning (1 for barebones and 2 for M5Stack and M5StickC)


Swiftnesses:
@1technophile I have a number of these devices: http://www.lilygo.cn/prod_view.aspx?TypeId=50033&Id=1126&FId=t3:50033:3

Is it possible to use the display on them using a variation of the above?

1technophile:
Not off the shelf like the M5Stack ones, it will require some development.

Issue #593, opened 2020-05-01
Low Power Mode changes GPIO pin even when disabled


jmw6773:
**Describe the bug**
The GPIO used for LOW_POWER_LED is unuable even when LOW_POWER_MODE is disabled when using the BT module no ESP32.

The error occurs in ZgatewayBT.ino:583 as there is no condition to check if LOW_POWER_MODE is enabled before setting the GPIO to value LOW_POWER_LED_OFF. 
https://github.com/1technophile/OpenMQTTGateway/blob/cf641851afc96796f341b7e512a7b57a81cbcada/main/ZgatewayBT.ino#L583

**To Reproduce**
Try to use GPIO2 on ESP32 when BT module is enabled. 

**Expected behavior**
When LOW_POWER_MODE isn't enabled, the pin defined by LOW_POWER_LED shouldn't be changed.

**Proposed fix**
ZgatewayBT.ino:583
```
//only change LOW_POWER_LED if low power mode is enabled
if (low_power_mode > 0)
  digitalWrite(LOW_POWER_LED, LOW_POWER_LED_OFF);
```

**Environment (please complete the following information):**
OpenMQTTGateway version used V0.9.4beta
BT module for ESP32

1technophile:
Thanks for pointing that

Issue #591, opened 2020-04-29
v0.9.4beta Weather Station library missing in all-libraries


Alomon31:
Hi,

Title is explicit, it seeme like WeatherStationDataRx_ID6800 folder is missing in esp32dev-all-libraries.zip (at least, I didn't check for the other hardware).

Thanks for support.

1technophile:
I confirm, for whom interested don't hesitate to propose a pull request with a modified platformio.ini.
The goal is to add the library and the gateway to esp32dev-all and novemcuv2-all env.

Alomon31:
Issue closed as it was corrected in last commit.
However, the contributor has added `weatherstationrx = WeatherStationDataRx@0.3.3` in the libraries although `rfWeatherStation = WeatherStationDataRx@0.3.1` was already declared. Shouldn't it be merged?
I've seen that Pilight is missing as well in esp32-all, I will try and add it.

jmw6773:
@Alomon31, you're right. Good catch, I didn't see the previous declaration for the library. Do you know if version 0.3.3 of the library is okay to use, or should 0.3.1 be kept?

I can make a pull request to remove one of them and update the build definitions as needed.

Alomon31:
Sorry but I don't know the libraries, just saw that when having a first look at the file (I'm quite new here).

jmw6773:
I'm new here too. Jumped in last week to try to help out a little.

I didn't see many changes between the versions when looking at the libraries github page, but I'll keep the original version as the initial push had it and let the sensors' contributor update it if needed.

I'll made a Pull Request tomorrow morning (JST) to fix the platformio.ini file definitions.

PR #594, opened 2020-05-01
Fixing Issues #593 & #591

jmw6773:
- [Issue #593](https://github.com/1technophile/OpenMQTTGateway/issues/593) - Added the condition to address the issue I previously opened to check that Low Power Mode is enabled before making changes to the GPIO pin
- [Issue 591](https://github.com/1technophile/OpenMQTTGateway/issues/591) - Added the WeatherStation libraries and added the module definitions to esp32dev-all and nodemcuv2-all

I don't think any doc changes are needed for these changes, but please let me know if there are any.


1technophile:
Thanks!

1technophile:
Doc changes are not needed indeed

Issue #595, opened 2020-05-01
Make more options configurable via platformio.ini


jmw6773:
**Is your feature request related to a problem? Please describe.**
I have a few different designs that use OpenMQTTPlatform that use different modules and pins. It is a hassle to have multiple versions of the code. It would be much easier to be able to create an entry in the platform.ini file and be able to define the pin and some of the commonly changed attributes for the sensors.

**Describe the solution you'd like**
Where possible, I'd like to add definition checks, like is done in User_config.h for the Gateway_name, so that if the variable/pin is defined in the platform.ini file, that value is used.

i.e. For config_ADC.h:38
```
#if defined(ESP8266) || !defined(ADC_PIN)
  #define ADC_PIN A0 //on nodeMCU this is D3 GPIO0
#endif
```

**Describe alternatives you've considered**
Some pins can't be changed depending on the platform. (i.e. ADC for ESP8266 can only be A0, and only analog pins for ATMEGA and ESP32. It's up to the user to specify an appropriate pin.)

The user may put bad data for non-pin defines. (i.e. Negative values for TimeBetweenReadingADC.)
Can users be trusted to not do this?

One drawback to this is that it could make the header files more cluttered and difficult to read if it's overdone. But if done for pin assignments and commonly changed values, it might be okay.

1technophile:
I agree to generalize the use of ifndef in config files for all pertinents parameters

PR #597, opened 2020-05-01
Allow pins/settings to be defined in INI file

jmw6773:
As mentioned in Issue #595, I've added #ifndef's to several sensors. 

For sensors that require multiple pins, I have it set so the user must specify all required pins in the platformio.ini file. (i.e. For the BT module, the user can't specify the TX pin and use the default RX pin. They must specify both RX and TX pins.)
I did this to force the user to keep all the definitions together, but they can be separated if needed.

If you see any other fields that would benefit from being assigned in the platformio.ini file (like the 2G BAUD rate), let me know and I can add them.

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -341,8 +341,10 @@ class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks
     { //if not black listed mac we go AND if we have no white mac or this mac is  white we go out
       if (advertisedDevice.haveName())
         BLEdata.set("name", (char *)advertisedDevice.getName().c_str());
+      #if pubBLEManufacturerData
       if (advertisedDevice.haveManufacturerData())
         BLEdata.set("manufacturerdata", (char *)advertisedDevice.getManufacturerData().c_str());
+      #endif
       if (advertisedDevice.haveRSSI())
         BLEdata.set("rssi", (int)advertisedDevice.getRSSI());
       if (advertisedDevice.haveTXPower())
--- a/main/config_BT.h
+++ b/main/config_BT.h
@@ -48,7 +48,7 @@ extern void MQTTtoBT(char * topicOri, JsonObject& RFdata);
 #endif
 
 #ifndef pubBLEManufacturerData
-  #define pubBLEManufacturerData true // comment if you don't want to publish the manufacturer's data (in case data has characters that aren't valid with receiving client)
+  #define pubBLEManufacturerData false // define true if you want to publish the manufacturer's data (sometimes contains characters that aren't valid with receiving client)
 #endif
 
 /*-------------------HOME ASSISTANT ROOM PRESENCE ----------------------*/
--- a/main/config_BT.h
+++ b/main/config_BT.h
@@ -39,7 +39,10 @@ extern void MQTTtoBT(char * topicOri, JsonObject& RFdata);
 //#define HM-11 // uncomment this line if you use HM-11 and comment the line above
 //#define HM_BLUE_LED_STOP true //uncomment to stop the blue led light of HM1X
 #define BLEdelimiter "4f4b2b444953413a"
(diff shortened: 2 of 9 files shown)


main/config_BT.h
```diff
@@ -32,14 +32,24 @@ extern void MQTTtoBT(char * topicOri, JsonObject& RFdata);
...
+#ifndef TimeBtw_Read
+  #define TimeBtw_Read 55555 //define default time between 2 scans
+#endif
+
+#ifndef pubBLEServiceData
+  #define pubBLEServiceData true // comment if you don't want to publish service data (in case you are having too heavy service data) https://github.com/1technophile/OpenMQTTGateway/issues/318#issuecomment-446064707
+#endif
+
+#ifndef pubBLEManufacturerData
+  #define pubBLEManufacturerData true // comment if you don't want to publish the manufacturer's data (in case data has characters that aren't valid with receiving client)
```
1technophile, line 51: I would comment it per default, as I don't see the use case behind it. Moreover this field is badly encoded and I'm suspecting it can cause troubles.

main/config_BT.h
```diff
@@ -32,18 +32,25 @@ extern void MQTTtoBT(char * topicOri, JsonObject& RFdata);
...
+#endif
+
 #ifndef pubBLEServiceData
   #define pubBLEServiceData true // comment if you don't want to publish service data (in case you are having too heavy service data) https://github.com/1technophile/OpenMQTTGateway/issues/318#issuecomment-446064707
 #endif
 
+#ifndef pubBLEManufacturerData
+  #define pubBLEManufacturerData true // comment if you don't want to publish the manufacturer's data (in case data has characters that aren't valid with receiving client)
+#endif
+
```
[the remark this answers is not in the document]
jmw6773: I set the value to 'false' by default.

main/config_BT.h
```diff
@@ -48,7 +48,7 @@ extern void MQTTtoBT(char * topicOri, JsonObject& RFdata);
 #endif
 
 #ifndef pubBLEManufacturerData
-  #define pubBLEManufacturerData true // comment if you don't want to publish the manufacturer's data (in case data has characters that aren't valid with receiving client)
+  #define pubBLEManufacturerData false // define true if you want to publish the manufacturer's data (sometimes contains characters that aren't valid with receiving client)
```
1technophile, line 51: We should comment it I think, if it is only set to false it will be considered as defined in ZgatewayBT

main/ZgatewayBT.ino
```diff
@@ -341,8 +341,10 @@ class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks
     { //if not black listed mac we go AND if we have no white mac or this mac is  white we go out
       if (advertisedDevice.haveName())
         BLEdata.set("name", (char *)advertisedDevice.getName().c_str());
+      #if pubBLEManufacturerData
       if (advertisedDevice.haveManufacturerData())
         BLEdata.set("manufacturerdata", (char *)advertisedDevice.getManufacturerData().c_str());
+      #endif
```
jmw6773, line 347: I'm using the #if directive in ZgatewayBT.ino instead of the #ifndef directive so the user can override the value of 'pubBLEManufacturerData' in the platformio.ini file. The value defaults to 'false' if no define is made and the ManufacturerData is not published.

If using the #ifdef / #ifndef directives, it can be confusing to the user to set their value as true/false if not needed. They both evaluate to the value being defined and the code running.

main/config_BT.h
jmw6773: I couldn't find how to insert a linked code block to a different file in the reply, so I made a comment below with the justification to your comment.

If you'd still prefer commenting out the #define to disable it, I'll update ZgatewayBT,ino to match this method.

main/ZgatewayBT.ino
1technophile: I've just learnt something, I didn't know that we can use #if 
Perfect!

1technophile:
If you are ok I will squash it and merge?

jmw6773:
I haven't had any issues with this change and it shouldn't have a negative impact on current users.

Issue #599, opened 2020-05-02
HA discovery should group the entities created by the OMG device


phdelodder:
**Is your feature request related to a problem? Please describe.**
Using HA discovery all the entities are scared, not group by the device that creates the entries. 

**Describe the solution you'd like**
Group the entries per device, also the MAC adress should be added in the entity_id. 

**Describe alternatives you've considered**
Search for the entry one by one.

version 0.9.4-beta

jmw6773:
Might be doable by concatenating OMG's Gateway_Short_Name (or Gateway_Name) with the sensors' name in the createDiscovery method.

https://github.com/1technophile/OpenMQTTGateway/blob/470790b42e9ea1a8774239e7a11745b8fc283da7/main/ZmqttDiscovery.ino#L85

The Gateway Name is configurable in User_config.h or through the platformio.ini file.

Making this change would make everything show in the form 'sensor.OMG_adc' instead of 'sensor.adc'

@1technophile - If this is useful, I can submit a Pull Request.

1technophile:
The context we have to care about is when using multiple OMG gateways, we must avoid duplicating entities into HA and I think using the Gateway_Short_Name will do the job, contrary to the Gateway_Name that needs to be changed when using several boards.

Yes you're PR is welcome!

phdelodder:
Might I add that for example, I have several mi flora using one gateway i can'( distint one from another there for I would suggest: sensor.'Gateway_Short_Name'_'MiFlora_MAC'_name

1technophile:
> i can't distint one from another

Even with the mac adress?

phdelodder:
With the mac adress of the gateway or the BLE sensor? If the it's the BLE sensor then it would be sufficient.

1technophile:
I propose to continue the discussion [here ](https://community.openmqttgateway.com/t/notes-on-issues-with-ha-discovery/953)if you don't mind

PR #600, opened 2020-05-02
Concat OMG's Short Name to each sensor name

jmw6773:
The Gateway_Short_Name is added as a prefix to each sensor name in ZmqttDiscovery so that they are grouped/searchable within Home Assistant. ( Brought up in Issue #599 )

I'm not sure if using the String concat is the most efficient method to do this, but as the method is only run once per sensor per boot, it shouldn't cause issues or slowdown.

### DIFF
--- a/main/ZmqttDiscovery.ino
+++ b/main/ZmqttDiscovery.ino
@@ -82,7 +82,11 @@ void createDiscovery(char *sensor_type,
   strcat(state_topic, st_topic);
   sensor.set("stat_t", state_topic); //state_topic
 
-  sensor.set("name", s_name);       //name
+  String grouped_s_name = String(Gateway_Short_Name);
+  grouped_s_name.concat('_');
+  grouped_s_name.concat(s_name);
+
+  sensor.set("name", grouped_s_name);       //name
   sensor.set("uniq_id", unique_id); //unique_id
   if (device_class[0])
     sensor.set("dev_cla", device_class); //device_class
--- a/main/ZmqttDiscovery.ino
+++ b/main/ZmqttDiscovery.ino
@@ -82,9 +82,7 @@ void createDiscovery(char *sensor_type,
   strcat(state_topic, st_topic);
   sensor.set("stat_t", state_topic); //state_topic
 
-  String grouped_s_name = String(Gateway_Short_Name);
-  grouped_s_name.concat('_');
-  grouped_s_name.concat(s_name);
+  String grouped_s_name = String(Gateway_Short_Name) + "_" + String(s_name);
 
   sensor.set("name", grouped_s_name);       //name
   sensor.set("uniq_id", unique_id); //unique_id


jmw6773:
Should mention that this could be a Breaking Change for users that are upgrading from a previous version of OpenMQTTGateway and have it configured in Home Assistant. 

The name and entity_id will change for each sensor, requiring the user to update their Home Assistant UI, automations, and another location they are using the entity_id.

1technophile:
Do you know if there is an easy way to update the entities in bulk in HA?

jmw6773:
I don't really use MQTT Discovery in Home Assistant, so I'm not sure to to update entities in HA.

Looking further into this, the way mqttDiscovery might need to be rethought for some of the elements. As it is now, it can be confusing when you have multiple OMG's connected.

If entity_id isn't unique, Home Assistant postfix's a counter on the end of the entity_id. This makes it more difficult to identity which sensor belongs to which device. 

Sensors directly attached to OMG and the internal commands (erase/restart) should probably be prefixed using the full Gateway Name so they are easily identified/unique and match the devices' name.

```
sensor.restart_omg --> sensor.openmqttgateway_restart_omg
sensor.adc         --> sensor.openmqttgateway_adc
```

Should sensors that are created from wireless devices (i.e. BT), have a unique identifier in their entity_id as well? (I have 2 MiBands at my house. Should the MAC be in the Entity_ID to make them identifiable? Or at least in the MQTT packet sent with the MiBand Steps information?)

sensor.omg_restart_omg belongs to which device?
Is it guaranteed that sensor.omg_erase_omg belongs to the same device?
HA_discovery

1technophile:
> Should sensors that are created from wireless devices (i.e. BT), have a unique identifier in their entity_id as well?

In my point of view yes, devices not physically attached to OMG should have an unique identifier accross the different gateways that detected it, but for easy searching we can add the fact that they have been discovered by OMG, the mac adress is a good way to do that. Something like sensor.OMG_582D3450FDAF_miband_steps maybe?

> Or at least in the MQTT packet sent with the MiBand Steps information?)
> sensor.omg_restart_omg belongs to which device?
> Is it guaranteed that sensor.omg_erase_omg belongs to the same device?

The devices physically attached or functions (erase, restart) may take the mac adress of the gateway itself or the Gateway_Name with also the short name concatenation (to have a consistent way of searching).

That's some ideas, I'm open to other proposals.

By the way thanks for your help :-)

jmw6773:
Let's cancel this Pull Request for now and have a think about what would work best for HADiscovery.

I think there are two or three issues that need to be discussed and resolved first. Merging this pull request without thinking it though may cause more issues for users and future builds.

1technophile:
I agree, maybe it may be good to start a thread here to reach opinions:
https://community.openmqttgateway.com/

jmw6773:
Thank you for your time in this awesome project. I hope I can help a little.

[Discussion thread created](https://community.openmqttgateway.com/t/notes-on-issues-with-ha-discovery/953)

PR #604, opened 2020-05-05
Remove duplicate Weatherstation library definition from platformio.ini

jmw6773:
@Alomon31 from Issue #591 noticed that the WeatherStation library was already in the Platformio.ini file, so my entry was a duplicate. I removed my entry and changed the *-all definitions to use the sensor author's version.
There are also definitions for ATMega to use this library, so I added it to the [env:atmega-all] block.


1technophile:
thanks

PR #610, opened 2020-05-10
Add support for multiple IR LEDs

vossim:
- IR_EMITTER_PIN renamed to IR_EMITTER_PIN_0
- Added IR_EMITTER_PIN_1, can be expanded to more IR emitters if
  required

### DIFF
--- a/main/ZgatewayIR.ino
+++ b/main/ZgatewayIR.ino
@@ -39,11 +39,14 @@ IRrecv irrecv(IR_RECEIVER_PIN, 1024, 15U, true);
 #else
 IRrecv irrecv(IR_RECEIVER_PIN);
 #endif
-IRsend irsend(IR_EMITTER_PIN);
+IRsend irsend0(IR_EMITTER_PIN_0);
+#ifdef IR_EMITTER_PIN_1
+IRsend irsend1(IR_EMITTER_PIN_1);
+#endif
 #else
 #include <IRremote.h>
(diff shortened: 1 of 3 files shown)


jmw6773:
Are both IR pins sending the same data. If this is is the case, why not just put the IR LED's in parallel, each with a current limiting resistor? I have this on my gateways, one with 5 IR LED's in parallel without issues. Just be sure to drive them using a transistor or mosfet.

With multiple IR LED's transmitting the same data, one after the other (they don't transmit synchronously), a receiving device may receive both signals and behave oddly. (Mute button for TV mutes then immediately unmutes, etc)

vossim:
The output pin can be "selected" by passing in a `short` named `out` in the JSON.

```{"value":46505948479624,"protocol_name":"DENON","out":0}```
Would trigger the default emitter pin

```{"value":46505948479624,"protocol_name":"DENON","out":1}```
Would trigger the second emitter

1technophile:
May you indicate the use case behind this proposal?

vossim:
I'm using this to control multiple devices using the same IR signals independently. So I can control 2 similar receivers using one esp8266, otherwise I would need 2 OMG devices.

1technophile:
Thanks for the PR, nevertheless I think there is other way to answer to your use case.

If you want an integration that answer it, I would propose to do a PR with the capacity to change the IR emitter pin when sending or by a special MQTT command.
We can discuss about that on the design principle if you are interested.

Issue #331, opened 2018-12-19
BLE Whitelist.


maxi1134:
**Is your feature request related to a problem? Please describe.**
A Whitelist for the Bluetooth Scan for MQTT presence location.

**Describe the solution you'd like**
A whitelist, this would allow to reduce irrelevant data on the MQTT server, Living in an appartment block, I usually get 20 to 30 devices detected per scan. 

**Describe alternatives you've considered**
A black list could also be usefull.

**Additional context**
Pretty straightforward.

1technophile:
Do you think an mqtt json payload with the list of whitelist & blacklist devices answer the need?

rickitaly:
Agree, would be great to scan and get mqtt msg only for relevant BLE devices!

tunip:
Thats the way ESPHome works. Define MAC for relevant devices. Would be great OMG has this feature too.

NogNog87:
It would be great to get a scan of one or more g-tags (beacons) that you define yourself (whitelist). Perfect would be information that the beacon is in range or not. If it not in range you get the payload "off" or similar

1technophile:
I'm planning to add this feature to the next release. Nevertheless I'm still thinking about how to set this whitelist. @maxi1134 the drawback with hardcoded list is that users that take directly the binary file will not be able to use this feature. Why would you not be interested in passing the list by mqtt?

maxi1134:
@1technophile I'm not against the idea. MQTT Would be a more elegant solution now that I think about it.
Easier to append more devices without reflashing everything.

1technophile:
ok thanks for the feedback I will design it like that

1technophile:
60f68d00f4dbf0acb629e11c5f42485228084e9a

NogNog87:
Hello,

where i can set the Whitelist?

Greetz
Daniel

1technophile:
Hello, the infos were on the commit msg:
set black list
`mosquitto_pub -t home/OpenMQTTGateway/commands/MQTTtoBT/set -m '{"black-list":["012314551615","4C6577889C79","edfrgtyjulik"]}'`
set white list
`mosquitto_pub -t home/OpenMQTTGateway/commands/MQTTtoBT/set -m '{"white-list":["012314551615","4C65A5553C79","edfrgtyjulik"]}'`

maxi1134:
[...]

PR #615, opened 2020-05-13
Add support to disable/enable white list control

ahome-wugu:
#331 It would be more convenient to manage white list by adding config message to temporarily disable the already  set white list  and enable later , e.g. {"white-list":"false"} and {"white-list":"true"}, because sometimes we may need to observe the new approaching BLE mac via mqtt gateway by temp disable the already set white list, and add the new desire one and then enable the white list again, fyi !


ahome-wugu:
sorry, not yet familiarize how to submit PR.

Issue #598, opened 2020-05-02
OneWiretoMQTT support HA discovery


phdelodder:
**Is your feature request related to a problem? Please describe.**
OneWiretoMQTT doesn't use HA discovery

**Describe the solution you'd like**
OneWiretoMQTT should support HA discovery

**Describe alternatives you've considered**
Manual Configuration

**Additional context**

Version 0.9.4-Beta

jmw6773:
By OneWire, do you mean the DS1820 temperature sensor?

jmw6773:
@1technophile, if this functionality is something you'd like added, I think the DS1820 sensor could be made HA Discovery compatible by making each DS1820 sensor write to a sub-MQTT topic using it's unique identifier.

In the device setup, for each sensor detected, it would publish the HA Discovery information. (Like how BT gateway does it, but only once during the DS1820 setup.)

ZsensorDS1820.ino: L172
`pub(OW_TOPIC + String("/") + ds1820_addr[i], DS1820data);`


The createDiscoveryFromList would need to be re-written, or a new function created, to allow other topics.
https://github.com/1technophile/OpenMQTTGateway/blob/0afcb61717f5d8bd98afb951e1a108bfe8327af1/main/ZmqttDiscovery.ino#L53

1technophile:
Hello,

Yes we could add it, PR welcome!

jmw6773:
@phdelodder , could you pull the dev version to see if it resolves the HADiscovery for OneWire issue? If you can't build it, I can compile it for you if you send me which configuration you would like to use.

phdelodder:
Did some testing but the sensor can't be found any more, tried other firmware and it was discovered. Don't know what changed but it's unusable now.

jmw6773:
Do you see any sensors under the MQTT integration for the device?

HA --> Configuration --> Integrations --> MQTT
   Under MQTT, click the `# devices` link
   Find and click on your OMG device

phdelodder:
No nothing, the device isn't even found physically

jmw6773:
Could you let me know what hardware you are using (ESP8266, ESP32, etc), which pin you have the DS18b20 attached to, and the config you have for your platform.ini? I'll try to duplicate your setup and see if I can replicate the issue.

phdelodder:
I was using a ESP32, on pin 2. 

```
[env:esp32dev-ble]
platform = ${com.esp32_platform}
board = esp32dev
[...]

PR #621, opened 2020-06-04
Ds1820 HADiscovery

jmw6773:
Pull Request in response to Issue #598 ; to add HADiscovery for DS18b20 1-Wire sensors

@1technophile 
   Looking at the source code and through testing with hardware, I was only able to get 1 sensor at a time to work using the original ZsensorDS1820 code. As it looks as the intention was to allow up to 8 sensors by default, I've modified the following:

-    I changed setupZsensorDS1820 to use the DallasTemperature library to scan the 1-Wire bus and validate that the sensor is a valid DS1820 sensor that can be used instead of using the OneWire library directly.
-    I changed the DS1820 MQTT topic so that all sensors are published as a sub-topic of the DS1820 topic, using their device address as the topic name.
-    Added function to publish the HADiscovery config for each DS1820 device found. This is called during the normal 'pubMqttDiscovery()' call from the main loop() function on the first WiFi connect.


1technophile:
Perfect! Thanks

PR #634, opened 2020-06-14
Add a lock mechanism to improve the OTA chances of success with ESP32 and BLE

1technophile:
Lock BLE scan and MQTT process when OTA is started

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -386,26 +386,28 @@ void coreTask(void* pvParameters) {
     Log.trace(F("BT Task running on core: %d" CR), xPortGetCoreID());
     if (!low_power_mode)
       delay(BLEinterval);
-    int n = 0;
-    while (client.state() != 0 && n <= TimeBeforeMQTTconnect) {
-      n++;
-      Log.trace(F("Wait for MQTT on core: %d attempt: %d" CR), xPortGetCoreID(), n);
-      delay(1000);
-    }
-    if (client.state() != 0 || ProcessLock) {
-      Log.warning(F("MQTT client disconnected or OTA in progress no BLE scan" CR));
-      delay(1000);
-    } else {
-      pinMode(LOW_POWER_LED, OUTPUT);
-      if (low_power_mode == 2)
-        digitalWrite(LOW_POWER_LED, 1 - LOW_POWER_LED_OFF);
-      BLEscan();
-      //only change LOW_POWER_LED if low power mode is enabled
+    if (!ProcessLock) {
+      int n = 0;
+      while (client.state() != 0 && n <= TimeBeforeMQTTconnect) {
+        n++;
+        Log.trace(F("Wait for MQTT on core: %d attempt: %d" CR), xPortGetCoreID(), n);
+        delay(1000);
+      }
+      if (client.state() != 0) {
+        Log.warning(F("MQTT client disconnected or OTA in progress no BLE scan" CR));
+        delay(1000);
+      } else {
+        pinMode(LOW_POWER_LED, OUTPUT);
+        if (low_power_mode == 2)
+          digitalWrite(LOW_POWER_LED, 1 - LOW_POWER_LED_OFF);
+        BLEscan();
+        //only change LOW_POWER_LED if low power mode is enabled
+        if (low_power_mode)
+          digitalWrite(LOW_POWER_LED, LOW_POWER_LED_OFF);
+      }
       if (low_power_mode)
-        digitalWrite(LOW_POWER_LED, LOW_POWER_LED_OFF);
+        lowPowerESP32();
     }
-    if (low_power_mode)
-      lowPowerESP32();
   }
 }
 
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -30,17 +30,17 @@ Thanks to wolass https://github.com/wolass for suggesting me HM 10 and dinosd ht
 */
 #include "User_config.h"
 
-#ifdef ESP32
-#  include "FreeRTOS.h"
+#ifdef ZgatewayBT
+
+#  ifdef ESP32
+#    include "FreeRTOS.h"
 FreeRTOS::Semaphore semaphoreCreateOrUpdateDevice = FreeRTOS::Semaphore("createOrUpdateDevice");
 // Headers used for deep sleep functions
-#  include <driver/adc.h>
-#  include <esp_bt.h>
-#  include <esp_bt_main.h>
-#  include <esp_wifi.h>
-#endif
-
-#ifdef ZgatewayBT
+#    include <driver/adc.h>
+#    include <esp_bt.h>
+#    include <esp_bt_main.h>
+#    include <esp_wifi.h>
+#  endif
 
 #  include <vector>
 using namespace std;
@@ -299,6 +299,8 @@ void MiBandDiscovery(char* mac) {}
 
 //core on which the BLE detection task will run
 static int taskCore = 0;
+// Process lock when we want to use a critical function like OTA for example
+bool ProcessLock = false;
 
 class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks {
   void onResult(BLEAdvertisedDevice advertisedDevice) {
@@ -381,21 +383,28 @@ void BLEscan() {
   esp_bt_controller_deinit();
 }
 
+void stopProcessing() {
+  Log.notice(F("Stop BLE processing" CR));
+  ProcessLock = true;
+}
+
+void startProcessing() {
+  Log.notice(F("Start BLE processing" CR));
+  ProcessLock = false;
+}
+
 void coreTask(void* pvParameters) {
   while (true) {
     Log.trace(F("BT Task running on core: %d" CR), xPortGetCoreID());
-    if (!low_power_mode)
-      delay(BLEinterval);
     if (!ProcessLock) {
       int n = 0;
-      while (client.state() != 0 && n <= TimeBeforeMQTTconnect) {
+      while (client.state() != 0 && n <= InitialMQTTConnectionTimeout) {
         n++;
         Log.trace(F("Wait for MQTT on core: %d attempt: %d" CR), xPortGetCoreID(), n);
         delay(1000);
       }
       if (client.state() != 0) {
-        Log.warning(F("MQTT client disconnected or OTA in progress no BLE scan" CR));
-        delay(1000);
+        Log.warning(F("MQTT client disconnected no BLE scan" CR));
       } else {
         pinMode(LOW_POWER_LED, OUTPUT);
         if (low_power_mode == 2)
@@ -405,8 +414,11 @@ void coreTask(void* pvParameters) {
         if (low_power_mode)
           digitalWrite(LOW_POWER_LED, LOW_POWER_LED_OFF);
       }
-      if (low_power_mode)
+      if (low_power_mode) {
         lowPowerESP32();
+      } else {
+        delay(BLEinterval);
+      }
     }
   }
 }
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -392,8 +392,8 @@ void coreTask(void* pvParameters) {
       Log.trace(F("Wait for MQTT on core: %d attempt: %d" CR), xPortGetCoreID(), n);
       delay(1000);
     }
-    if (client.state() != 0) {
-      Log.warning(F("MQTT client disconnected no BLE scan" CR));
+    if (client.state() != 0 || ProcessLock) {
+      Log.warning(F("MQTT client disconnected or OTA in progress no BLE scan" CR));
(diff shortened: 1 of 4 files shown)


main/main.ino
```diff
@@ -1037,106 +1049,105 @@ void loop() {
 
 #if defined(ESP8266) || defined(ESP32)
   if (WiFi.status() == WL_CONNECTED) {
+    ArduinoOTA.handle();
```
Legion2, line 1052: is this blocking?

main/main.ino
```diff
@@ -1037,106 +1049,105 @@ void loop() {
...
 #else
   if ((Ethernet.hardwareStatus() != EthernetW5100 && Ethernet.linkStatus() == LinkON) || (Ethernet.hardwareStatus() == EthernetW5100)) { //we are able to detect disconnection only on w5200 and w5500
 #endif
     failure_number_ntwk = 0;
-    if (client.connected()) {
+    if (!ProcessLock) { // If no blocking process we can handle MQTT processing
```
Legion2, line 1057: What happens if this is set to true after this if check?

main/main.ino
```diff
@@ -1037,106 +1049,105 @@ void loop() {
...
-      failure_number_ntwk = 0;
+        connectedOnce = true;
+        failure_number_ntwk = 0;
 
-      client.loop();
-
-#if defined(ESP8266) || defined(ESP32)
-      ArduinoOTA.handle();
-#endif
+        client.loop();
```
Legion2, line 1066: Does this block, has this a callback, how long does it loop?

main/main.ino
Legion2, line 1057: can we instead do something like?
```c++
if (ProcessLock) return;
```

main/ZgatewayBT.ino
```diff
@@ -386,26 +386,28 @@ void coreTask(void* pvParameters) {
...
-      if (low_power_mode == 2)
-        digitalWrite(LOW_POWER_LED, 1 - LOW_POWER_LED_OFF);
-      BLEscan();
-      //only change LOW_POWER_LED if low power mode is enabled
+    if (!ProcessLock) {
+      int n = 0;
+      while (client.state() != 0 && n <= TimeBeforeMQTTconnect) {
+        n++;
+        Log.trace(F("Wait for MQTT on core: %d attempt: %d" CR), xPortGetCoreID(), n);
+        delay(1000);
```
Legion2, line 394: Does this block the main loop?

1technophile: no, we are on the ESP32 second core, not used by the main loop

main/main.ino
1technophile: in my point of view it is not possible until `ArduinoOTA.handle();` is called, and this call is done before this condition

1technophile: yes we can

main/main.ino
1technophile: No idea, what are you thinking about?

main/ZgatewayBT.ino
1technophile: Let me do some tests

main/main.ino
1technophile: I have checked and it is not blocking.

main/main.ino
Legion2: After looking into the [code](https://github.com/esp8266/Arduino/blob/master/libraries/ArduinoOTA/ArduinoOTA.cpp)  `ProcessLock` should never be `true` here, because when `ArduinoOTA.handle()` returns there is no update in progress. Either there was no update or it failed. When the update was successful the board is restarted and `ArduinoOTA.handle()` should not have returned.

main/main.ino
Legion2: according to https://github.com/esp8266/Arduino/blob/master/libraries/ArduinoOTA/ArduinoOTA.cpp I don't see why it should not block, It runs on the same thread as the loop so it should be blocking.

main/main.ino
1technophile: ok I will remove this check

1technophile: It was added due to the fact that we were seeing MQTT messages during the OTA process:
https://community.openmqttgateway.com/t/ota-upload-on-esp32/1013/5?u=1technophile

Legion2: the mqtt connect comes after the `Error[4]: E: End Failed` so it's after the OTA failed. The OTA kills all network connections before the update, this includes the mqtt connection. After a failed update the mqtt client reconnects. You don't see that on an successful update because in that case the esp is restarted.

main/main.ino
1technophile: In normal conditions ...

Legion2: In which case it does not block the loop thread while doing an update. And on which thread the update is then executed?

main/main.ino
1technophile: Thanks for the analysis!

main/main.ino
1technophile: Maybe the ESP32 does not use the same implementation, but from my test I confirm it is not blocking:
```
void loop() {
Serial.println("begin times");
times = millis();
Serial.println(times);
  ArduinoOTA.handle();
Serial.println("end times");
times = millis();
Serial.println(times);
}
```

I was able to see the serial monitors prints when uploading with OTA in parallel.

Legion2: then we should make the OTA update blocking by waiting until it's finished.

1technophile: Do you have some proposal so as to do that ? Or do I just add the return when we detect that the OTA is in progress through the ProcessLock status?

The state of the OTA is not a public property from what I have seen.

Legion2: https://github.com/espressif/arduino-esp32/blob/master/libraries/ArduinoOTA/src/ArduinoOTA.cpp this is the implementation for the ESP32, it's a little bit different form the esp8266 one but also blocking (or I don't see where the execution is moved to another thread).

1technophile: I'm redoing the test to confirm or contradict

1technophile: It's strange, I have 2 different behaviours:
* Serial monitor started before the OTA upload, the loop continue displaying prints while the OTA is in progress,
* Serial monitor started during the OTA upload, I see only OTA progress and not the loop prints.

So I have modified my sketch to display into the` .onProgress `callback a variable incremented in the loop, and guess what, it is blocking!
```
1392137s: 0%
1392137s: 0%
1392137s: 0%
1392137s: 0%
1392137s: 0%
1392137s: 0%
1392137s: 0%
1392137s: 0%
1392137s: 1%
1392137s: 1%
1392137s: 1%
1392137s: 1%
1392137s: 1%
1392137s: 1%
1392137s: 1%
1392137s: 2%
1392137s: 2%
1392137s: 2%
1392137s: 2%
1392137s: 2%
1392137s: 2%
1392137s: 2%
1392137s: 3%
1392137s: 3%
1392137s: 3%
1392137s: 3%
1392137s: 3%
1392137s: 3%
1392137s: 3%
1392137s: 3%
1392137s: 4%
1392137s: 4%
1392137s: 4%
1392137s: 4%
1392137s: 4%
1392137s: 4%
1392137s: 4%
1392137s: 5%
1392137s: 5%
1392137s: 5%
1392137s: 5%
1392137s: 5%
1392137s: 5%
1392137s: 5%
1392137s: 6%
1392137s: 6%
1392137s: 6%
1392137s: 6%
1392137s: 6%
1392137s: 6%
1392137s: 6%
1392137s: 7%
1392137s: 7%
1392137s: 7%
1392137s: 7%
1392137s: 7%
1392137s: 7%
1392137s: 7%
1392137s: 7%
1392137s: 8%
1392137s: 8%
1392137s: 8%
1392137s: 8%
1392137s: 8%
1392137s: 8%
1392137s: 8%
1392137s: 9%
1392137s: 9%
1392137s: 9%
1392137s: 9%
1392137s: 9%
1392137s: 9%
1392137s: 9%
1392137s: 10%
1392137s: 10%
1392137s: 10%
1392137s: 10%
1392137s: 10%
1392137s: 10%
1392137s: 10%
1392137s: 10%
1392137s: 11%
1392137s: 11%
1392137s: 11%
1392137s: 11%
1392137s: 11%
1392137s: 11%
1392137s: 11%
1392137s: 12%
1392137s: 12%
1392137s: 12%
1392137s: 12%
1392137s: 12%
1392137s: 12%
1392137s: 12%
1392137s: 13%
1392137s: 13%
1392137s: 13%
1392137s: 13%
1392137s: 13%
1392137s: 13%
1392137s: 13%
1392137s: 14%
1392137s: 14%
1392137s: 14%
1392137s: 14%
1392137s: 14%
1392137s: 14%
1392137s: 14%
1392137s: 14%
1392137s: 15%
1392137s: 15%
1392137s: 15%
1392137s: 15%
```
Sorry for the bad result of the first test.

I will remove the checks that are not related to second core use.

1technophile:
@Legion2 I'm asking if we should not use a task deletion or suspension mechanism https://www.freertos.org/a00130.html instead of the current lock. I don't have experimented these functions yet but it could be a cleaner option. What are your thoughts on that?

Legion2:
Suspending is a very low level operation and could stop the thread in the middle of some critical operation. Currently the BL module is the only one which uses a separate thread/task. Instead of the global lock we can forward the onStart and onEnd callbacks to the BL module which then can create a local lock.

I look into the ZgatewayBT.ino and it is mainly for BL but there is also the `deepSleep` implemented, which disable the esp_wifi. But this does not stop the main loop and so I wonder if the deepsleep works.

1technophile:
>  we can forward the onStart and onEnd callbacks to the BL module which then can create a local lock.

How do you see this implementation?

> the deepSleep implemented, which disable the esp_wifi. But this does not stop the main loop and so I wonder if the deepsleep works.

It works, I have made measurement of power consumption when used.

main/ZgatewayBT.ino
```diff
@@ -386,26 +386,28 @@ void coreTask(void* pvParameters) {
...
-      delay(1000);
-    } else {
-      pinMode(LOW_POWER_LED, OUTPUT);
-      if (low_power_mode == 2)
-        digitalWrite(LOW_POWER_LED, 1 - LOW_POWER_LED_OFF);
-      BLEscan();
-      //only change LOW_POWER_LED if low power mode is enabled
+    if (!ProcessLock) {
+      int n = 0;
+      while (client.state() != 0 && n <= TimeBeforeMQTTconnect) {
```
Legion2, line 391: we should rename `TimeBeforeMQTTconnect` to `InitialMQTTConnectionTimeout`

main/ZgatewayBT.ino
```diff
@@ -386,26 +386,28 @@ void coreTask(void* pvParameters) {
     Log.trace(F("BT Task running on core: %d" CR), xPortGetCoreID());
     if (!low_power_mode)
       delay(BLEinterval);
```
Legion2, line 388: move this to the end of the while true in the else branch of
```
      if (low_power_mode)
        lowPowerESP32();
```

main/ZgatewayBT.ino
```diff
@@ -386,26 +386,28 @@ void coreTask(void* pvParameters) {
...
+    if (!ProcessLock) {
+      int n = 0;
+      while (client.state() != 0 && n <= TimeBeforeMQTTconnect) {
+        n++;
+        Log.trace(F("Wait for MQTT on core: %d attempt: %d" CR), xPortGetCoreID(), n);
+        delay(1000);
+      }
+      if (client.state() != 0) {
+        Log.warning(F("MQTT client disconnected or OTA in progress no BLE scan" CR));
+        delay(1000);
```
Legion2, line 398:
```suggestion
```
why should we wait here. We will wait `BLEinterval` or go to deep sleep any way.

main/main.ino
```diff
@@ -161,6 +161,9 @@ PubSubClient client(eClient);
 //MQTT last attemps reconnection date
 unsigned long lastMQTTReconnectAttempt = 0;
 
+// Process lock when we want to use a critical function like OTA for example
+bool ProcessLock = false;
```
Legion2, line 165: move this into ZgatewayBT

main/main.ino
```diff
@@ -701,15 +704,24 @@ void setOTA() {
   ArduinoOTA.setPassword(ota_password);
 
   ArduinoOTA.onStart([]() {
-    Log.trace(F("Start OTA" CR));
+    Log.trace(F("Start OTA, lock other functions" CR));
+    ProcessLock = true;
```
Legion2, line 708: create a `stopProcessing` in ZgatewayBT and call it here. Same with onEnd and onError create a `startProcessing`

1technophile:
Thanks for the relevant comments!

Legion2:
@1technophile you don't delete branches after merging them?

1technophile:
I delete it in normal situation. But when I give the link of the branch to an user I wait for some days before removing it so as to avoid him to search.

PR #637, opened 2020-06-16
Automatically set mqtt buffer size

balvig:
Thank you for creating (and maintaining!) this fantastic project!

I happened to notice that current versions of `pubsubclient` allow the buffer size to be set
dynamically: https://github.com/knolleary/pubsubclient/pull/282

We can take advantage of this to avoid having to modify one of the
dependencies and instead automatically increase the buffer size when a
powerful board is detected.

It also reduces a step in the documentation.

My first PR to an Arduino project so let me know if I'm doing something
wrong! 😅 🙏 

I suppose this would mean changing the version of the library that is bundled with the
project (`PubSubClient_ID89`?) as well, but I'm not sure how to go about doing so...? 🙇

### DIFF
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -99,9 +99,11 @@ const byte mac[] = {0xDE, 0xED, 0xBA, 0xFE, 0x54, 0x95}; //W5100 ethernet shield
 #if defined(ESP8266) || defined(ESP32) || defined(__AVR_ATmega2560__) || defined(__AVR_ATmega1280__)
 #  define parameters_size     20
 #  define mqtt_topic_max_size 100
+#  define mqtt_max_packet_size 1024
 #else
 #  define parameters_size     15
 #  define mqtt_topic_max_size 50
+#  define mqtt_max_packet_size 128
 #endif
 char mqtt_user[parameters_size] = "your_username"; // not compulsory only if your broker needs authentication
 char mqtt_pass[parameters_size] = "your_password"; // not compulsory only if your broker needs authentication
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -97,12 +97,12 @@ const byte mac[] = {0xDE, 0xED, 0xBA, 0xFE, 0x54, 0x95}; //W5100 ethernet shield
 //MQTT Parameters definition
 //#define mqtt_server_name "www.mqtt_broker.com" // instead of defining the server by its IP you can define it by its name, uncomment this line and set the correct MQTT server host name
 #if defined(ESP8266) || defined(ESP32) || defined(__AVR_ATmega2560__) || defined(__AVR_ATmega1280__)
-#  define parameters_size     20
-#  define mqtt_topic_max_size 100
+#  define parameters_size      20
+#  define mqtt_topic_max_size  100
 #  define mqtt_max_packet_size 1024
 #else
-#  define parameters_size     15
-#  define mqtt_topic_max_size 50
+#  define parameters_size      15
+#  define mqtt_topic_max_size  50
 #  define mqtt_max_packet_size 128
 #endif
 char mqtt_user[parameters_size] = "your_username"; // not compulsory only if your broker needs authentication
--- a/main/main.ino
+++ b/main/main.ino
@@ -605,11 +605,7 @@ void setup() {
 #ifdef ZsensorDHT
   setupDHT();
 #endif
-  Log.trace(F("MQTT_MAX_PACKET_SIZE: %d" CR), MQTT_MAX_PACKET_SIZE);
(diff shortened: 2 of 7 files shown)


1technophile:
> Thank you for creating (and maintaining!) this fantastic project!

Thanks for the feedback!

>We can take advantage of this to avoid having to modify one of the
dependencies and instead automatically increase the buffer size when a
powerful board is detected.

Well seen, this will avoid confusions regarding the MQTT MAX PACKET SIZE modifications.

>I suppose this would mean changing the version of the library that is bundled with the
project (PubSubClient_ID89?) as well, but I'm not sure how to go about doing so...?

It will be changed when we release a new version, and done automaticaly by the github actions setup made by @Legion2 

I begin the review, thanks for your participation!

main/main.ino
```diff
@@ -401,6 +401,7 @@ void connectMQTT() {
   char topic[mqtt_topic_max_size];
   strcpy(topic, mqtt_topic);
   strcat(topic, will_Topic);
+  client.setBufferSize(mqtt_max_packet_size);
```
1technophile, line 404: ok for this change,

below you can find also a warning 
```
#if defined(ESP8266) || defined(ESP32) || defined(__AVR_ATmega2560__) || defined(__AVR_ATmega1280__)
  if (MQTT_MAX_PACKET_SIZE == 128)
    Log.error(F("WRONG PUBSUBCLIENT LIBRARY USED PLEASE INSTALL THE ONE FROM RELEASE PAGE" CR));
#endif
```
I think it could be removed and the trace just before log `mqtt_max_packet_size` instead of `MQTT_MAX_PACKET_SIZE `

main/User_config.h
```diff
@@ -99,9 +99,11 @@ const byte mac[] = {0xDE, 0xED, 0xBA, 0xFE, 0x54, 0x95}; //W5100 ethernet shield
 #if defined(ESP8266) || defined(ESP32) || defined(__AVR_ATmega2560__) || defined(__AVR_ATmega1280__)
 #  define parameters_size     20
 #  define mqtt_topic_max_size 100
+#  define mqtt_max_packet_size 1024
 #else
 #  define parameters_size     15
 #  define mqtt_topic_max_size 50
+#  define mqtt_max_packet_size 128
 #endif
```
1technophile, line 107: You should apply clang format to the files, (we didn't updated the dev docs yet sorry).
Install the Clang-format extension to PIO and do a right click inside the main window of the file and click "format document"

balvig:
Thank you @1technophile! 

I actually discovered https://github.com/1technophile/OpenMQTTGateway/commit/cd3a2a0499d4dadc8fd72f1e8a567717d4430c4f which seemed to suggest the feature was already there at some point, but then later removed?

I therefore started having doubts whether maybe there was a problem with it I'm overlooking...if you think it looks ok however, I will take a look at the comments above and apply the changes to this PR 👍

1technophile:
> I actually discovered cd3a2a0 which seemed to suggest the feature was already there at some point, but then later removed?

Yes later removed due to the fact it was not working, the `#define` was not taken into account.

> I therefore started having doubts whether maybe there was a problem with it I'm overlooking...if you think it looks ok however, I will take a look at the comments above and apply the changes to this PR 👍

No doubt your PR is still relevant ;-)

balvig:
@1technophile I found these as well 🙇 :

https://github.com/balvig/OpenMQTTGateway/blob/e398dfbc7a60cdfb550a30ce5e0c12950acf5adb/docs/upload/troubleshoot.md#L48-L52

https://github.com/balvig/OpenMQTTGateway/blob/e398dfbc7a60cdfb550a30ce5e0c12950acf5adb/docs/use/ir.md#L111-L114

https://github.com/balvig/OpenMQTTGateway/blob/e398dfbc7a60cdfb550a30ce5e0c12950acf5adb/platformio.ini#L121

https://github.com/balvig/OpenMQTTGateway/blob/e398dfbc7a60cdfb550a30ce5e0c12950acf5adb/platformio.ini#L133

I suppose we should do something about those as well? I assume they will get ignored if we begin using `setBufferSize()` 🤔 What do you think?

1technophile:
Well seen ! 😉
All these files should be modified also to take into account the automatic setting you've added.
You can remove MQTT_MAX_PACKET_SIZE from the .ini definitions and also from the docs.

balvig:
@1technophile I had a crack at updating the docs in https://github.com/1technophile/OpenMQTTGateway/pull/637/commits/50819d327aee392f2a177e2e83dc102ba0c7f9c9, seemed that some of the tips in there were still relevant, so I simply updated the place to make changes, but let me know if I'm off track!

1technophile:
No, you are on tracks ;-). Thanks

PR #687, opened 2020-07-04
Add AHTx0 Sensor (AHT10 and AHT20)

jmw6773:
I blew up my last HTU21 sensor (reverse polarity, oops) when updating to the newest OMG version, so I replaced it with an AHT10 temp/humidity sensor. Thought others might find it useful as the AHT sensors are very cheap.

Note: The Adafruit AHTX0 library that I've used supports both the AHT10 and AHT20 sensors, but I've only tested with the AHT10 sensor. Compiles for ESP8266, ESP32, and ATMega, but only physically tested on ESP8266.

The sensor uses the I2C protocol, so no pins definitions are required as it uses the default I2C pins, but can be defined for the ESP32 in config_AHTx0.h if non-standard pins are used.

*As code formatting has changed, please let me know if I'm not following the rules.


1technophile:
Looks good, thanks Jon!

PR #689, opened 2020-07-05
Detect when AHT isn't connected

jmw6773:
Added logic to detect when sensor isn't successfully initialized or a reading from the sensor fails.


jmw6773:
Will create a new request, but want to branch this off first.

PR #695, opened 2020-07-09
Ahtx0 detections

jmw6773:
Minor changes to add logic to detect when sensor isn't available on startup or fails read the sensor correctly.


1technophile:
Thanks for the PR

Issue #101, opened 2017-09-10
add ssl support to connect to mqtt broker


superyarik:
now it can only connect via not secure port, so connection is not secure.

superyarik:
no, i'm using 0.3.1, and can not connect to ssl port of broker. maybe problem in pubsubclient, maybe there is some settings need to be setted additionally.

1technophile:
Coudl you give more inputs, what are the settings you set  ?

superyarik:
I use cloudmqtt
https://www.cloudmqtt.com/docs.html

Port 13378 work perfect, but ssl port is notwork, client not connect.

1technophile:
ssl seems not supported by pubsubclient:
https://github.com/knolleary/pubsubclient/pull/251

superyarik:
yep, another clients(mqtt dashboards for android) work well.

look like this is another part of joke that "S" in "IoT" is Security :)

8666:
I think that SSL is possible on ESP but not on the Arduino (there are libraries). Arduino UNO does not have memory for whole SSL magic (https://github.com/256dpi/arduino-mqtt)

1technophile:
thanks for the link

Legion2:
For TLS we need a the current time as unix epoch, this could be hardcoded which is bad and will require to update the sketch from time to time. The better solution is to use the NTP protocol and let the esp request the time from a time server.

Legion2:
I found this library https://github.com/arduino-libraries/NTPClient which can be used.

Legion2:
I now have TLS working on ESP8266 with server certificate, but it does not work with the root ca certificate.

@1technophile thanks for sponsoring me 👍

1technophile:
> I now have TLS working on ESP8266 with server certificate

Good news! It's already a good step

> @1technophile thanks for sponsoring me 👍

With pleasure, you have made very good contributions to this project

Legion2:
Ok it now also works on ESP32 but there only the root ca certificate works and not the server certificate. I will create a PR in the next days.

Legion2:
The NTP library is not needed, ESP8266 and ESP32 have them build in and also the CA certificate works now (I don't know why, I did not change anything). Here is the PR #699

PR #696, opened 2020-07-09
Set the domain name as server of the PubSubClient

Legion2:
don't do a manual DNS lookup

I'm working on getting TLS connections to work #101.
I think this is required to enable TLS support as the hostname must be verified by the `WiFiClientSecure`.


1technophile:
> I'm working on getting TLS connections to work #101.

Great!

Issue #693, opened 2020-07-07
Send 40-bit rf signal with some bits encoded incorrectly


sleepwalkera:
**Describe the bug**
Use MQTTto433 to send a 40-bit rf signal, but some of the bits in the actual signal are encoded incorrectly.

**To Reproduce**
1.  Publish to `mosquitto_pub -h localhost -u mqtt -P mqtt -t "home/OpenMQTTGateway/commands/MQTTto433" -m '{"value":"1001010110101100101101010000010100010001","protocol":1,"length":40,"repeat":4}'`
2. In serial monitor `N: Received json : {"value":"1001010110101100101101010000010100010001","protocol":1,"length":40,"repeat":4}`
3. The signal actual send is “0000000010010101010101101110110000010001“.

**Expected behavior**
The signal actual send should be "1001010110101100101101010000010100010001".

**Screenshots**
![image](https://user-images.githubusercontent.com/1573639/86802423-9a5bcb80-c0a7-11ea-9a9b-8c0d7d239be9.png)

**Environment (please complete the following information):**
* OpenMQTTGateway version used (OpenMQTTGateway v0.9.4)
* Library version related to the problem you have (nodemcuv2-rf-libraries.zip comes with OpenMQTTGateway v0.9.4 release)
* Have tested with the basic examples SendDemo given with rc-switch, send 40-bit signal correctly.

**Additional context**
Also tried send "1111111111111111111111111111111111111111" and "0000000011111111111111111111111111111111", the signal actual send is “0000000011000111000111000111000111000111“.
![image](https://user-images.githubusercontent.com/1573639/86804675-b19bb880-c0a9-11ea-9f25-52737e954c01.png)

When I use decimal code
`mosquitto_pub -h localhost -u mqtt -P mqtt -t "home/OpenMQTTGateway/commands/MQTTto433" -m '{"value":642847671569,"protocol":1,"length":40,"repeat":4}'
Decimal 642847671569 which is binary 1001010110101100101101010000010100010001
The signal actual send is “0000000010010101010101101110110000010001“
It is more similar to the original code. Only the highest 8 bits are inconsistent.

![image](https://user-images.githubusercontent.com/1573639/86807866-d5143280-c0ac-11ea-9c84-c9b484a3920e.png)

sleepwalkera:
When I use decimal code
[...]

PR #698, opened 2020-07-10
[RF] handle 64 bits values (except for UNO)

1technophile:
#693

This PR enables to handle 64 bits values except for Arduino UNO.
Why limiting it on UNO, mainly to reduce the sketch size. To handle 64 bits values we need to activate `ARDUINOJSON_USE_LONG_LONG` and this activation generate a too big sketch for the Uno when using the RF gateway.
Added to that I found also a limit on arduino boards, the usage of ·´strtoull´ function is not available. This function is not a lot used in the code. Only with the ´simpleReceiving` method.
This method is currently enable by default but the question of its relevantness is asked below : 
https://community.openmqttgateway.com/t/vote-removing-simplereceiving-and-simplepublishing/1044
simpleReceiving and simplePublishing enables to receive and send with a lite payload (no json), it was the first data format supported by OpenMQTTGateway. 
Now I suspect that json is the mostly i used format.
We could develop an ´strtoull´ for Arduino but before doing that I prefer to know the simpleReceiving usage.


1technophile:
@Legion2 As the [poll ](https://community.openmqttgateway.com/t/vote-removing-simplereceiving-and-simplepublishing/1044/2) goes into a little usage of simpleReceiving versus json I think this PR may be merged like that, minus your comments/suggestions.

Another PR would remove the simpleReceiving function when the poll is confirmed.

main/main.ino
```diff
@@ -1348,15 +1357,9 @@ void receivingMQTT(char* topicOri, char* datacallback) {
...
-#ifdef jsonPublishing
-    if (jsondata.success())
-      data = jsondata["value"];
-#endif
+    TYPE_UL_ULL data = 0;
 
-#ifdef simplePublishing
-    data = strtoul(datacallback, NULL, 10); // we will not be able to pass values > 4294967295
-#endif
+    jsondata.success() ? data = jsondata["value"] : data = STRTO_UL_ULL(datacallback, NULL, 10);
```
Legion2, line 1362:
```suggestion
    data = jsondata.success() ? jsondata["value"] : STRTO_UL_ULL(datacallback, NULL, 10);
```
Does that also work?

main/main.ino
```diff
@@ -30,14 +30,17 @@
 // array to store previous received RFs, IRs codes and their timestamps
 #if defined(ESP8266) || defined(ESP32) || defined(__AVR_ATmega2560__) || defined(__AVR_ATmega1280__)
 #  define array_size 12
-unsigned long ReceivedSignal[array_size][2] = {{0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}};
+unsigned long long ReceivedSignal[array_size][2] = {{0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}, {0, 0}};
```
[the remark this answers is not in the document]
1technophile: Due to the fact that there is already a condition on boards here, I can use it anyway if you think it is better

main/main.ino
1technophile: Thanks

main/main.ino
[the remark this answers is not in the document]
Legion2: This also relates to the general question "How do we decided where to use the TYPE_UL_ULL type?". As I see it, we introduced this type macro for the handling of 64 bit codes in the RF and IR modules. This is fine, but then we should only use this macro for such codes and use this macro everywhere we handle such codes.

The `storeValue` and `isAduplicate` functions are in the `main.ino` and therefore provided as a feature to all modules.  These functionality is strongly related to the two modules and I think should not be in the main.ino. Else they should not expose the type of the RF and IR codes. The functions should at least document that they are related to RF and IR dedup and should not be used otherwise, then I think it's fine to use the macro here. I then would also rename the macro to make it clear it represent the size of the RF and IR codes.

The type of `ReceivedSignal` should be changed, it represents the codes in the first array and the times in the second, but the times are only `unsigned long` as you can see in the `isAduplicate` function. So the 2d array is not the best data structure to represent this type of data.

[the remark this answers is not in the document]
1technophile: Thanks for the comments.

> This is fine, but then we should only use this macro for such codes.

`value` key is used by IR, RF and SRFB gateways
`storeValue` and `isAduplicate` functions are used in `ZgatewayWeatherStation` with 32 bits value, `ZgatewaySRFB` with 32 bits value, `ZgatewayIR` with 32 or 64 bits and `ZgatewayRF` with 32 or 64 bits 

I took the hypothesis that a 64bits variable can handle a 32bits ones for `ZgatewaySRFB` .
So the functions in `main.ino`  should be able to handle the 32 bits values.

Maybe the only issue is here:
`if (p.value.is<TYPE_UL_ULL>()`

But we are in `simplePublishing `scope, so will be removed soon and not a lot used.

>The storeValue and isAduplicate functions are in the main.ino and therefore provided as a feature to all modules. These functionality is strongly related to the two modules and I think should not be in the main.ino. Else they should not expose the type of the RF and IR codes. The functions should at least document that they are related to RF and IR dedup and should not be used otherwise, then I think it's fine to use the macro here. I then would also rename the macro to make it clear it represent the size of the RF and IR codes.

ok, I will modify the code

> The type of ReceivedSignal should be changed, it represents the codes in the first array and the times in the second, but the times are only unsigned long as you can see in the isAduplicate function. So the 2d array is not the best data structure to represent this type of data.

ok, I will change the data structure

main/main.ino
```diff
@@ -27,17 +27,29 @@
 */
 #include "User_config.h"
 
+// Macros and structure to enable the duplicates removing on the following gateways
+#if defined(ZgatewayRF) || defined(ZgatewayIR) || defined(ZgatewaySRFB) || defined(ZgatewaySRFB) || defined(ZgatewayWeatherStation)
+// variables to avoid duplicates
+#  define time_avoid_duplicate 3000 // if you want to avoid duplicate mqtt message received set this to > 0, the value is the time in milliseconds during which we don't publish duplicates
```
Legion2, line 33: should this be part of the User_config.h?

main/main.ino
```diff
@@ -1300,67 +1318,65 @@ void stateMeasures() {
 }
 #endif
 
-void storeValue(unsigned long MQTTvalue) {
+#if defined(ZgatewayRF) || defined(ZgatewayIR) || defined(ZgatewaySRFB) || defined(ZgatewaySRFB) || defined(ZgatewayWeatherStation)
+void storeSignalValue(SIGNAL_SIZE_UL_ULL MQTTvalue) { // Store signal values from RF, IR, SRFB or Weather stations so as to avoid duplicates
```
Legion2, line 1322: put the comment in front of the method as doc-comment

main/main.ino
```diff
@@ -1300,67 +1318,65 @@ void stateMeasures() {
...
+    if (receivedSignal[i].time < minimum) {
+      minimum = receivedSignal[i].time;
       minindex = i;
     }
   }
   return minindex;
 }
 
-bool isAduplicate(unsigned long value) {
+bool isAduplicateSignal(SIGNAL_SIZE_UL_ULL value) { // Check if signal values from RF, IR, SRFB or Weather stations are duplicates
```
Legion2, line 1349: here also put the documentation in the line before the method

main/main.ino
```diff
@@ -1319,7 +1317,10 @@ void stateMeasures() {
 #endif
 
 #if defined(ZgatewayRF) || defined(ZgatewayIR) || defined(ZgatewaySRFB) || defined(ZgatewaySRFB) || defined(ZgatewayWeatherStation)
-void storeSignalValue(SIGNAL_SIZE_UL_ULL MQTTvalue) { // Store signal values from RF, IR, SRFB or Weather stations so as to avoid duplicates
+/** 
+* Store signal values from RF, IR, SRFB or Weather stations so as to avoid duplicates
+*/
```
Legion2, line 1322: normally there should be a whitespace before the `*`
```suggestion
/** 
 * Store signal values from RF, IR, SRFB or Weather stations so as to avoid duplicates
 */
```

1technophile:
@Legion2 Thanks for the advices on this PR!

main/main.ino
```diff
@@ -1300,67 +1316,71 @@ void stateMeasures() {
...
       minindex = i;
     }
   }
   return minindex;
 }
 
-bool isAduplicate(unsigned long value) {
+/** 
+* Check if signal values from RF, IR, SRFB or Weather stations are duplicates
+*/
```
Legion2, line 1355: ;)
```suggestion
/** 
 * Check if signal values from RF, IR, SRFB or Weather stations are duplicates
 */
```

1technophile: Not enough scrolled ;-)

PR #699, opened 2020-07-14
added support for secure TLS connections

### LINKED ISSUE #101: add ssl support to connect to mqtt broker

Legion2:
support TLS on ESP32 and ESP8266 close #101

TODO:
- [x] documentation
- [x] tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .pio
 .vscode
+*_env.ini
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -63,6 +63,7 @@ module.exports = {
             'upload/pio',
             'upload/arduino-ide',
             'upload/portal',
+            'upload/advanced-configuration',
             'upload/troubleshoot'
           ]
         },
--- a/docs/integrate/home_assistant.md
+++ b/docs/integrate/home_assistant.md
@@ -1,6 +1,6 @@
 # Integrate Home Assistant
 ## Auto discovery
-So as to enable HASS auto discovery with MQTT you have to uncomment [ZmqttDiscovery](https://github.com/1technophile/OpenMQTTGateway/blob/0180a0dbd55ed8e0799e30ee84f68070a6f478fa/User_config.h#L99) in user_config.h
+So as to enable HASS auto discovery with MQTT you have to uncomment [ZmqttDiscovery](https://github.com/1technophile/OpenMQTTGateway/blob/0180a0dbd55ed8e0799e30ee84f68070a6f478fa/User_config.h#L99) in User_config.h
 And enable discovery on your MQTT integration definition in HASS.
 
 OMG will use the auto discovery functionality of home assistant to create sensors and gateways into your HASS instance automaticaly.
--- a/docs/upload/advanced-configuration.md
+++ b/docs/upload/advanced-configuration.md
@@ -28,4 +28,7 @@ const char* certificate CERT_ATTRIBUTE = R"EOF("
 ")EOF";
 ```
 
+If you have no ntp server in your local network (included in the router) or not using dhcp, you should uncomment `//#    define NTP_SERVER "pool.ntp.org"` to use a ntp server for time synchronization.
+This is related to the `W: failed, ssl error code=54` error message, which indicate that the time of the esp is not correct.
+
 You can know compile and upload to your board and the gateway should connect with TLS to your broker.
--- a/docs/upload/advanced-configuration.md
+++ b/docs/upload/advanced-configuration.md
@@ -1,0 +1,31 @@
+# Advanced configuration of the gateway
+
+## Secure connection to the broker
+By default the connection between the gateway and the MQTT broker is not encrypted and the identity of the broker is not verified.
+For connections in a private local area network this might not be a big issue, but if you connect to a broker on the internet or the gateway is connected to a public network there is a high security risk.
+
+To secure the connection, Transport Layer Security (TLS) can be used which is supported by most MQTT brokers.
+Setting up and configuring TLS in the MQTT broker is a complex process and perhaps include creating a self-signed certificate.
+The configuration of the broker is not covered here, you should look into the documentation of your broker.
+
+### Prerequisites
+The MQTT broker is configured for TLS and you have access to the CA certificate which was used to sign the MQTT broker certificate.
+You are using ESP8266 or ESP32, for other boards TLS is not supported.
+
+### Configure secure connection in the gateway
+To enable the secure connection and use TLS uncomment `//#define SECURE_CONNECTION` in `User_config.h`.
+Set `MQTT_SERVER` to the Common Name (CN) of the certificate of the broker.
+This can be the hostname or the ip of the broker.
+
+The CA certificate should be in PEM ascii format.
+If your CA certificate is not in the correct format or you don't know the format, use `openssl` to convert the certificate to the correct format.
+In `User_config.h` replace the `...` with the content of your certificate which is between the `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----` lines:
+```cpp
+const char* certificate CERT_ATTRIBUTE = R"EOF("
+-----BEGIN CERTIFICATE-----
+...
+-----END CERTIFICATE-----
+")EOF";
+```
+
+You can know compile and upload to your board and the gateway should connect with TLS to your broker.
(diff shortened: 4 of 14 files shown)


1technophile:
Thanks for the PR, if you need, I would be able to test it next week.

Legion2:
I already did manual tests, but I want it to be "tested" in the ci.

Legion2:
@1technophile should we keep the `Test_config.h` in sync with the `User_config.h`?

1technophile:
> @1technophile should we keep the `Test_config.h` in sync with the `User_config.h`?

We should remove it, while we don't test Arduino IDE with the CI we don't need it anymore

Legion2:
how we then "test"(compile) the TLS code in the CI pipeline?

1technophile:
If you need `Test_config.h` to store a test certificate you can keep it.
Do you need other things to test this function?

Legion2:
For just compiling we don't need a real certificat, and for manual testing I use my own ca from my MQTT broker.

Legion2:
@1technophile Were in the docs should I put the documentation for the TLS MQTT connection? Is there a section for the MQTT broker connection setup, I did not found any.

1technophile:
@Legion2 maybe we should add an advanced_configuration.md in the beginning of the upload section?

Legion2:
An advanced configuration section sounds good.

The tests fail on ESP32 because there is no more free Flash when compiling the esp32dev-all example:
```
RAM:   [==        ]  22.0% (used 72156 bytes from 327680 bytes)
Flash: [==========]  104.9% (used 2062180 bytes from 1966080 bytes)
Error: The program size (2062180 bytes) is greater than maximum allowed (1966080 bytes)
```

1technophile:
I don't think it is a big issue, as `[env:esp32dev-all]` is more a test environment, maybe you should add a test env dedicated to the certificate testing?

I think also that we should differentiate test and production env and do not upload the binaries to github for the test ones.

Legion2:
Ok good idea to make special test environments, @1technophile which of the current env do you consider to be only test related?
We can give them a `test-` prefix so we can filter them before upload.

1technophile:
Here they are:
```
esp32dev-all
nodemcuv2-all
atmega-all
test-manual-wifi
```

We should keep on extracting the libraries from these environments as it enables to have all the libraries in one zip file. But their binaries are in my point of view not necessary and can mislead the user.

Legion2:
why does the `atmega-all` env not have a .bin file uploaded?

image

1technophile:
I have never uploaded a BIN file directly on an Arduino Mega neither searched how to do it.

Legion2:
oh ok

Legion2:
currently we use the name of the env for the name of the uploaded files. So if we add `test-` also the library zip will start with `test-`.

1technophile:
Maybe we should do it the other way and don't publish the binaries that end with `-all`?

Legion2:
but then other `test-` binaries are still published

1technophile:
Like that?
*-all we publish libraries and not binaries
*-test we don't publish
others we publish libraries and binaries

Legion2:
@1technophile I create a new pull request for this change

Legion2:
@1technophile how do I add a new page to the docs?

1technophile:
You create a new .md page in the wished directory and you add the directory/page_name in the [config.js](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/.vuepress/config.js) for the docs (sidebar section)

Legion2:
> W: failed, ssl error code=54

mean "Certificate is expired or not yet valid." this can be due to NTP does not set the time correctly. Can you confirm that the time is set correctly, this maybe take a while and it should be eventually set correctly. The normal connection retry should then successfully connect with the correct time set.

1technophile:
> mean "Certificate is expired or not yet valid." this can be due to NTP does not set the time correctly. Can you confirm that the time is set correctly, this maybe take a while and it should be eventually set correctly. The normal connection retry should then successfully connect with the correct time set.

On target, I just uncommented `#    define NTP_SERVER "pool.ntp.org"` and now the ESP8266 is connected by TLS.
For info it can take several attempts before connecting as you mentionned.

Sumup
|Board|Modules or env|Result|Comments|
|-|-|-|-|
|barebone ESP32| no| OK| |
|barebone ESP32| IR & RF| OK| |
|m5stickc| esp32-m5stick-c-ble| -| CoreDump after BLE Scan begin|
|barebone ESP32| esp32dev-ble| -| CoreDump after BLE Scan begin|
|nodemcuv2| RF, BT, DHT, ONOFF, ADC, GPIOInput| OK|-|
|nodemcuv2| RF| OK|- |

Should we add something on the [troubleshooting ](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/upload/troubleshoot.md) doc about `W: failed, ssl error code=54`

Legion2:
> Should we add something on the troubleshooting doc about W: failed, ssl error code=54 ?

yes I will also add `# define NTP_SERVER "pool.ntp.org"` to the docs

Legion2:
can you decode the core dump so we know in which function the error happens

1technophile:
> can you decode the core dump so we know in which function the error happens

Never done that, but there is always a first time! Let's try

1technophile:
Here is the decoding of the stack trace:

```
PC: 0x4000c46c
EXCVADDR: 0x00000000

Decoding stack results
0x401bf165: bta_sys_init at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/bt/bluedroid/bta/sys/bta_sys_main.c line 170
0x40194591: btu_task_start_up at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/bt/bluedroid/stack/btu/btu_task.c line 284
0x401945cb: btu_task_thread_handler at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/bt/bluedroid/stack/btu/btu_task.c line 226
0x4008eb71: vPortTaskWrapper at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/freertos/port.c line 143
```

1technophile:
Seems that we are not alone on this:
https://www.gitmemory.com/issue/nkolban/esp32-snippets/842/476151941

1technophile:
As a complementary info, I also tried to change the partition type to `board_build.partitions = huge_app.csv` with:
`esp32dev-ble`

I'm getting:
```
RAM:   [==        ]  20.7% (used 67924 bytes from 327680 bytes)
Flash: [======    ]  60.1% (used 1891856 bytes from 3145728 bytes)
```

And still the core dump when the scan begin:
```
N: Changing LOW POWER mode to: 0
N: Scan begin
Guru Meditation Error: Core  0 panic'ed (StoreProhibited). Exception was unhandled.
Core 0 register dump:
PC      : 0x4000c46c  PS      : 0x00060f30  A0      : 0x801b47f0  A1      : 0x3fff7830
A2      : 0x00000000  A3      : 0x00000000  A4      : 0x00000130  A5      : 0x00000000
A6      : 0x3f426428  A7      : 0x00000013  A8      : 0x80085039  A9      : 0x3fff77d0
A10     : 0x00000000  A11     : 0x00001800  A12     : 0x00000000  A13     : 0x00000003
A14     : 0x00000001  A15     : 0x00000006  SAR     : 0x00000008  EXCCAUSE: 0x0000001d
EXCVADDR: 0x00000000  LBEG    : 0x4000c46c  LEND    : 0x4000c477  LCOUNT  : 0x00000012

Backtrace: 0x4000c46c:0x3fff7830 0x401b47ed:0x3fff7840 0x40188b29:0x3fff7860 0x40188b63:0x3fff7880 0x4008eb71:0x3fff78b0
```

Legion2:
yeah, from what I read, both (TLS and BLE) allocate huge amount of memory and so they can't be used together. We would have to do many memory usage optimizations to get both working at the same time.

1technophile:
Some [ideas ](https://github.com/espressif/arduino-esp32/issues/2175 )here also.

I'm going to do some tests this week by changing different parameters and having a closer memory followup.

In all the case ESP-IDF [will support ](https://github.com/espressif/esp-idf/releases/tag/v4.1-beta2)NimBLE; a lighter BLE stack compared to BlueDroid, so as a last option we may add a note in the documentation about the current incompatibility and wait for the NimBLE integration into [Arduino ESP32](https://github.com/espressif/arduino-esp32).

1technophile:
A better solution, I have found this library:
https://github.com/h2zero/NimBLE-Arduino
Which consumes less memory compared to Bluedroid, so I propose to merge this PR and we will integrate the lib futher.

Legion2:
Ok

PR #701, opened 2020-07-18
Rf receive only

jmw6773:
@1technophile , 
  I'm not sure if this is useful for many, but I ran out of pins on my ESP8266 and needed one more. The RF library has the option of disable transmission, which would free up the pin that would be used for the transmitter. I only use my gateway to receive RF signals, so don't use a transmitter.

  I've added that if `RF_DISABLE_TRANSMIT` is defined (with any value), then transmit is disabled. This DEFINE isn't added to the config_RF.h file. I've added a short section to the docs.

  Please feel free to reject this pull request if you don't want to add this functionality. (Could probably be added to IR as well.)


jmw6773:
Sorry for the close/re-open of the pull request. I forgot to rebase my fork.

1technophile:
No problem, let me know when you think it is ready.

jmw6773:
@1technophile ,
  It should be ready now. I closed it before to rebase, then reopened the pull request.

-Jon

main/ZgatewayRF.ino
```diff
@@ -63,7 +63,11 @@ void setupRF() {
   ELECHOUSE_cc1101.setMHZ(CC1101_FREQUENCY);
   ELECHOUSE_cc1101.SetRx(CC1101_FREQUENCY);
 #  endif
+#  ifdef RF_DISABLE_TRANSMIT
+  mySwitch.disableTransmit();
+#  else
   mySwitch.enableTransmit(RF_EMITTER_GPIO);
+#  endif
   mySwitch.setRepeatTransmit(RF_EMITTER_REPEAT);
```
1technophile, line 71: Should we include the `mySwitch.setRepeatTransmit(RF_EMITTER_REPEAT);` into the `RF_DISABLE_TRANSMIT` scope ?

jmw6773: It doesn't matter. The `disableTransmit` function sets the transmission PIN to `-1`. 
https://github.com/1technophile/rc-switch/blob/385a7e065f225c159bcd0d1b6278495423a1bd23/src/RCSwitch.cpp#L227-L229

The RCSwitch::send() function then checks if the transmission PIN is `-1` and returns if it is. 
https://github.com/1technophile/rc-switch/blob/385a7e065f225c159bcd0d1b6278495423a1bd23/src/RCSwitch.cpp#L554-L556

Setting the repeat value doesn't change the transmission PIN, so the transmit function stays disabled.

1technophile:
Could you add the macro example into config_RF.h, commented by default?

PR #702, opened 2020-07-18
PIR Notify LED

jmw6773:
Many motion sensors have an LED that turns on when motion is detected.
This can be for security, or used as a 'nightlight' instead of turning on a main light.

I've added functionality so that the PIR value will be mirrored to another PIN that can be used to drive an LED, or a MOSFET if needed. (In the case of the ESP32-CAM board, there is an on-board LED used for the camera flash that can be used.)
This functionality is disabled by default, but can be used by defining a PIN number for `HCSR501_LED_NOTIFY_PIN`.

`#define HCSR501_LED_NOTIFY_PIN 4`

I also added a section in the docs for the HCSR501 sensor.

@1technophile,
  Please feel free to reject this pull request if this isn't something your wanting for the OpenMQTTGateway project.
  I found the functionality useful, but these 'hidden' functions might not be useful to many.

-Jon


docs/use/sensors.md
```diff
@@ -36,4 +36,15 @@ Each sensor will be published under the following topic using each sensors' addr
 
 The units for temperature readings are sent in Celcius by default can be changed to ferenheight by setting DS1820_FAHRENHEIT = true in in config_DS1820.h
 
-If you don't want to resend values that haven't changed you can set DS1820_ALWAYS = false in config_DS1820.h
\ No newline at end of file
+If you don't want to resend values that haven't changed you can set DS1820_ALWAYS = false in config_DS1820.h
+
+
+### HCSR501
+The a boolean value of the PIR sensor when a state change occurs. The length of time that the PIR stays triggered depends on the PIR hardware and is not changed by OpenMQTTGateway.
```
1technophile, line 43: small typo

main/ZsensorHCSR501.ino
```diff
@@ -33,6 +33,10 @@
 void setupHCSR501() {
   Log.notice(F("HCSR501 pin: %d" CR), HCSR501_GPIO);
   pinMode(HCSR501_GPIO, INPUT); // declare HC SR-501 GPIO as input
```
1technophile, line 35: Could you add also the macros to the corresponding config_ files ?

main/ZsensorHCSR501.ino
```diff
@@ -33,6 +33,10 @@
...
+#  ifdef HCSR501_LED_NOTIFY_PIN
+  pinMode(HCSR501_LED_NOTIFY_PIN, OUTPUT);
```
1technophile, line 37: also this macro should be added to the corresponding config file

docs/use/sensors.md
jmw6773: Nice catch. I've reworded this section.

main/ZsensorHCSR501.ino
jmw6773: I added an example in the config_.h file (commented out)
`//#define HCSR501_LED_NOTIFY_GPIO 4 //Uncomment this line to mirror the state of the PIR sensor to the specified GPIO`

main/ZsensorHCSR501.ino
jmw6773: I'm not sure what you're wanting done here. `HCSR501_GPIO` is already defined in the config_.h file.

jmw6773:
I also changed the define name used from `HCSR501_LED_NOTIFY_PIN` to `HCSR501_LED_NOTIFY_GPIO` as all other PIN definitions are using the GPIO name convention.

1technophile:
Thanks for the PR and the modifications.

Issue #690, opened 2020-07-05
MQTT Discovery value key definitions


jmw6773:
**Describe the bug**
  The value key definitions for temperature (BME, HTU, AHT) and presence (HCSR501 sensor) do not align with the keys that these sensors use. This causes the mqttDiscovery to fail for these devices. (It is a partial fail for the temperature sensor as the 'tempf' and 'hum' works as is.)

**Temperature**
https://github.com/1technophile/OpenMQTTGateway/blob/0e257a58c137124c81967086970353640ec411c8/main/config_mqttDiscovery.h#L80

**PIR**
https://github.com/1technophile/OpenMQTTGateway/blob/0e257a58c137124c81967086970353640ec411c8/main/config_mqttDiscovery.h#L83

**Expected behavior**
The following value keys work for the (BME, HTU, AHT) temperature sensors and the HCSR501.
```
#  define jsonTemp     "{{ value_json.tempc | is_defined }}"
#  define jsonPresence "{{ value_json.hcsr501 | is_defined }}"
```

The DHT sensor would require the following (or a change to the ZsensorDHT.ino file to change the published JSON key to match other keys).
```
#  define jsonTemp     "{{ value_json.temp | is_defined }}"
```

jmw6773:
The Presence is used for Bluetooth, so this key shouldn't be changed.

https://github.com/1technophile/OpenMQTTGateway/blob/0e257a58c137124c81967086970353640ec411c8/main/config_mqttDiscovery.h#L83

Instead the HCSR501 should update it's key to use the 'presence' key for its data value instead of 'hcsr501'.
https://github.com/1technophile/OpenMQTTGateway/blob/0e257a58c137124c81967086970353640ec411c8/main/ZsensorHCSR501.ino#L51-L57

1technophile:
Thanks for pointing this.

Currently we have modules pushing celsius temperature with the `tem` and `tempc` keys, we should choose one of them and apply the change in the code I think.  Another possibility could be to consider the use of `tempc` and `tempf ` keys for the modules that send both and `tem` for the modules that send only celsius.
Unfortunately this will be a breaking change, so we have to take this into account also.

I agree for the hcsr501.

jmw6773:
The DHT module is pushing `temp` as its key, meaning we have at least three sensors using different keys.

[...]

PR #703, opened 2020-07-18
Standardize MQTTDiscovery Temp Keys [deprecate tem key]

jmw6773:
**This change will eventually break current deployments.**
As discussed, the old `value_json.tem` key will remain for the upcoming release of OMG, but will be removed after the following release.

From issue #690, I'd like to transmit Celsius and Fahrenheit values for all temperature sensors. When the specific sensor library does not support temperature conversions a MACRO has been added to the user_config.h file to convert between C and F. (Also added to Test_config.h)

These changes need to be tested for the following sensors:
- All Bluetooth devices (I do not own any of the Bluetooth devices that have temperature output)
- WeatherStation - I do not own this hardware
- BME280  - I have this sensor and can test
- HTU21    - I have this sensor and can test
- DS1820  - I have this sensor and can test
- DHT        - I have this sensor and can test
- AHT10    - Tested and working

I am able to test the environment sensors, but would need help testing the Bluetooth and WeatherStation equipment.


jmw6773:
@1technophile , 
  If you could review these changes and confirm that this is what we discussed, I'd appreciate it.

  I'm also finding that I am unable to set the temperature unit in mqttDiscovery to `°F`. It transmits correctly, but HomeAssistant doesn't read it properly. (The sensor reads the `°C` value.) If I change the unit to `F` or `° F`, everything works okay.

  This isn't a bug in OMG, but if we remove the `°` from temperature units, it may help. (No issues with the `°C` sensor. Maybe due to the ordering it defaults to the first value in the JSON message?)

https://github.com/1technophile/OpenMQTTGateway/blob/24c8c2501df73c588c57fcae01689ae824c154ae/main/ZmqttDiscovery.ino#L194

main/User_config.h
```diff
@@ -287,4 +287,7 @@ uint8_t wifiProtocol = 0; // default mode, automatic selection
 /*-------------------DEFINE LOG LEVEL----------------------*/
 #define LOG_LEVEL LOG_LEVEL_NOTICE
 
+#define convertTemp_CtoF(c) ((c * 1.8) + 32)
+#define convertTemp_FtoC(f) ((f - 32) * 5/9)
```
Legion2, line 291: why are these macros part of the user_config.h and not in the main.ino, they are not configurable? Also what would be the different when using a function instead of a macro?

main/ZgatewayWeatherStation.ino
```diff
@@ -105,7 +105,9 @@ void sendTemperatureData(byte id, float temperature, int humidity, byte battery_
     StaticJsonBuffer<JSON_MSG_CALC_BUFFER> jsonBuffer;
     JsonObject& RFdata = jsonBuffer.createObject();
     RFdata.set("sensor", id);
-    RFdata.set("temperature", temperature);
+    RFdata.set("temperature", temperature); // remove for 0.9.6 release
+    RFdata.set("tempc", temperature);
+    RFdata.set("tempf", wsdr.convertCtoF(temperature));
```
Legion2, line 110: why is here another conversion function used?

jmw6773: When a library supports its own conversion method, I used it. Otherwise I used the conversion macros in user_config.h

main/User_config.h
jmw6773: I was thinking that `user_config.h` would be more accessible to all of the sensors, but you're right; it should be in `main.ino` instead.

The main benefit in using a macro instead of a function is that the macro doesn't need have the variable type defined. Meaning it doesn't matter if an int, double, or float called with the macro. (Otherwise, there would need to be a function defined for each type.)
From my understanding, this is because the compiler basically copies the code block to be inline with where it's being called from.
BT devices are using 'double' and AHT is using 'float' values.

jmw6773: I moved it to `main.ino` and pushed the change.

Legion2:
@jmw6773 if you also think this is ready to be merged just press the "Ready for review" button.

jmw6773:
@legion2, I'm planning to test the BME280, HTU21, DS1820, and DHT sensors tonight.
I don't have any way to test the WeatherStation or Bluetooth devices though.

I'll also need to push a change to remove the `°` symbol from the units. It seems break the sensor from reading that value in Home Assistant.

jmw6773:
@Legion2,
  Tested with the BME280, HTU21, AHT, DS1820, and DHT sensors.

-   I've removed the `°` symbol in all MQTTDiscovery packets
-   Found a mistake when publishing the DHT MQTTDiscovery packet
-   Added the `tempc` key to all BT sensors (with temperature). The legacy `tem` key should be removed for release 0.9.6

I believe that I rebased this properly this time, but please let me know if I screwed it up again.

Legion2:
> I believe that I rebased this properly this time, but please let me know if I screwed it up again.

yeah looks better, but you didn't used the current `1technophile/development` branch for the rebase, you used your merged version. You need to [configure the upstream repository in git](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/configuring-a-remote-for-a-fork) and then use the upstream for the rebase.

jmw6773:
@Legion2 - Rebased again using the upstream. Did I finally get it right? Looking at the commit history, I'm not seeing duplicate commits or anything obvious that I messed up again.

jmw6773:
I'm unable to test with Bluetooth devices and the WeatherStation, but other environment sensors are reporting values as expected.

If there any place that needs to be noted in docs that old temp keys will be removed in a future release?

1technophile:
>I'm unable to test with Bluetooth devices and the WeatherStation, but other environment sensors are reporting values as expected.

I'm confident on this. In all the case I do a full test campaign before each release.

>If there any place that needs to be noted in docs that old temp keys will be removed in a future release?

I will add it in the v0.9.5 release note when publishing it.

jmw6773:
@1technophile , it's okay to merge. Thank you.

1technophile:
Thanks for the PR!

PR #704, opened 2020-07-21
Fix ESPWifiManualSetup / Move MQTT variables

jmw6773:
I think I started to hit the limit of available RAM in my esp8266's, so I tried setting up ESPWifiManualSetup to save a few Kb's.
In doing so, I found that it wasn't working as expected... or at all.

In the User_config.h `wifi_ssid` and `wifi_password` are set using `#define`, where the `Wifi.begin()` function requires a char[]. An error wasn't being thrown during compile or at runtime in the Serial monitor, but WiFi never connects.

I left the #define's in `User_config.h`, just adding checks to see if they're already set, so they can be set via platformio.ini, and added 2 char[] variables in `main.ino` and assigned them the `User_config.h` values. Doing it this way seemed the least disruptive to users

------

I've also moved the MQTT config variables from the `User_config.h` file into the `main.ino` file and added `#define` values to replace the original default values. This will allow users to set the values in the `platformio.ini` file for easier custom/advanced builds.

### DIFF
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -305,6 +305,6 @@ uint8_t wifiProtocol = 0; // default mode, automatic selection
 #define subjectMQTTtoSYSset          "/commands/MQTTtoSYS/config"
 
 /*-------------------DEFINE LOG LEVEL----------------------*/
-#define LOG_LEVEL LOG_LEVEL_TRACE
+#define LOG_LEVEL LOG_LEVEL_NOTICE
 
 #endif
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -105,12 +105,11 @@ const byte mac[] = {0xDE, 0xED, 0xBA, 0xFE, 0x54, 0x95}; //W5100 ethernet shield
 #  define mqtt_topic_max_size  50
 #  define mqtt_max_packet_size 128
 #endif
-char mqtt_user[parameters_size] = "your_username"; // not compulsory only if your broker needs authentication
-char mqtt_pass[parameters_size] = "your_password"; // not compulsory only if your broker needs authentication
-char mqtt_server[parameters_size] = "192.168.1.17";
-char mqtt_port[6] = "1883";
-char mqtt_topic[mqtt_topic_max_size] = Base_Topic;
-char gateway_name[parameters_size * 2] = Gateway_Name;
+
+#define MQTT_USER_DEFAULT "your_username"
+#define MQTT_PASS_DEFAULT "your_password"
+#define MQTT_MQTT_DEFAULT "192.168.1.17"
+#define MQTT_PORT_DEFAULT "1883"
 
 #if defined(ESP8266) || defined(ESP32)
 #  define ATTEMPTS_BEFORE_BG 10 // Number of wifi connection attempts before going to BG protocol
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -84,7 +84,6 @@ const byte mac[] = {0xDE, 0xED, 0xBA, 0xFE, 0x54, 0x95}; //W5100 ethernet shield
 #  endif
 #endif
 
-
 #define WifiManager_password            "your_password" //this is going to be the WPA2-PSK password for the initial setup access point
 #define WifiManager_ssid                Gateway_Name //this is the network name of the initial setup access point
 #define WifiManager_ConfigPortalTimeOut 120
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -76,9 +76,14 @@ const byte mac[] = {0xDE, 0xED, 0xBA, 0xFE, 0x54, 0x95}; //W5100 ethernet shield
 #endif
 
 #if defined(ESPWifiManualSetup) // for nodemcu, weemos and esp8266
-#  define wifi_ssid     "wifi ssid"
-#  define wifi_password "wifi password"
+#  ifndef wifi_ssid
+#    define wifi_ssid    "wifi ssid"
+#  endif
+#  ifndef wifi_password
+#    define wifi_password "wifi password"
+#  endif
 #endif
+
 
 #define WifiManager_password            "your_password" //this is going to be the WPA2-PSK password for the initial setup access point
 #define WifiManager_ssid                Gateway_Name //this is the network name of the initial setup access point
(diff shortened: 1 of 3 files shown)


main/User_config.h
```diff
@@ -285,6 +296,6 @@ uint8_t wifiProtocol = 0; // default mode, automatic selection
 #define subjectMQTTtoSYSset          "/commands/MQTTtoSYS/config"
 
 /*-------------------DEFINE LOG LEVEL----------------------*/
-#define LOG_LEVEL LOG_LEVEL_NOTICE
+#define LOG_LEVEL LOG_LEVEL_TRACE
```
1technophile, line 299: Hello Jon,
Could you put it back to NOTICE level please

jmw6773: Oops. Re-pushing now

1technophile:
Good idea!
@Legion2 is it ok for you?

main/User_config.h
```diff
@@ -105,12 +109,19 @@ const byte mac[] = {0xDE, 0xED, 0xBA, 0xFE, 0x54, 0x95}; //W5100 ethernet shield
...
+#endif
+#ifndef MQTT_PASS_DEFAULT
+#  define MQTT_PASS_DEFAULT "your_password"
+#endif
+#ifndef MQTT_MQTT_DEFAULT
+#  define MQTT_MQTT_DEFAULT "192.168.1.17"
+#endif
+#ifndef MQTT_PORT_DEFAULT
+#  define MQTT_PORT_DEFAULT "1883"
+#endif
```
Legion2, line 124: I would not name these _DEFAULT

jmw6773: Any naming recommendation? As C language is case sensitive, I could use names like `MQTT_USER` without the `_DEFAULT`, but thought it might be confusing with the corresponding `char[]` created in `main.ino` with the lowercase name `mqtt_user`. 

Also, should each variable check for if it's already defined, or should it be grouped into two logical blocks? Meaning, should we require the user to set not set any values, or set them all.

```
if !defined(MQTT_USER_DEFAULT) || !defined(MQTT_PASS_DEFAULT)
#  define MQTT_USER_DEFAULT "your_username"
#  define MQTT_PASS_DEFAULT "your_password"
#endif

if !defined(MQTT_MQTT_DEFAULT) || !defined(MQTT_PORT_DEFAULT)
#  define MQTT_MQTT_DEFAULT "192.168.1.17"
#  define MQTT_PORT_DEFAULT "1883"
#endif
```

Legion2: The internal name `user_name` in the main.ino file is not visible by the user, the users should only look into the config. And for developer it should be clear that one is a macro and the other a variable. Also the macro is only used at one place, so only user_name should be used in the code.

I would not group the definitions, as you probably will not change the default mqtt port but the ip/hostname.

Btw. the `mqtt_server_name` should also be handled like the other variables, maybe we can merge it with the `mqtt_server` variable, as they are handled the same (I could currently enter the hostname in mqtt_server and it would work). @1technophile We also have to think about the string length of the parameters, as they might not be long enough. FYI, the password length of 20 is not enough for my generated passwords.

main/main.ino
```diff
@@ -123,6 +123,18 @@ unsigned long ReceivedSignal[array_size][2] = {{0, 0}, {0, 0}, {0, 0}, {0, 0}};
 //adding this to bypass the problem of the arduino builder issue 50
 void callback(char* topic, byte* payload, unsigned int length);
 
+#ifdef ESPWifiManualSetup
+char manual_wifi_ssid[] = wifi_ssid;
+char manual_wifi_password[] = wifi_password;
+#endif
```
Legion2, line 129: to reduce the memory footprint we can move this into `setup_wifi`

main/User_config.h
jmw6773: I've removed the `_DEFAULT` postfix and renamed `MQTT_MQTT` to be `MQTT_SERVER` so it's more descriptive as to its purpose.

I agree that `mqtt_server_name` is redundant as the PubSubClient library handles DNS and IP addresses the same and both can be configured using the same config location.
I've removed the `mqtt_server_name` definition the User_config.h and the reduced the code block from `main.ino`. 

https://github.com/1technophile/OpenMQTTGateway/blob/40a69037e5066b00a0ecc9bc88e9b6a5c4b176ca/main/main.ino#L539-L545

main/main.ino
jmw6773: Good call. I've moved them to `setup_wifi`

Legion2:
@jmw6773 you should rebase on the current development branch and resolve the merge conflict

main/User_config.h
Legion2: > I agree that mqtt_server_name is redundant as the PubSubClient library handles DNS and IP addresses the same and both can be configured using the same MQTT_SERVER setting.

This was already part of a other #696 PR, you should rebase your branch on the latest development branch, so there are no merge conflicts.

jmw6773: Didn't see that PR. I guess I need to rebase more often. I didn't realize I was that far behind on my fork.

Why are we keeping `mqtt_server_name` in PR #696? Couldn't `mqtt_server_name` and `mqtt_server` be combined?

jmw6773:
I can't get my branch to rebase to the current dev branch correctly. 
Still trying..

main/User_config.h
Legion2: `mqtt_server_name` and `mqtt_server` can be combined but the PR was only to fix TLS connections, so I didn't refactor it as much.

Legion2: > Didn't see that PR. I guess I need to rebase more often. I didn't realize I was that far behind on my fork.

@jmw6773 you may use `git rebase -i upstream/development` after you added the upstream repository and fetched it, to cleanup this pull request.

jmw6773: @Legion2 , thank you for the command. I'm still learning how to use `git`. I've rebased to upstream/development, then to this branch.

Legion2: your rebase didn't work, now the commits are duplicated. After a rebase you should not do a `git pull`! because it will merge it with itself. After a rebase you must force push the change to the branch or you could push them to a new branch. I will fix that and push it to a new branch.

Legion2: here is how it looks after the rebase, all the merges are gone and I removed the "revert" commits and commits already in the development branch https://github.com/1technophile/OpenMQTTGateway/compare/Move-MQTT-Config-rebased

jmw6773: The rebase you did has the `Var naming and efficiency change ` commit where I removed `mqtt_server_name` and combined it with `MQTT_SERVER`.

Should I try rebasing on my side again?

Legion2: > The rebase you did has the Var naming and efficiency change commit where I removed mqtt_server_name and combined it with MQTT_SERVER.

yes this was the plan

> Should I try rebasing on my side again?

do you know how git rebase -i works?

jmw6773: Not really. I know what it's suppose to be used for, but not really how to use it.

If it would be easier, we can can close this PR and create new one from the branch you created. 
Otherwise, I can refork, re-apply the changes, and then open a new PR.
I've begun to lose confidence in fixing this PR.

Legion2: If you want to learn some cool git feature look into `git rebase -i`, it must be used with care but you can fix any git problem with it. On the other hand doing rebases there is much which can go wrong, so you have to really know what you are doing.

We can use the `Move-MQTT-Config-rebased` branch I created, it only contains the necessary commits of this PR. I can force push them on your branch then, we don't have to create a new PR. If I force push on your branch you should delete your local branch (`git checkout development && git branch -D Move-MQTT-Config && git pull && git checkout Move-MQTT-Config`) to prevent any new merge conflicts.

jmw6773: I'll definitely look into it. I got far enough to know that I could cause problems. I'll create a test repo to practice with.

I don't know if you need it, but I've invited you as a Collaborator to my fork so you can push any changes you need. I'll backup my local clone and re-clone my fork locally once you've make any changes.

Sorry for making you do all of this extra work.

Legion2: thanks for the invitation, but it is not required, as there is a PR setting in GitHub that allows upstream collaborators to push to the branch of the pull request, even if the branch comes from a fork.

> I'll backup my local clone and re-clone my fork locally once you've make any changes.

It should be enough to do a `git fetch` and `git reset --hard` when you are on the Move-MQTT-Config branch.

> Sorry for making you do all of this extra work.

No problem

Legion2: *`git reset --hard origin/Move-MQTT-Config`

jmw6773: I've hard reset on all my local copies.

1technophile:
Thanks!

PR #706, opened 2020-07-23
Add Avatto/Bakeey IR boards definition

1technophile:
Tested with S06
-Emitting IR OK
-Receiving IR OK
-TRIGGER GPIO OK
-led info, brief but OK


DigiH:
Got two of these nice small IR remotes for $7.50 each after reading your blog post about the possibility to flash them with OpenMQTTGateway. Although they were not Avatto branded they had the same TYWE3S micro controller, so following your blog instructions went very smoothly. Great reach with their IR LED blaster array.

Thanks again for the tip and nice to see that it will be included in a future release.

1technophile:
Thanks for your feedback,
Indeed it is a nice compact way of controlling IR appliance, it's a kind of Sonoff RF bridge for infrared with more protocols supported.

epicurean70:
Is there a simple step by step guide on how to flash these Avatto IR boards for a newbie like myself?

1technophile:
Here is it : 
https://1technophile.blogspot.com/2020/07/avatto-s06-ir-gateway-compatible-with.html

PR #710, opened 2020-07-26
[Breaking Change] Change HCSR501 MQTT key to 'presence'

jmw6773:
@1technophile ,
  As discussed in issue #690, I've changed the key used for the HCSR501 PIR sensor to be 'presence'. This will fix MQTTDiscovery, but will break users who manually have the sensor setup in HA.

  Please let me know if I've missed any other doc locations.


1technophile:
Thanks!

Issue #715, opened 2020-08-02
Received RF values are stored twice


jmw6773:
**Describe the bug**
Values received by ZgatewayRF (and most likely other gateways) are being stored in the `receivedSignal` buffer twice. (Once for the receiving of the signal and once for the MQTT callback to the `subjectMultiGTWKey` topic.

**To Reproduce**
Run the Zgateway RF with `LOG_LEVEL` set to `LOG_LEVEL_TRACE`.

When a RF signal is received:
1. the code is stored (if not duplicated)
2. OMG transmits to 433toMQTT
3. OMG receives callback from 433toMQTT (to prevent duplicate signals from other OMG devices)
4. the code is stored again (no check for duplicate value)

**Expected behavior**
Code should only be stored once. Duplicate values should always be checked.

**Screenshots**

**Environment (please complete the following information):**
* Current development

**Additional context**
I believe this should be fixable by changing L1388, but don't know if it will have other adverse effects.
`    if (data != 0) {`
to
`    if (data != 0 && !isAduplicateSignal(data)) {`

https://github.com/1technophile/OpenMQTTGateway/blob/60d173ff5117aaaf5cf805f6a2146dca06fe0e93/main/main.ino#L1379-L1390

1technophile:
I think it's a good suggestion, feel free to submit a PR. 
And at first view no impact except the processing time.

jmw6773:
Looking closer, the test for `if (strstr(topicOri, subjectMultiGTWKey) != NULL)` (L1383) looks as though it can be removed as it is already performed in the `callback` function (L482). This should improve performance for all Gateways since it was outside the `#define`.

1technophile:
When we detect ` (strstr(topic, subjectGTWSendKey) != NULL)` in L482, (an MQTT command) we must not pass on the L1383 block so as to avoid storing the MQTT commands for deduplication.
Maybe we should change:
`  if (strstr(topicOri, subjectMultiGTWKey) != NULL) `
by
`  if ((strstr(topic, subjectGTWSendKey) == NULL) `
If you think it is more explicit.

jmw6773:
Oh, I understand what it's doing now. Yeah, it's fine as it is then. But it should probably be inside the `#define` as Gateways that aren't using RF/IR don't need to run the comparison, right?

```
[...]

PR #717, opened 2020-08-03
Prevent storing duplicate RF/IT signals

jmw6773:
As discussed in issue #715, this will check for duplicate RF/IR signals before storing them in the `receivedSignal` array. This will have a slight performance impact, but should be unnoticeable.

I've also:
- removed a redundant `if` statement from the `receivingMQTT` function
- in the `getMin` function, started the loop on index 1 as index 0 is always the first minimum (no need to check against itself)
- cast the SIGNAL_SIZE_UL_ULL values to a (unsigned long) for Log.trace functions in some functions (uint64_t values aren't supported by the ArduinoLog library)

### DIFF
--- a/main/ZgatewayRF.ino
+++ b/main/ZgatewayRF.ino
@@ -94,7 +94,8 @@ void RFtoMQTT() {
       RFtoMQTTdiscovery(MQTTvalue);
 #  endif
       pub(subjectRFtoMQTT, RFdata);
-      Log.trace(F("Store val: %lu" CR), MQTTvalue);
+      // Casting "receivedSignal[o].value" to (unsigned long) because ArduinoLog doesn't support uint64_t for ESP's 
+      Log.trace(F("Store val: %u" CR), (unsigned long)MQTTvalue);
       storeSignalValue(MQTTvalue);
       if (repeatRFwMQTT) {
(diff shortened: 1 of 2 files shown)


main/ZgatewayRF.ino
```diff
@@ -94,7 +94,8 @@ void RFtoMQTT() {
       RFtoMQTTdiscovery(MQTTvalue);
 #  endif
       pub(subjectRFtoMQTT, RFdata);
-      Log.trace(F("Store val: %lu" CR), MQTTvalue);
+      // Casting "receivedSignal[o].value" to (unsigned long) because ArduinoLog doesn't support uint64_t for ESP's
```
1technophile, line 97: I think we should propose a PR to the ArduinoLog repository, I can do it if you want.

jmw6773: To be honest, I'm not really sure how to deal with printing uint64_t values. I haven't used them yet in other projects.

I know I'm only printing the partial value with by using my cast, but I'm not sure how to fix it in the ArduinoLog library.

1technophile: You are right, not so easy as the `print `function doesn't support long long.

PR #718, opened 2020-08-03
Enable BT use for Arduino Mega

1technophile:
by adding a ArduinoSTL library,
currently the vector function is missing for Arduino AVR boards.

### DIFF
--- a/docs/prerequisites/board.md
+++ b/docs/prerequisites/board.md
@@ -29,12 +29,16 @@ The boards below doesn't need hardware modification (or a lite one).
 The boards below need hardware modifications and electronic/hardware competencies. You have more possibilities of gateways combination compared to off the shelves ones.
 |DIY boards|RF|IR|BLE|LORA|GSM|Button|Relay|
 |-|:-:|:-:|:-:|:-:|:-:|:-:|:-:|
-|Arduino UNO|X|X(limited compared to ESP)|X|-|X|X|
+|Arduino UNO|X|X(limited compared to ESP)|-|-|X|X|
 |Arduino MEGA|X|X(limited compared to ESP)|X|-|-|X|X|
 |ESP32|X|X|X|X|not tested|X|X|
 |ESP8266|X|X|X|not tested|X|X|X|
 
-*Note that Pilight is only supported on ESP and Arduino UNO handle only 32bits values in our context.*
+::: INFO
+Pilight is only supported on ESP, Arduino UNO handle only 32bits values in our context.
+Setup based on HM10 doesn't support some BLE [devices](devices.md#for-ble-devices).
+:::
+*Note that *
 
 ![boards](../img/OpenMQTTGateway_boards.png)
 
--- a/docs/prerequisites/devices.md
+++ b/docs/prerequisites/devices.md
@@ -21,15 +21,18 @@ OpenMQTTGateway is able to scan all the BLE devices that advertise their data so
 | XIAOMI Mi Flora |HHCCJCY01HHCC|temperature/moisture/luminance/fertility|
 | XIAOMI Mi Jia |LYWSDCGO|temperature/humidity/battery|
 | XIAOMI Mi Lamp |MUE4094RT|presence|
-| INKBIRD |IBS-TH1|temperature/humidity/battery|
+| INKBIRD *|IBS-TH1|temperature/humidity/battery|
 | ClearGrass |CGG1|temperature/humidity/battery|
 | ClearGrass alarm clock|CGD1|temperature/humidity|
 | ClearGrass with atmospheric pressure |CGP1W|temperature/humidity/air pressure|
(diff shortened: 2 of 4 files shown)


1technophile:
@jmw6773 I added you as a reviewer as you may have found another solution regarding the vector library issue.

1technophile:
Hmm, the build is OK on my env :-(

jmw6773:
@1technophile,
  I just started looking at the BLE vector object yesterday. I'm trying to familiarize myself with how to use vectors now.

  A quick list of concerns that I want to look into are:
-   Are discovered devices ever removed from the vector, or does it grow indefinitely?
    - i.e. In a high traffic area, will the vector eventually overflow the HEAP/RAM?
-   Should be a garbage collector to remove stale devices?
    - Black and White listed devices shouldn't be removed
    - Should remove the HADiscovery topic if created as to not clutter HA?
-   As vector size isn't initialized with a set value, does adding values fragment memory?
    - i.e. if underlying storage is an array, is array realloc'd for every addition

  If you know the anwser to any of the above, I'd love to learn.

jmw6773:
I only use ESP8266 on a custom PCB for my OMG's with BT and ESP32 devices.
I don't think I have any HM-10 modules that are on breakout boards to test with a standard Arduino chip.

1technophile:
> Are discovered devices ever removed from the vector, or does it grow indefinitely?
i.e. In a high traffic area, will the vector eventually overflow the HEAP/RAM?

We are storing lists (white or black) objects and discoverable ones. By discoverables I mean all the one that have an identified `model `, so sensors mainly. Beacons and other BLE devices are not stored so it may be an issue for someone having hundreds of sensors but not in the regular use cases I think.

> Should be a garbage collector to remove stale devices?
Black and White listed devices shouldn't be removed
Should remove the HADiscovery topic if created as to not clutter HA?

For information the discovery devices are published with a `retain `flag, in an ideal way we would check if the device has its corresponding topic and rely on the broker for the discovery storing. I don't know if it changes things for HA but at least it would reduce the size of the list in OMG side.

> As vector size isn't initialized with a set value, does adding values fragment memory?
i.e. if underlying storage is an array, is array realloc'd for every addition

I don't have the  expertise to answer to this one :-) 

> I only use ESP8266 on a custom PCB for my OMG's with BT and ESP32 devices.
I don't think I have any HM-10 modules that are on breakout boards to test with a standard Arduino chip.

ok no problem

jmw6773:
Didn't catch that it was only storing devices with models (and white/black lists). It shouldn't be a problem on ESP devices then.
For safely, it might be good to set a MAX size though. Might also solve the issue with the UNO boards.

Hypothetically, what would happen if we re-transmit the HADiscovery topic every time the sensor is 'discovered' instead of storing them? (Aside from the performance, bandwidth (lol), and power consumption hit?) The unique ID and other values would remain the same, so HA shouldn't create a new device, right?

In PlatformIO if I open the definition for `vector`, it looks like the `push_back` function might reallocate the storage for each item added, but it's way over my head.

1technophile:
> Hypothetically, what would happen if we re-transmit the HADiscovery topic every time the sensor is 'discovered' instead of storing them? (Aside from the performance, bandwidth (lol), and power consumption hit?) The unique ID and other values would remain the same, so HA shouldn't create a new device, right?

This is what I was doing before but it generated some non relevant message on the broker and the users were complaining about them.

> Didn't catch that it was only storing devices with models (and white/black lists). It shouldn't be a problem on ESP devices then.
For safely, it might be good to set a MAX size though. Might also solve the issue with the UNO boards.

Yep, it makes sense, I will take a look

1technophile:
> Yep, it makes sense, I will take a look

Well, not so [trivial](https://stackoverflow.com/questions/31750562/unintrusive-way-to-enforce-size-constraint-on-stdvector), I'm asking myself if it is relevant to spend time on this as I don't think there is so much users with an UNO + HM10.
The solution is still relevant for a MEGA.
So a the solution could be to avoid the use of BT with Arduino UNO, added to that I'm not sure the MQTT_MAX_PACKET_SIZE on the UNO can handle the big length of the BT discovery topic.

jmw6773:
I'd have to agree that using an UNO + BT would be very restrictive. Sounds like it would be easiest to make wrapper function for the `push_back` to check the vector's size and remove oldest values in order to keep the vector from growing over a MAX size. (Like how you're doing the `receiveSignal` array.

If you want to spend the time on this, I would take a poll with the community to see how many people would actually use UNO +BT first though.

1technophile:
> If you want to spend the time on this, I would take a poll with the community to see how many people would actually use UNO +BT first though.

Good idea, here is [it](https://community.openmqttgateway.com/t/do-we-need-to-maintain-the-arduino-uno-comptability-with-bluetooth-through-hm10/1062)

PR #721, opened 2020-08-05
BT MAC addr ':' fix

jmw6773:
@1technophile ,
  Sorry, my suggestion broke the loop that adds the ':' into the BT Mac addresses.


1technophile:
Haven't seen it too

1technophile:
Thanks for pointing it!

PR #724, opened 2020-08-08
BLE Optimizations and buffer overflow fix

jmw6773:
For the past two weeks I have been investigating random crashes and sometime my OMG losing it's Gateway_Name after a random length of time. (Sometimes 5 minutes, others 3 hours.)
My first thoughts was that I was running out of RAM on my ESP8266 due to running BLE, RF, IR, BME280, PIR, and Analog. Long story short for all my testing, the heap and stack seem to be fine.

However, in the `BTtoMQTT()` method for ESP8266's, the length of the 'Rest Data' is never checked to be less than the decompose.extract[60] buffer. 
https://github.com/1technophile/OpenMQTTGateway/blob/519eb73deee87ae74ffcc48d31cccbf8eb36c83c/main/ZgatewayBT.ino#L607-L611

I believe that this buffer was being overflowed in the main.ino:`extract_char` function when 
https://github.com/1technophile/OpenMQTTGateway/blob/519eb73deee87ae74ffcc48d31cccbf8eb36c83c/main/main.ino#L227

I have been running two OMG's (both ESP8266) for 24hrs now without any resets, but would appreciate if someone else could test that I haven't messed up the MAC or Service Data parsing of the packets. (I don't have any Mi Flora or other devices that can be read by the HM-10.)

### Changes

- Instead of just adding a verification for the `rest data length`, I've updated the `BTtoMQTT` function to parse BLE devices as they are received by the HM-10 module, instead of waiting for the entire reply from the HM-10 discovery. This reduces the size of the `returnedString` String (from >800 to <300 on my tests).
This is accomplished by checking the received `returnString` for both the `BLEdelimiter` as the start of a BLE device and a `CR LR` ("0d0a") as the [end of a packet.](https://1technophile.blogspot.com/2017/11/mi-flora-integration-to-openmqttgateway.html)
The `returnedString` is cleared after every `TimeBtw_Read`.

- As I'm also reserving space for the `returnedString` (512 length) at setup, it should reduce HEAP fragmentation as the `returnedString` isn't being reallocated as it grows. (Unless it goes over 512.)

- Even though I rewrote the `extract_char` function to use pointer arithmetic instead of creating temporary arrays, I've removed calls to the `extract_char` function from the BLE gateway. Instead I'm sectioning the BLE packet using strings. This should reduce the HEAP by not needing the decompose.extract[60] for each section.

- I shifted the `decompose` start values by -16 as I'm removing the `BLEdelimiter` from the inspected BLE packet.

- Changed String concatenations to use `+=` as [I've read that it's more efficient](https://cpp4arduino.com/2018/11/21/eight-tips-to-use-the-string-class-efficiently.html#tip-5-mutate-a-string-instead-of-creating-temporaries).

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -61,11 +61,9 @@ struct BLEdevice {
 #  define device_flags_isBlackL 1 << 2
 
 struct decompose {
-  char subject[4];
   int start;
   int len;
   bool reverse;
-  char extract[60];
 };
 
 vector<BLEdevice> devices;
@@ -402,9 +400,9 @@ class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks {
           for (int i = 0; i < serviceDataLength; i++) {
             int a = serviceData[i];
             if (a < 16) {
-              returnedString = returnedString + "0";
+              returnedString += F("0");
             }
-            returnedString = returnedString + String(a, HEX);
+            returnedString += String(a, HEX);
           }
           char service_data[returnedString.length() + 1];
           returnedString.toCharArray(service_data, returnedString.length() + 1);
@@ -552,11 +550,12 @@ bool BTtoMQTT() { // for on demand BLE scans
 
 SoftwareSerial softserial(BT_RX, BT_TX);
 
-String returnedString = "";
+String returnedString((char*)0);
 unsigned long timebt = 0;
 
 // this struct define which parts of the hexadecimal chain we extract and what to do with these parts
-struct decompose d[6] = {{"mac", 16, 12, true}, {"typ", 28, 2, false}, {"rsi", 30, 2, false}, {"rdl", 32, 2, false}, {"sty", 44, 4, true}, {"rda", 48, 60, false}};
+// {"mac"}, {"typ"}, {"rsi"}, {"rdl"}, {"sty"}, {"rda"}
+struct decompose d[6] = {{0, 12, true}, {12, 2, false}, {14, 2, false}, {16, 2, false}, {28, 4, true}, {32, 60, false}};
 
 void setupBT() {
   BLEinterval = TimeBtw_Read;
@@ -573,6 +572,9 @@ void setupBT() {
   softserial.print(F("AT+PIO11" CR)); // When not connected (as in BLE mode) the LED is off. When connected the LED is solid on.
 #    endif
   delay(100);
+#    if defined(ESP8266)
+  returnedString.reserve(512); //reserve memory space for BT Serial. (size should depend on available RAM)
+#    endif
   Log.trace(F("ZgatewayBT HM1X setup done " CR));
 }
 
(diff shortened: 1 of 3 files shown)


1technophile:
Thanks for these optimizations, I will push it to my bench and see how things are going.

1technophile:
I'm successfully retrieving data from LYWSDCGQ, JQJCY01YM & CGG1.
I will let it run to see how stable it is.

1technophile:
Well, running for a while now without issues:
`{"uptime":89646,"version":"version_tag","freemem":37184,"rssi":-46,"SSID":"Mediacom","ip":"192.168.1.9","mac":"00:11:22:33:44:55","wifiprt":0,"interval":55555,"modules":"BTHADiscovery"}`

Freemem is stable:
image

And it is still retrieving sensors data.

1technophile:
Well, running for a while now without issues:
`{"uptime":89646,"version":"version_tag","freemem":37184,"rssi":-46,"SSID":"wifissid","ip":"192.168.1.9","mac":"00:11:22:33:44:55","wifiprt":0,"interval":55555,"modules":"BTHADiscovery"}`

Freemem is stable:
image

And it is still retrieving sensors data.

main/ZgatewayBT.ino
```diff
@@ -581,73 +583,78 @@ bool BTtoMQTT() {
...
+        if (restDataLength <= 60)
+          restData = token.substring(d[5].start, (d[5].start + restDataLength));
+
+        Log.trace(F("Creating BLE buffer" CR));
+        StaticJsonBuffer<JSON_MSG_BUFFER> jsonBuffer;
+        JsonObject& BLEdata = jsonBuffer.createObject();
+
+        Log.trace(F("Id %s" CR), (char*)mac.c_str());
+        BLEdata.set("id", (char*)mac.c_str());
+        BLEdevice* device = getDeviceByMac((char*)mac.c_str());
```
1technophile, line 642: You will need to merge the dev branch before an integration, especialy the modifications done [here ](https://github.com/1technophile/OpenMQTTGateway/pull/720)and [here](https://github.com/1technophile/OpenMQTTGateway/commit/0cd2464bf11c61b37f6e70fed1d86bac3f5860c4)

jmw6773: Both of the commits mentioned are in this [branch commit history](https://github.com/jmw6773/OpenMQTTGateway/commits/HM10---BT-parsing). 

I did a quick glance over the code and the changes appear to be this this PR code. I can try re-basing to the `dev` branch again if you think I'm missing something. (Still new to rebase, so it's highly possible.)

Regarding the modifications mentioned, I'm pulling the MAC address directly from the BLE `token` string instead of using the `extract_char` function. This elements the `for` loop that was previously added, and uses the `start` and `len` values defined in the decompose struct. It should be injecting `:` into the correct places. (Couldn't get the relevant code block to inline.)
https://github.com/1technophile/OpenMQTTGateway/pull/724/files#diff-f711e1aab0b74a5422256770df31e7a1R617-R624

1technophile: You don't need to rebase, my mistake.

1technophile:
Gateway still runing on my NodeMCU V2 without issue

jmw6773:
After 4 days running, I experienced another buffer overflow on my test board running (BLE, RF, BME280) that overwrote part of the OMG Gateway Name.

I've restarted that OMG and setup another board with just the BLE sensor. If I can get the RF board to crash but not the BLE-only board, I might need to look into the RF sensor library to see if it might be causing the issue. I'm seeing some odd RF signals at times.

1technophile:
> After 5 days running, I experienced another buffer overflow on my test board running (BLE, RF, BME280) that overwrote part of the OMG Gateway Name.

Mine is still running since 60h now with BT only.

1technophile:
120h of continuous operation for me, do you think we could merge this PR?

jmw6773:
I haven't seen any issues with my two gateways here either. I'll rebase as there have several other merges since this PR and take it out of 'draft'.

1technophile:
Thanks!

PR #727, opened 2020-08-09
Allow prod_env.ini parameters to override User_config.h

mpember:
More parameters are now able to be defined in prod_env.ini, allowing the values in User_config.h to be ignored.


1technophile:
You need to apply a [cclang ](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md)format to the files:

mpember:
It seems a bit random as to which parameters are padded to align the values and which only have a single space between the parameter name and the value.

e.g. the Gateway_* defines are only single spacing between parameter and value, but WifiManager_* are padded.

Is there some underlying reason that I am overlooking?

1technophile:
> Is there some underlying reason that I am overlooking?

I don't think so, the value seems to be aligned by paragraph

PR #729, opened 2020-08-10
Allow prod_env.ini parameters to override User_config.h

mpember:
More parameters are now able to be defined in prod_env.ini, allowing the values in User_config.h to be ignored.

### DIFF
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -89,16 +89,16 @@ const byte mac[] = {0xDE, 0xED, 0xBA, 0xFE, 0x54, 0x95}; //W5100 ethernet shield
 #endif
 
 #ifndef WifiManager_password
-#  define WifiManager_password "your_password" //this is going to be the WPA2-PSK password for the initial setup access point
+#  define WifiManager_password            "your_password" //this is going to be the WPA2-PSK password for the initial setup access point
 #endif
 #ifndef WifiManager_ssid
-#  define WifiManager_ssid Gateway_Name //this is the network name of the initial setup access point
+#  define WifiManager_ssid                Gateway_Name //this is the network name of the initial setup access point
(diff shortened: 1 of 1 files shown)


1technophile:
Clang format it like that:

```
#ifndef WifiManager_password
#  define WifiManager_password "your_password" //this is going to be the WPA2-PSK password for the initial setup access point
#endif
#ifndef WifiManager_ssid
#  define WifiManager_ssid Gateway_Name //this is the network name of the initial setup access point
#endif
#ifndef WifiManager_ConfigPortalTimeOut
#  define WifiManager_ConfigPortalTimeOut 120
#endif
#ifndef WifiManager_TimeOut
#  define WifiManager_TimeOut 5
#endif
```

mpember:
The problem appears to be caused by the new lines making cclang no longer treat the sections as a single paragraph.  This means the automatic formatting doesn't get done by the extension in VS Code.

mpember:
> Clang format it like that:
> 
> ```
> #ifndef WifiManager_password
> #  define WifiManager_password "your_password" //this is going to be the WPA2-PSK password for the initial setup access point
> #endif
> #ifndef WifiManager_ssid
> #  define WifiManager_ssid Gateway_Name //this is the network name of the initial setup access point
> #endif
> #ifndef WifiManager_ConfigPortalTimeOut
> #  define WifiManager_ConfigPortalTimeOut 120
> #endif
> #ifndef WifiManager_TimeOut
> #  define WifiManager_TimeOut 5
> #endif
> ```

That is the exact formatting which is causing the conflict.  The original code includes padding, my updated code does not.

1technophile:
Maybe try to push the PR again and I will try to apply the format and debug it.

PR #735, opened 2020-08-17
Add the possibility to retrieve BLE sensors data through a connect function

1technophile:


1technophile:
rebasing instead of merge

Issue #725, opened 2020-08-08
Memory leakage in getting manufacture data and the fix.


l8l8l:
BLE Scan runs out of memory when processing manufacture data, 'freemem'  reduced very quickly and leads frequent crash.

---------ZgatewayBT.ino-------------

      if (advertisedDevice.haveManufacturerData()) {
        char* manufacturerdata = BLEUtils::buildHexData(NULL, (uint8_t*)advertisedDevice.getManufacturerData().data(), advertisedDevice.getManufacturerData().length());
        Log.trace(F("Manufacturer Data: %s" CR), manufacturerdata);
        BLEdata.set("manufacturerdata", manufacturerdata);
        free(manufacturerdata); **<-  Add the pointer free here.**
      }
-------------------------------------
Tested with default_envs = esp32dev-ble, pubUnknownBLEServiceData, pubBLEManufacturerData, pubBLEServiceUUID are enabled to scan BLE Beacons. freemem stays at 50KB.

1technophile:
Thanks for pointing that

jmw6773:
@l8l8l , great find. This seems to have resolved the random crashing I was seeing on my ESP32 device.

PR #736, opened 2020-08-17
BLE sensors data through a connect, LYWSD03MMC and MHO-C401 add

1technophile:
Add LYWSD03MMC low-cost BLE temperature sensor
Add the possibility to have 1 BLE connect operation every X BLE scans
Add MHO-C401


docs/use/ble.md
```diff
@@ -73,7 +73,7 @@ So as to keep your white/black list persistent you can publish it with the retai
 `mosquitto_pub -r -t home/OpenMQTTGateway/commands/MQTTtoBT/config -m '{"white-list":["01:23:14:55:16:15","4C:65:77:88:9C:79","4C:65:A6:66:3C:79"]}'`
 :::
 
-## Setting the time between scans and force a scan
+## Setting the time between BLE scans and force a scan
 
 If you want to change the time between readings you can change the interval by MQTT, if you want the BLE scan every 66seconds:
```
jmw6773, line 78: Splitting these sentences and adding a space helps the readability.

```
If you want to change the time between readings you can change the interval by MQTT.
For example, if you want the BLE to scan every 66 seconds:
```

main/main.ino
```diff
@@ -1293,6 +1293,7 @@ void stateMeasures() {
   SYSdata["lowpowermode"] = (int)low_power_mode;
 #    endif
   SYSdata["interval"] = BLEinterval;
+  SYSdata["scanbcnct"] = BLEscanBeforeConnect;
```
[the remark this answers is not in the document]
Legion2: We should also think about a refactoring how gateways are integrated in the main.h. Each gateway could be implemented as a class which inherit from a Gateway base class. Then only instances of these classes must be handled in the main.h and we have less `#ifdef ZgatewayXY` in the code.

main/ZgatewayBT.ino
```diff
@@ -566,8 +706,9 @@ unsigned long timebt = 0;
 struct decompose d[6] = {{0, 12, true}, {12, 2, false}, {14, 2, false}, {16, 2, false}, {28, 4, true}, {32, 60, false}};
 
 void setupBT() {
-  BLEinterval = TimeBtw_Read;
+  BLEinterval = TimeBtwRead;
```
Legion2, line 709: Remove this here, it is already set in `config_BT.h`

main/config_BT.h
```diff
@@ -29,27 +29,35 @@
...
-#define ServicedataMinLength 29
 
-#ifndef TimeBtw_Read
-#  define TimeBtw_Read 55555 //define default time between 2 scans
+#ifndef Scan_duration
+#  define Scan_duration 10000 //define the time for a scan --WARNING-- changing this value can lead to instability on ESP32
+#endif
+#ifndef ScanPerConnect
+#  define ScanBeforeConnect 10 //define number of scans before connecting to BLE devices (ESP32 only)
+#endif
```
jmw6773: Should this be:
```suggestion
#ifndef ScanBeforeConnect
#  define ScanBeforeConnect 10 //define number of scans before connecting to BLE devices (ESP32 only)
#endif
```

jmw6773:
This PR also looks to address issue #725

1technophile:
> This PR also looks to address issue #725

You are right

main/main.ino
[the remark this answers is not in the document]
1technophile: > For the future refactoring this should be done with a stateMeasures callback function in each gateway. Then we don't have to put the gateway related variables in public visibility.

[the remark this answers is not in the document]
1technophile: > We should also think about a refactoring how gateways are integrated in the main.h. Each gateway could be implemented as a class which inherit from a Gateway base class. Then only instances of these classes must be handled in the main.h and we have less #ifdef ZgatewayXY in the code.

I'm really interested in discussing the "ideal" architecture and responsibilities separation. I will begin a thread into the community forum.

1technophile:
> I'm sorry that I don't have either of these devices to test in my OMG environment.

We all contribute with the means and time we have :-) and I have no problem with that.

main/ZgatewayBT.ino
```diff
@@ -471,9 +605,12 @@ void coreTask(void* pvParameters) {
         if (low_power_mode == 2)
           digitalWrite(LOW_POWER_LED, 1 - LOW_POWER_LED_OFF);
         BLEscan();
-        launch_discovery();
+        // Launching a connect every BLEscanBeforeConnect
+        if (!((millis() / (BLEinterval + Scan_duration)) % BLEscanBeforeConnect))
+          BLEconnect();
```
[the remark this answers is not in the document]
1technophile: I'm going to replace this by a BLEscan counter it will be more reliable.

main/config_BT.h
1technophile: yes, it is corrected

main/ZgatewayBT.ino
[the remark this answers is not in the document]
1technophile: done

jmw6773:
Should the comment for this variable read something like: `//Number of BLE scans between connection cycles`

https://github.com/1technophile/OpenMQTTGateway/blob/23cbe1f5c2d0b283399505d59cff0ca681e44f9f/main/config_BT.h#L60

1technophile:
> Should the comment for this variable read something like: //Number of BLE scans between connection cycles

Yes, corrected

main/ZgatewayBT.ino
```diff
@@ -75,6 +88,10 @@ static bool oneWhite = false;
 
 int minRssi = abs(MinimumRSSI); //minimum rssi value
 
+unsigned int scanCount = 0;
```
Legion2, line 91: move this into `coreTask`

main/ZgatewayBT.ino
```diff
@@ -611,6 +754,8 @@ bool BTtoMQTT() {
     if (returnedString.equals(F(BLEEndOfDiscovery))) //OK+DISCE
     {
       returnedString.remove(0); //clear data string
+      scanCount++;
+      Log.notice(F("Scan number %d end " CR), scanCount);
```
Legion2, line 758: why do we increase scanCount here?

main/ZgatewayBT.ino
```diff
@@ -471,9 +608,12 @@ void coreTask(void* pvParameters) {
         if (low_power_mode == 2)
           digitalWrite(LOW_POWER_LED, 1 - LOW_POWER_LED_OFF);
         BLEscan();
-        launch_discovery();
+        // Launching a connect every BLEscanBeforeConnect
+        if (!(scanCount % BLEscanBeforeConnect) || scanCount == 1)
```
Legion2, line 612:
```suggestion
        if (scanCount % BLEscanBeforeConnect == 0)
```

main/ZgatewayBT.ino
```diff
@@ -439,11 +482,105 @@ void BLEscan() {
   BLEScan* pBLEScan = BLEDevice::getScan(); //create new scan
   pBLEScan->setAdvertisedDeviceCallbacks(new MyAdvertisedDeviceCallbacks());
   pBLEScan->setActiveScan(true); //active scan uses more power, but get results faster
-  BLEScanResults foundDevices = pBLEScan->start(Scan_duration, false);
-  Log.notice(F("Found %d devices, scan end deinit controller" CR), foundDevices.getCount());
+  BLEScanResults foundDevices = pBLEScan->start(Scan_duration / 1000, false);
+  scanCount++;
+  Log.notice(F("Found %d devices, scan number %d end deinit controller" CR), foundDevices.getCount(), scanCount);
```
Legion2, line 487: The counter should be added where it is used.

main/ZgatewayBT.ino
1technophile: It is hidden in the commit, this is for the future publication of the scan count to SYStoMQTT

main/ZgatewayBT.ino
1technophile: We need to do the BLEconnect after the first scan if we change it like that we will only do it at `scanCount = BLEscanBeforeConnect`
This is required by the low power mode, this mode puts the ESP to sleep after the BLEscan and the BLEconnect operations.

1technophile:
Thanks for your comments @Legion2 & @jmw6773 !

PR #737, opened 2020-08-18
Add new actuator : PWM controller

OliWright:
Thank you for this awesome project.  It fit my needs perfectly.

This PR adds a new actuator to control LEDs with the PWM outputs of the ESP32.  I don't have other microcontrollers to test it on, but it should be simple to adapt as long as there is sufficient PWM control.

I am using it for controlling lights in an aquarium.

It uses the 16-bit duty cycle mode of the ESP32 to eliminate the problem of not having fine control over LEDs when they are very dim.

It features smooth fading between states, and the ability to calibrate the gamma curve and the min and max levels for each channel.

Hopefully I have commented the code sufficiently so that I do not need a very detailed description here.

I didn't quite understand the comments in the development guidelines relating to `Test_config.h` - sorry.


1technophile:
Hi @OliWright ,

Thanks for this PR.
There is a FASTLED module available with the current code, and I'm asking myself if it can cover this use case, or is it two different functionalities?

OliWright:
Hi @1technophile ,

> There is a FASTLED module available with the current code, and I'm asking myself if it can cover this use case, or is it two different functionalities?

IIUC, the FASTLED module is for controlling addressable LED strips (like [NeoPixel](https://www.adafruit.com/category/168) strips).

The PWMLED module is for controlling LEDs that are directly connected to the micro-controller, using PWM outputs to modulate the brightness.

The downside of this approach over the FASTLED module is the number of LEDs that can be controlled.  Addressable LED strips can have hundreds of individually addressable LEDs.  But the PWMLED module currently only controls 5 outputs, and is restricted by the number of PWM outputs on the microcontroller.

The benefit of the PWMLED approach though is that those LEDs can typically be much higher power.  My current application is for controlling some aquarium lights.  There are different combinations of 12v LEDs, including RGB strips and high power spot lights.  Each channel is connected to the microcontroller through a MOSFET.  I have a beefy 12v power supply, and a small DC-DC buck converter to tap off the 3.3v for the ESP32.

So I'd say they're sufficiently different.  For non-disco applications, PWMLED is a winner :-)

1technophile:
Understood, thanks for the explanation!

1technophile:
Are there other use cases for a PWM module outside the LED scope?
I'm thinking of myself if this module should not be more generic so as to address others needs?

OliWright:
Good question.

Some other things that use PWM control that I'm aware of are fans, and servos - so a generalised PWM controller might well be a useful idea.

Some parts of the source are quite specific for LEDs though - like gamma conversion.   And the naming of the channels is R,G,B,W0,W1.  But it shouldn't hurt much to rename those to be more general numbered channels.  And if I change the default gamma to 1.0 (linear), then that would be more general purpose too.

Yes - I think that makes a lot of sense actually.

I could add configuration options in the header file, for channel naming.

Leave it with me....

1technophile:
Great, if we can extend the use cases taken into account it would be nice!

OliWright:
It's more general purpose now.

- Renamed the actuator from `PWMLED` to just `PWM`.
- There are still references to LEDs in the comments, but more by way of example.
- Number of channels can be configured in the config header.
- Channel names can be configured in the config header (instead of always being "r", "g", "b", "w0", "w1")
- The value range to control the output is now 0-1 instead of 0-255, which makes more sense for a general purpose controller.
- Fading between values over time can now be controlled individually for each channel.  E.g. you can send an MQTT message to change 3 channels over the next 4 seconds, then you can send another message to change a different channel over 10 seconds.

I've not tested it yet beyond compiling.  Just pushed so you can take a look.

I'm sure somebody can come up with an entertaining project involving servos, fans and LEDs.

main/ZactuatorPWM.ino
```diff
@@ -0,0 +1,282 @@
...
+  }
+}
+#endif
+
+#ifdef simpleReceiving
+void MQTTtoPWM(char *topicOri, char *datacallback)
+{
+  // We currently only support JSON
+}
+#endif
```
1technophile, line 280: You can remove this, simplePublishing will be removed

main/main.ino
```diff
@@ -1528,6 +1543,9 @@ void receivingMQTT(char *topicOri, char *datacallback)
       #ifdef ZactuatorFASTLED
       MQTTtoFASTLED(topicOri, datacallback);
       #endif
+      #ifdef ZactuatorPWM
+      MQTTtoPWM(topicOri, datacallback);
+      #endif
```
1technophile, line 1548: you can remove this one also

main/ZactuatorPWM.ino
```diff
@@ -0,0 +1,282 @@
...
+    {
+      // Remap according to the calibration
+      linear = (linear * (calibrationMaxLinear[i] - calibrationMinLinear[i])) + calibrationMinLinear[i];
+    }
+
+    long dutyCycle = (long) (linear * kUNormToDutyCycle);
+#if ESP32
+    ledcWrite(i, dutyCycle);
+#else
+#   error TODO: Write code here for different microcontrollers
```
1technophile, line 188: would it be possible to adapt it for esp8266 easily?

main/User_config.h
```diff
@@ -175,10 +175,11 @@ int low_power_mode = DEFAULT_LOW_POWER_MODE;
...
 //#define ZsensorGPIOInput "GPIOInput" //ESP8266, Arduino, ESP32
 //#define ZmqttDiscovery "HADiscovery"//ESP8266, Arduino, ESP32, Sonoff RF Bridge
-//#define ZactuatorFASTLED "FASTLED"  //ESP8266, Arduino, ESP32, Sonoff RF Bridge
+//#define ZactuatorFASTLED "FASTLED" //ESP8266, Arduino, ESP32, Sonoff RF Bridge
 //#define ZboardM5STICKC "M5StickC"
-//#define ZboardM5STACK "ZboardM5STACK"
-//#define ZradioCC1101  "CC1101" //ESP8266, ESP32
+//#define ZboardM5STACK  "ZboardM5STACK"
+//#define ZradioCC1101   "CC1101"   //ESP8266, ESP32
+//#define ZactuatorPWM   "PWM"      //ESP32
```
1technophile, line 182: Could you add also please the module into [esp32dev-all environment](https://github.com/1technophile/OpenMQTTGateway/blob/605da549b94be482e0d0e3abfff97ff2e3bb5cb7/platformio.ini#L169) so as to enable the CI to build the code.

OliWright: Done - but I had to make the following changes locally to make it build....

```
-rfm69 = RFM69@4a6f77ad47
+rfm69 = RFM69

-bmp180 = BMP180@efac46bd8d
+bmp180 = BMP180

-htu21 = SparkFun HTU21D Humidity and Temperature Sensor Br@1.1.3
+htu21 = SparkFun HTU21D Humidity and Temperature Sensor Breakout@1.1.3
```

main/ZactuatorPWM.ino
OliWright: Yes, it should be possible.

I don't have an esp8266 to test on, but I can write some code and make sure it at least compiles.  Pretty sure it'll work first time, because stuff usually does, right? 😉 

Also - I'll order one.

OliWright:
Updated the docs.

main/ZactuatorPWM.ino
OliWright: The module now compiles for esp8266 and Arduino.

1technophile:
Thanks for the work and the comments integration!

There is currently a conflict between the development branch and your branch related to main.ino, could you resolve it please, I can also give you help on that if needed.

OliWright:
It turns out that my fork was very out of date.  I've updated it now, and rebased my branch - so the conflicts are resolved now.

I've also taken most of the lint changes, except for a few that make the code less readable IMO.  Let me know if you need me to take those too.

1technophile:
Thanks for the rebase, 
I would prefer that you follow the current Lint for this PR, you may propose some changes to the existing one on another PR if you would like.

1technophile:
@OliWright Hi!

I would like to integrate this PR, may I change the lint formatting and integrate, or do you think you will find some time to do it?

The stock of PR is growing 🙂 and I would like to avoid too much rebase and merge 😵 for everybody.

OliWright:
Hi Florian,

I'll try to get it done today.  Just got a little swamped at work :-)

  Oli

1technophile:
Thanks!

OliWright:
Woot!

PR #746, opened 2020-08-27
 Autoreset option in case of not connection with the broker

Odyno:
This is an proposal of improvement: auto reset of configuration if no broker is available.

The details:
- Autoreset in case of not connection with the mqtt broker
- Improved the information on terminal in order to provide information about the TRIGGER_GPIO or if it's configured or not the autoreset
- Provide the options to enable or disable this feature, as default is disabled.
- Updated the documentation

Tested on my device: ESP32 Development Board with CP2102

### DIFF
--- a/docs/upload/portal.md
+++ b/docs/upload/portal.md
@@ -5,28 +5,33 @@ From your smartphone search for your OpenMQTTGateway wifi network and connect to
 
 ![Wifi manager menu](../img/OpenMQTTGateway_Wifi_Manager_menu.png)
 
-* Click on Configure WiFi
+- Click on Configure WiFi
 
 ![Wifi manager parameters](../img/OpenMQTTGateway_Wifi_Manager_enter_parameters.png)
 
-* Select your wifi
-* Set your wifi password
-* Set your MQTT Server IP
-* Set your MQTT Server username (facultative)
-* Set your MQTT Server password (facultative)
-* Set your MQTT base topic if you need to change it (you must keep the / at the end)
-* Set your gateway name if you need to change it
+- Select your wifi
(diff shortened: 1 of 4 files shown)


mpember:
If using an internet-based broker, could something as simple as an ISP outage result in a reset of all your devices?  That seems like a rather extreme response to what could be a short-term outage.

Is it not possible to resolve a misconfiguration with an OTA flash?

jmw6773:
I have to to agree with @mpember. I just went through bad storm and needed to shutdown my local MQTT server due to power concerns. However, doing any hardware maintenance, or extended server patching would also cause the MQTT service to be down long enough for all of my gateways to reset.

Having the OMG automatically erase all of it's settings seems a bit extreme for not being able to connect to MQTT. (Restarting the OMG when it can't connect to MQTT would be nice though.)

Assuming `ESPWifiManualSetup` isn't used, would it possible to restart the OMG and go into the WiFiManager config mode when you are unable to connect to MQTT? (`setup_wifimanager(false)`)
This should auto-populate the WiFiManager config page with the current values from the config.json, but also allow you to change you MQTT settings.

mpember:
> Assuming `ESPWifiManualSetup` isn't used, would it possible to restart the OMG and go into the WiFiManager config mode when you are unable to connect to MQTT? (`setup_wifimanager(false)`)
> This should auto-populate the WiFiManager config page with the current values from the config.json, but also allow you to change you MQTT settings.

I would suggest making this the action when the WLAN fails to connect (after a set number of retries).  This way, simply turning off your wireless radio (in your access point / router) and rebooting the devices would achieve the desired outcome.

Issue #408, opened 2019-05-28
Lack of valueAsASubject in ZgatewayPilight messages


wired-dev:
I resolved the issue for myself by adding 
`
    RFPiLightdata.set("message", (char *)message.c_str());
      #ifdef valueAsASubject
        StaticJsonBuffer<JSON_MSG_BUFFER> jsonBuffer2;
        JsonObject& RFPiLightdatain=jsonBuffer2.parseObject((char *)message.c_str());
        RFPiLightdata.set("value", RFPiLightdatain["id"]);
      #endif
`
to ZgatewayPilight;
this also made it easier to read the Digoo-DG-R8S device.

wired-dev:
`    JsonObject& RFPiLightdatain=jsonBuffer2.parseObject((char *)message.c_str());
    if(isDigit(RFPiLightdatain["id"])){
    RFPiLightdata.set("value", RFPiLightdatain["id"]);
    } else {
    RFPiLightdata.set("value", RFPiLightdatain["id"].as<String>());  
    }`

wired-dev:
then in the main c file I changed it to do strings too.

`    #ifdef valueAsASubject
      if(isDigit(data["value"])){
        unsigned long value = data["value"];
//      if (value != 0){
        topic = topic + "/"+ String(value);
      } else if(data["value"].as<String>()!= 0) {
        topic = topic + "/"+ data["value"].as<String>();
      }
    #endif`
this makes it work for more devices on pilight.

akasma74:
Where did you add it and how does it help?
I own some DG-R8S sensors and use them with HA - could you share your relevant HA config?
I tried to use valueAsASubject, but didn't find it useful and currently use demultiplexers.

p.s maybe it's more reasonable to create a PR? ;)

wired-dev:
not sure how to do a PR; but yeah just starting with git

1technophile:
@wired-dev if you need help for the PR don't hesitate to ask questions, in some simple step:
- download github desktop
- fork OMG development branch
- clone it localy
- modify your files localy
- commit your modifications and publish them from github desktop
- go to your fork on github
- you should see a button to submit a PR

PR #752, opened 2020-08-31
Fix pilight value as subject

hugokernel:
Thank you for this project ;)

The Pilight gateway with the valueAsASubject parameter does not work (compilation problem).

This PR fixes this problem and closes the #408 ticket.

I also added a patch that allows to find a correct ID (in order to have a non null topic) because PiLight only reads ID and ignores the others fields.

One question, in the ZgatewayPilight.ino file, there are this line: `RFPiLightdata.set("length", (char*)deviceID.c_str());` is it normal to set the length value from the deviceID value?

Thank you.

### DIFF
--- a/main/ZgatewayPilight.ino
+++ b/main/ZgatewayPilight.ino
@@ -47,6 +47,7 @@ void pilightCallback(const String& protocol, const String& message, int status,
     RFPiLightdata.set("message", msg);
     RFPiLightdata.set("protocol", (char*)protocol.c_str());
     RFPiLightdata.set("length", (char*)deviceID.c_str());
+    RFPiLightdata.set("value", (char*)deviceID.c_str());
     RFPiLightdata.set("repeats", (int)repeats);
     RFPiLightdata.set("status", (int)status);
     pub(subjectPilighttoMQTT, RFPiLightdata);
--- a/main/ZgatewayPilight.ino
+++ b/main/ZgatewayPilight.ino
@@ -52,7 +52,7 @@ void pilightCallback(const String& protocol, const String& message, int status,
     if (!strlen(device_id)) {
       // deviceID returned from Pilight is only extracted from id field
       // but some device may use another name as unique identifier
-      char * choices[] = {"key", "unit", "systemcode", "unitcode", "programcode"};
+      char* choices[] = {"key", "unit", "systemcode", "unitcode", "programcode"};
 
       for (uint8_t i = 0; i < 5; i++) {
         if (msg[choices[i]]) {
--- a/main/ZgatewayPilight.ino
+++ b/main/ZgatewayPilight.ino
@@ -47,7 +47,22 @@ void pilightCallback(const String& protocol, const String& message, int status,
     RFPiLightdata.set("message", msg);
     RFPiLightdata.set("protocol", (char*)protocol.c_str());
     RFPiLightdata.set("length", (char*)deviceID.c_str());
-    RFPiLightdata.set("value", (char*)deviceID.c_str());
+
+    const char* device_id = deviceID.c_str();
+    if (!strlen(device_id)) {
+      // deviceID returned from Pilight is only extracted from id field
+      // but some device may use another name as unique identifier
+      char * choices[] = {"key", "unit", "systemcode", "unitcode", "programcode"};
+
+      for (uint8_t i = 0; i < 5; i++) {
+        if (msg[choices[i]]) {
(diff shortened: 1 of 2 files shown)


1technophile:
Thanks for this PR!

> One question, in the ZgatewayPilight.ino file, there are this line: RFPiLightdata.set("length", (char*)deviceID.c_str()); is it normal to set the length value from the deviceID value?

Seems strange, I think it should be removed.

hugokernel:
> > One question, in the ZgatewayPilight.ino file, there are this line: RFPiLightdata.set("length", (char*)deviceID.c_str()); is it normal to set the length value from the deviceID value?
> 
> Seems strange, I think it should be removed.

Or replaced with the length of the payload ?

What is the purpose of the length field in general?

1technophile:
>Or replaced with the length of the payload ?
What is the purpose of the length field in general?

I think it is as an analogy with the length obtained by RCSwitch, which is the [number ](https://github.com/1technophile/OpenMQTTGateway/blob/7b9944d1ca43e63c9752a0f7706003a5c22d4779/main/ZgatewayRF.ino#L88) of signal bits.

We would have the `length `with Pilight if receiving a pulse train (the raw signal) through `setPulseTrainCallBack `, the current data used with Pilight throught  `setCallback` is at a higher level compared to RCSwitch , so I don't think we need it.

hugokernel:
> > Or replaced with the length of the payload ?
> > What is the purpose of the length field in general?
> 
> I think it is as an analogy with the `length `obtained by RCSwitch, which is the [number ](https://github.com/1technophile/OpenMQTTGateway/blob/7b9944d1ca43e63c9752a0f7706003a5c22d4779/main/ZgatewayRF.ino#L88) of signal bits.
> 
> We would have the `length `with Pilight if receiving a pulse train (the raw signal) through `setPulseTrainCallBack `, the current data used with Pilight throught `setCallback` is at a higher level compared to RCSwitch , so I don't think we need it.

:+1:

hugokernel:
When a radio frame arrives, Pilight will test all the protocols he knows in order to parse the content in search of the right one.

It is therefore quite possible to obtain several protocols that match for the same frame.

The problem is that the same frame can return several different protocols but with the same ID, the frames being published with the corresponding ID in the topic, all the protocols are not visible.

It would be interesting to have an option to add the protocol in the topic, for example, instead of having:

`home/OpenMQTTGateway/PilighttoMQTT/32342`.

We would have:

`home/OpenMQTTGateway/PilighttoMQTT/iwds07/32342`

The patch is extremely simple and very useful with PiLight (I use it), maybe it would be a good idea to add a general option for that or add it at least when Pilight is configured?

If you want, I can submit my commit in this PR.

1technophile:
> The problem is that the same frame can return several different protocols but with the same ID, the frames being published with the corresponding ID in the topic, all the protocols are not visible.

Do you mean that they are not published but overwritten by the last one?

hugokernel:
> > The problem is that the same frame can return several different protocols but with the same ID, the frames being published with the corresponding ID in the topic, all the protocols are not visible.
> 
> Do you mean that they are not published but overwritten by the last one?

Frame is published but overwritten by the last one.

hugokernel:
I saw you have a RFBridge hacked with Pilight on it, if you want to try:

```diff
diff --git a/main/main.ino b/main/main.ino
index 3c222e7..1cf9403 100644
--- a/main/main.ino
+++ b/main/main.ino
@@ -245,8 +245,9 @@ void pub(char* topicori, JsonObject& data) {
 #ifdef valueAsASubject
 #  ifdef ZgatewayPilight
     String value = data["value"];
+    String protocol = data["protocol"];
     if (value != 0) {
-      topic = topic + "/" + value;
+      topic = topic + "/" + protocol + "/" + value;
     }
 #  else
     SIGNAL_SIZE_UL_ULL value = data["value"];
```

The patch is only applied when Pilight is enabled but I think it could be interesting to have this option which can be useful especially to avoid collisions of devices with the same ID but different protocols or simply to classify things more finely.

If you think it will be interesting, I can do a pull request about that.

1technophile:
> If you think it will be interesting, I can do a pull request about that.

I agree, it can be interesting and enables us to have a more detailed view of the signal received. Nevertheless, I'm not sure about the best way to do it more globally. I have also some thinking around 433mhz that I would like to mature before making you work on this. 

Thanks for the proposal and the PR :-)

PR #754, opened 2020-09-01
Add LYWSD03MMC from ATC

1technophile:
https://github.com/atc1441/ATC_MiThermometer

https://www.youtube.com/watch?v=NXKzFG61lNs

With this custom firmware, the LYWSD03MMC advertises its data without encryption, so no need to connect to the sensor when using this custom firmware. The data are read every `BLEinterval` like all the other devices that advertise data.

Note that the flashing process takes less than 5minutes and is done over the air (impressive :-)).

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -819,15 +819,22 @@ void RemoveJsonPropertyIf(JsonObject& obj, char* key, bool condition) {
   }
 }
 
-double value_from_service_data(const char* service_data, int offset, int data_length) {
-  char rev_data[data_length + 1];
+/** 
+ * Retrieve a long value from a char array extract representing hexadecimal data, reversed or not
+ */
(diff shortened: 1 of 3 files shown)


main/ZgatewayBT.ino
```diff
@@ -819,15 +819,19 @@ void RemoveJsonPropertyIf(JsonObject& obj, char* key, bool condition) {
   }
 }
 
-double value_from_service_data(const char* service_data, int offset, int data_length) {
-  char rev_data[data_length + 1];
+double value_from_service_data(const char* service_data, int offset, int data_length, bool reverse) {
```
nemoo2580, line 822: Why does this function return double type instead of long? strtol returns long. It's problematic for bitwise operation on result (cast to long needed).

1technophile: Thanks for pointing that, I will take a look

PR #762, opened 2020-09-09
Fix ESPiLight transmission error reporting

melyux:
The `MQTTtoPilight` function's logic prevented the correct parsing of the error codes returned by ESPiLight. This fixes that and somewhat restructures the if/else logic therein for more efficiency.


1technophile:
Thanks for the fix!

PR #765, opened 2020-09-14
Add ability to invert the PIR Notify LED

jmw6773:
I've added the option to invert the PIR notification LED at build time. It's default value is set to drive an LED through a Mosfet. If the LED is driven directly from the AVR/ESP's pin, `INVERT_LED_NOTIFY` should be set to `true`.

### DIFF
--- a/docs/use/sensors.md
+++ b/docs/use/sensors.md
@@ -53,3 +53,6 @@ You can have another PIN mirror the value of the PIR sensor output by adding the
 This can be useful if you would like to connect an LED to turn on when motion is detected.
 
 `#define HCSR501_LED_NOTIFY_GPIO 4`
+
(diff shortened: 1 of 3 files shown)


1technophile:
Thanks for the PR!

On other code parts, I'm using a different method (see LED_RECEIVE_ON) by specifying if the ON status is at HIGH or LOW,
I think we should uniformize the method used, I have no bias for one or another method as long as they are easily understandable and the same.

jmw6773:
@1technophile, I don't have a preference either. I made a change that I think matches what you are doing with LED_RECEIVE_ON, but I haven't tested it, so I converted this PR to a draft. I'll try to build and push to my OMG tonight to confirm it functionality.

jmw6773:
My OMG has been stable with this change.

1technophile:
Thanks

PR #768, opened 2020-09-27
update temperature unit to work in HA

orrpan:
from "C" and "F" to "°C and "°F"

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -187,9 +187,7 @@ void MiFloraDiscovery(char* mac) {
   Log.trace(F("MiFloraDiscovery" CR));
   char* MiFlorasensor[MiFloraparametersCount][8] = {
       {"sensor", "MiFlora-lux", mac, "illuminance", jsonLux, "", "", "lx"},
-      {"sensor", "MiFlora-tem", mac, "temperature", jsonTempc, "", "", "°C"}, // remove for 0.9.6 release
-      {"sensor", "MiFlora-tempc", mac, "temperature", jsonTempc, "", "", "°C"},
-      {"sensor", "MiFlora-tempf", mac, "temperature", jsonTempf, "", "", "°F"},
+      {"sensor", "MiFlora-temp", mac, "temperature", jsonTempc, "", "", "°C"},
       {"sensor", "MiFlora-fer", mac, "", jsonFer, "", "", "µS/cm"},
       {"sensor", "MiFlora-moi", mac, "", jsonMoi, "", "", "%"}
       //component type,name,availability topic,device class,value template,payload on, payload off, unit of measurement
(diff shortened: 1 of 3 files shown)


orrpan:
I'm not 100% this works with F, have to test a little. (I don't use F but I see that my HA is thinking it also is and C, maybe it only needs one of them and then sets it by language?)

1technophile:
Thanks foe the PR.

@jw6773 as you already worked on this. Could you give your point of view about this PR ?

jmw6773:
@1technophile,
  I'm sorry. My wife just had our third child, so I don't know when I'll be able to test this fully. 

  I made the recommendation to remove the `°` in PR #703 as Home Assistant was ignoring the unit_of_measurement when the symbol was included and converting the temperature readings to a different unit. Upon removal of the symbol, the `tempc` and `tempf` sensor values displayed the transmitted unit without any conversions.

  From my understanding, which could be wrong, Home Assistant converts temperature values to what the user has set in the `configuration.yaml` file for the [`temperature_unit`](https://www.home-assistant.io/docs/configuration/basic/#temperature_unit) (C or F), or if that isn't defined, what it set for [`unit_system`](https://www.home-assistant.io/docs/configuration/basic/#unit_system) (metric or imperial). 

 In my Home Assistant setup I have the following set in the basic setup.
```
#  unit_system: metric
#  temperature_unit: C
```  

  This is an image from my OMG running current dev branch (just compiled) with an AHT10 sensor.
HA_Discovery

  Below is an image of the same device recompiled only adding the `°` in the AHTx0 mqttDiscovery definitions. (And a custom OMG version number.)
HA_Discovery2

  If we want, we could remove the `tempc` and `tempf` sensor definitions from mqttDiscovery and just transmit one temperature value with the appropriate unit_of_measurement, with the `°` symbol. The user would then rely on Home Assistant to do the conversions to what they have set in their config files. Both C and F values would still be posted to MQTT topics for other applications, but only one included in mqttDiscovery for Home Assistant.

orrpan:
Home assistant seems to kind of understand without the ° symbol. I notice the problem when I used the homekit integration, then home assistant couldn't pass any value to the sensor.

> If we want, we could remove the tempc and tempf sensor definitions from mqttDiscovery and just transmit one temperature value with the appropriate unit_of_measurement, with the ° symbol. The user would then rely on Home Assistant to do the conversions to what they have set in their config files. Both C and F values would still be posted to MQTT topics for other applications, but only one included in mqttDiscovery for Home Assistant.

This sounds good, should I then close this PR, create a new PR or juste remove all tempC and tempF and keep temp with "°C"?  
Btw should it be "temp" or "tem", seems to differ in a few places.

1technophile:
> My wife just had our third child

Congratulations for you and your family !!!

> This sounds good, should I then close this PR, create a new PR or just remove all tempC and tempF and keep temp with "°C"?

Just remove `tem` and `tempf` from the discovery (keep °C) on this PR or open a new one as you prefer.

`tem` will be removed in all the code later on.

jmw6773:
I tested your branch last night and it is crashing at startup. It's a remote device, so I couldn't get any logs.

However I believe it's because the count of parameters for each of the sensors needs to be updated to match. For example, with the MiFlora sensor, the `MiFloraparametersCount` should be `4` instead of the original `6`.

` #    define MiFloraparametersCount 4`

https://github.com/1technophile/OpenMQTTGateway/blob/e23eb1bb677701adebb93b7345cf4daa2bc3ca63/main/ZgatewayBT.ino#L186-L188

orrpan:
> I tested your branch last night and it is crashing at startup. It's a remote device, so I couldn't get any logs.
> 
> However I believe it's because the count of parameters for each of the sensors needs to be updated to match. For example, with the MiFlora sensor, the `MiFloraparametersCount` should be `4` instead of the original `6`.
> 
> ` # define MiFloraparametersCount 4`
> 
> https://github.com/1technophile/OpenMQTTGateway/blob/e23eb1bb677701adebb93b7345cf4daa2bc3ca63/main/ZgatewayBT.ino#L186-L188

Strange, mine is working. But anyway it seems like a smart thing to fix!

1technophile:
Thanks!

PR #770, opened 2020-09-30
The reset button function does not work if TRIGGER_GPIO = 0

NorthernMan54:
After a bit of a challenge getting the config correct, determined that the reset button option doesn't work on a ESP32 due to the definition logic


1technophile:
Thanks for pointing that!

PR #772, opened 2020-10-05
Dont make HADiscovery switch if RF is receive only

jmw6773:
@1technophile, thought of this while minor tweak while testing the HADiscovery from the previous PR.

If the RFGateway is only used to receive and not to transmit, via `RF_DISABLE_TRANSMIT`, is there a reason to create a switch in Home Assistant for the received code?

This PR disables creating a HADiscovery switch when the `RF_DISABLE_TRANSMIT` define is set.
Let me know if you want me to make a note of this in the docs and I'll push an update.

### DIFF
--- a/main/ZgatewayRF.ino
+++ b/main/ZgatewayRF.ino
@@ -37,7 +37,7 @@
 
 RCSwitch mySwitch = RCSwitch();
 
-#  ifdef ZmqttDiscovery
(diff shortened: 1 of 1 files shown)


1technophile:
Hi,

Yes well seen, 

> Let me know if you want me to make a note of this in the docs and I'll push an update.

I don't think it is necessary

PR #777, opened 2020-10-13
Multiple node

hugokernel:
I wish to use several OpenMQTTGateway node in order to cover all the surface of my house (the walls limit greatly).

However, I would like the node to use the same topic to report the information in order to have an unique collection point (I use Home Assistant).

Thank you for telling me what you think of this.

### DIFF
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -377,6 +377,8 @@ uint8_t wifiProtocol = 0; // default mode, automatic selection
 #define subjectSYStoMQTT             "/SYStoMQTT"
 #define subjectMQTTtoSYSset          "/commands/MQTTtoSYS/config"
 
+#define useMultipleNode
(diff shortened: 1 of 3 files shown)


1technophile:
> However, I would like the node to use the same topic to report the information in order to have a unique collection point (I use Home Assistant).

You could use the `+` function of MQTT so as to do that.
If you replace the gateway name by the `+` it is equivalent to a wild card.

hugokernel:
> > However, I would like the node to use the same topic to report the information in order to have a unique collection point (I use Home Assistant).
> 
> You could use the `+` function of MQTT so as to do that into your topic definition.
> If you replace the gateway name by the `+` it is equivalent to a wild card.

Indeed, it is a much simpler solution :+1:

PR #791, opened 2020-11-02
Add rf2 autodiscovery

hbraam:
Added Home Assistant discovery to RF2.

For each unique combination of unit, groupbit and address a switch will be added to Home Assistant called:  "RF2_[unit]_[groupbit])_[address].
As RF2 433Mhz switches do not render their state, no state topic will be provided in the discovery. 
This will cause the switch to be in optimistic mode in HA with separate on and off icons.  
The two separate on/off icons allow for subsequent on commands to support the dimming feature of KAKU switches like ACM-300.

### DIFF
--- a/main/ZgatewayRF2.ino
+++ b/main/ZgatewayRF2.ino
@@ -71,19 +71,19 @@ void setupRF2() {
   digitalWrite(RF_EMITTER_GPIO, LOW);
 }
 
-#  ifdef ZmqttDiscovery 
-//Regsiter for autodiscover in Home Assistant
+#  ifdef ZmqttDiscovery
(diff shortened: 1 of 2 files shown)


1technophile:
Thanks for the PR!
So as to format automatically your document you have to add the "clang-Format" extension to VSCode, once done, you can format the code by doing a right-click into the code file window and clicking "Format document".

hbraam:
Your welcome.
Sorry about the formatting, I ran the clang-format command using the google
style only and missed the '.clang-format' file included in the repository.
Will make sure to use it next time.

Cheers, Herman.

1technophile:
Thanks!

PR #803, opened 2020-11-15
Add ability to specify repeats for raw Pilight strings

melyux:
Enables usage of the new raw repeats parsing in ESPiLight 0.17.0. Now users can specify the `r:` parameter in their raw string to specify how many times their raw message should be repeated (previously, the default was 10, and is still 10 if the user does not specify it).

Also: removes the unused `PilightRAW` macro and adds macOS's `.DS_Store` files to the ignore list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .pio
 .vscode
 *_env.ini
+**/.DS_Store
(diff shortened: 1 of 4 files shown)


1technophile:
Thanks for the PR, would it be possible to document this function below:
https://docs.openmqttgateway.com/use/rf.html#pilight-gateway

melyux:
@1technophile Added instructions to the documentation

1technophile:
Thanks !

PR #804, opened 2020-11-17
Tls ble conflict resolution

animal6496:
First attempt for a solution, it addresses the issue where a BLEscan for BT devices drops the secure MQTT connection thereby blocking publication of BT messages. An ESP32 device running with this branch of OMG is running now for ca. 3 hours without issues.


1technophile (commented):
User_config.h should not be added at the repository level (it is already into the main folder)

Legion2:
This is more a hack than a solution.

1technophile:
I think the principle of having a limited queue when the connection is not available can be useful in several cases:
-ethernet or wifi connection lost
-broker not available

We could change the principle of the gateway from:
-I forward only if I have the broker and throw the messages if it is not available 
to:
-First IN first OUT, with a limited queue that forward the messages when the connection is available

The queue must be limited in time so as to avoid delaying the messages too much from their date of broadcasting.

And by the way, solve the issue with the TLS connection.

What do you think?

Legion2:
Sounds good to me, we have to evaluate if this has implications on how currently the api to send messages is used. Because when we put it in the queue we do not guarantee that it will be delivered.

animal6496:
Approach is fine with me. The current implementation it is a FIFO queue and it allows the user to specify size of the queue at compiler time. In combination with white/black listing this should be OK to keep size of the queue small enough. If there are too many messages for the queue the are ignored. That is not ideal for critical messages e.g. from door/windows or PIR sensors hence you can't use this for DIY alarm systems. There is no time limitation on messages in the queue. For my use case not an issue as after every BLEscan the broker connection is restored and the messages are published. My devices mostly LYWSD03MMC_ATC send messages every 2 minutes so the messages in the queue have a lifetime in the queue of max 2 minutes, also these message are non-critical. Adding a time stamp and filter on the messages should not be too hard. I've limited time to work on the TLS problem, I hope to have more time next week. 

PS I've ESP32's from another brand and on those the implementation (with the new BLE timings) is not stable, I haven't tested the implementation from the PR on those yet. After some random time (a few hours) they lose also the Wifi connection and they a very hard time to reconnect. Often the only way to do this is by power cycling or reset.

Legion2:
I think we should not implement a queue our self, but rather use a standard queue implementation, which is well tested.

animal6496:
I normally would fully agree with you in reusing tested code, however I didn't want to bloat the code with functionality not needed and I don't know how smart the compiler is. More important is my lack of knowledge, I'm also struggling with adding libraries etc. so I'll have to pass here but I'm happy to test any new versions. 
Also haven't been able to resolve the lint issue... I've used the clang-Format module in VScode but the result is not liked by lint.

1technophile:
Thanks for the PR, it is a good first step and answer to the issue with tls + ble.

@Legion2 do you have ideas/recommendations for standard queue implementation?

Legion2:
No I usually don't use C++, but there is a Queue library for Arduino, which should be better than an own implementation.

1technophile:
@animal6496 would you be interested in testing the development branch, a lockless queue was implemented and may solve the TLS issue.

animal6496:
Yes no problem I’ve a few spare esp32’s which I can use.
Is it the latest development branch.

1technophile:
> Is it the latest development branch

Yes it is

animal6496:
OK I'm testing. Because of my earlier experiences with the ESP32's I test with a dedicated power supply as the USB output of the PC results in spurious errors and  instabilities. This also means I don't have a serial out for monitoring available, so I'm following the MQTT message on the broker and router logs. I'm testing with ESP32 WROOM-32 dev boards from different vendors. These boards report slightly different RSSI's (-66 vs -70) .  Note my Wifi and MQTT parameters are directly compiled into the code so I don't need to go via initial access point setup.
First attempt was with undef SECURE_CONNECTION. I was disappointed as I didn't get a connection with the Router with either board. So I started to do some reading into the code and while doing that the board got spontaneously connected after about 30 minutes. I tried the other board (with better RSSI) and that took 12 minutes. Maybe me being impatient. After initial connection I haven't seen this behavior again. Power off/on and the connection to the Wifi is almost instantaneous. Also the connections to the router and the MQTT broker are stable without any problems.
With define SECURE_CONNECTION the situation is different. One board disconnects after publishing the blue tooth observations and immediately reconnects. The MQTT broker reports LWT Offline however the logs of the router do not show a reconnect. Initially the other board showed the same behavior but now for about more than an hour not anymore and the connection to the MQTT broker is stable. I'll keep that board running for a few more days to see how it behaves. So far this looks like a solution.

1technophile:
Thanks for the feedback, this is encouraging!

animal6496:
The gateway is running for three days without any issues. Next weekend I’ll upload it to my ‘production’ devices. Thank you for addressing this issue.

1technophile:
Thanks goes to @csiki2 who killed 2 birds with one stone

PR #805, opened 2020-11-17
Add files via upload

animal6496:
Fixed issue with precompiler options

### DIFF
--- a/main/main.ino
+++ b/main/main.ino
@@ -48,7 +48,6 @@ ReceivedSignal receivedSignal[struct_size] = {{0, 0}, {0, 0}, {0, 0}, {0, 0}};
 unsigned long timer_sys_measures = 0;
 #  define ARDUINOJSON_USE_LONG_LONG 1
 #endif
-
 #include <ArduinoJson.h>
 #include <ArduinoLog.h>
 #include <PubSubClient.h>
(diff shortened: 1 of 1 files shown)


1technophile:
I suppose that this PR is overiden by #804

animal6496:
Hi Florian,
Sorry for being unresponsive. Yes I somehow mixed up things. Also I’ve been having a lot problems with ESP32 platform. It doesn’t seem as stable as ESP8266. First problem is using USB from laptop as power supply, this leads to spurious errors or instabilities on some of the development boards I’m using. A dedicated power supply HiLink HLK PM03 with parallel cap solved the issue. However, some boards keep on struggling with stable WiFi connections. Sometimes they disconnect for long term from the router and refuse to reconnect for sometimes hours. Using a board with an ipx connector and external antenna solved the issue. However, this might also be a result of a more recent revision of the ESP32.  Ralph Bacon on Youtube showed a relation between chip revision and stability. 
So developing a gateway for the BLE devices took much more time than anticipated. I think I’ve stable platform now which hasn’t let me down since Christmas. After a month of stable performance I’ll move thew into their final case and to location in the house (I need two for BLE coverage).

Kind regards,
Rene

PR #818, opened 2020-12-01
ifndef condition for time_avoid_duplicate

jumika:
Allow time_avoid_duplicate to be configured by build_flags.


1technophile:
Thanks for the PR

PR #816, opened 2020-11-29
added ZactuatorSomfy

Legion2:
Add support for Somfy RTS remote control using https://github.com/Legion2/Somfy_Remote_Lib.
close #814

I can only test with ESP32 and CC1101 module. I also created a platformio env for esp8266.

TODO:
- [x] Documentation
- [x] Licence Headers

PR #834, opened 2020-12-13
Npm docs setup

Legion2:
added vuepress to package.json to track the used version and make it reproducible and self contained.

cherry picked from #816


1technophile (approved):
Thanks for the PR

PR #842, opened 2021-01-02
BLE - Disconnect after a timeout

### LINKED ISSUE #798: BLE connection will not be closed
I have noted a high battery usage of the MHO-C401 sensors and one LYWSD03MMC with original firmware.

Looking for the power usage I have seen that after connection to the deices the connection after received notification will not be closed.
So, in my case, the connection will be live for 1:06 minutes (!!) - until next scan, where a 
**BLEDevice::deinit(true);**
will kill all connections.
During this time the sensor take about 3-4mA (peaks between 1.5 and 5,5mA)  what will drain the battery of course.

For MMC is the alternate firmware a solution, but for MHO-C401 not.

Not sure how it can be solved, if connection to device cannot be closed after received notification - maybe make a 
**BLEDevice:deinit(true)** 
after a defined time (as sample max 10 seconds after request)?

Thanks
Michael

1technophile:
So as to save the devices battery

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -569,6 +569,8 @@ void BLEconnect() {
               if (pRemoteCharacteristic->canNotify()) {
                 Log.trace(F("Registering notification" CR));
                 pRemoteCharacteristic->subscribe(true, notifyCB);
+                delay(BLE_CNCT_TIMEOUT);
+                pClient->disconnect();
               } else {
                 Log.notice(F("Failed registering notification" CR));
                 pClient->disconnect();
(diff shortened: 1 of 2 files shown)


main/ZgatewayBT.ino
```diff
@@ -569,6 +569,8 @@ void BLEconnect() {
               if (pRemoteCharacteristic->canNotify()) {
                 Log.trace(F("Registering notification" CR));
                 pRemoteCharacteristic->subscribe(true, notifyCB);
+                delay(BLE_CNCT_TIMEOUT);
```
h2zero, line 572: Instead of a delay here we could use a task notification wait with BLE_CNCT_TIMEOUT as the max wait time. The notification callback could release the task before the timeout and this would save time and battery power for both devices if applicable. Thoughts?

1technophile: Yes good point!

PR #847, opened 2021-01-17
Support for changing CC1101 Receive and Transmit Frequency

NorthernMan54:
Adds the following

1 - Adds a new field on received messages, the MHz of the message ( only visible when compiled for CC1101 ) “mhz”: 303

2 - Add a new message, to set receiver frequency ( optional, default to 433.92 ). Ie { “mhz”: 303.0 }

3 - Add a optional field to transmit message “mhz”

4 - after transmission of a message, the receiver will go back frequency specified in # 2

5 - at this time these changes would only work on the zgatewayrf module.

6 - If an issue occurs while parsing the JSON message an error is transmitted over Mqtt 

Sample message to control a Hampton Bay Fan

```
{"value":1150,"protocol":6,"length":12,"delay":450,"repeat":8,"mhz":303.732}
```

To change receiver frequency

```
{"mhz":315.026}
```

Message Received from a GE Fan Remote on 315Mhz

```
{"value":4534142,"protocol":6,"length":26,"delay":356,"mhz":315.026}
```

Response to setting an invalid frequency

```
{'Status': 'Error'}
```

PS Two thumbs up on your usage of compiler definitions, makes customizing a build really easy


1technophile:
Thanks for the PR, could you update the docs also please : 
https://docs.openmqttgateway.com/use/rf.html#rcswitch-based-gateway

main/ZgatewayRF.ino
```diff
@@ -186,21 +191,63 @@ void MQTTtoRF(char* topicOri, JsonObject& RFdata) { // json object decoding
...
+#    endif
       mySwitch.setRepeatTransmit(valueRPT);
       mySwitch.setProtocol(valuePRT, valuePLSL);
       mySwitch.send(data, valueBITS);
       Log.notice(F("MQTTtoRF OK" CR));
       pub(subjectGTWRFtoMQTT, RFdata); // we acknowledge the sending by publishing the value to an acknowledgement topic, for the moment even if it is a signal repetition we acknowledge also
       mySwitch.setRepeatTransmit(RF_EMITTER_REPEAT); // Restore the default value
+#    ifdef ZradioCC1101 // set Receive off and Transmitt on
+      if (trMhz != valueMhz)
+        ELECHOUSE_cc1101.Init();
```
Legion2, line 211: Why is this call needed here?

main/ZgatewayRF.ino
```diff
@@ -186,21 +191,63 @@ void MQTTtoRF(char* topicOri, JsonObject& RFdata) { // json object decoding
...
     } else {
+#    ifdef ZradioCC1101 // set Receive on and Transmitt off
+      float tempMhz = RFdata["mhz"];
+      if (tempMhz != 0 && validFrequency(tempMhz)) {
+        valueMhz = tempMhz;
+        ELECHOUSE_cc1101.Init();
+        Log.notice(F("RF mhz: %d" CR), (int)valueMhz);
+        pub(subjectGTWRFtoMQTT, RFdata); // we acknowledge the sending by publishing the value to an acknowledgement topic, for the moment even if it is a signal repetition we acknowledge also
+      } else {
+        pub(subjectGTWRFtoMQTT, "{'Status': 'Error'}"); // Fail feedback
```
Legion2, line 222: I don't think this is a valid json string

main/ZgatewayRF.ino
```diff
@@ -186,21 +191,63 @@ void MQTTtoRF(char* topicOri, JsonObject& RFdata) { // json object decoding
...
+        Log.error(F("MQTTtoRF Fail json" CR));
+      }
+#    else
+#      ifndef ARDUINO_AVR_UNO // Space issues with the UNO
+      pub(subjectGTWRFtoMQTT, "{Status: 'error'}"); // Fail feedback
```
Legion2, line 227: this also

main/ZgatewayRF.ino
```diff
@@ -186,21 +191,63 @@ void MQTTtoRF(char* topicOri, JsonObject& RFdata) { // json object decoding
...
 #endif
+
+#ifdef ZradioCC1101
+bool validFrequency(float mhz) {
+  //  CC1101 valid frequencies 300-348 MHZ, 387-464MHZ and 779-928MHZ.
+  if (mhz >= 300 && mhz <= 348)
+    return true;
+  if (mhz >= 387 && mhz <= 464)
+    return true;
+  if (mhz >= 779 && mhz <= 928)
```
Legion2, line 249: all these compare statement have an implicit conversion of int and float.

Legion2 (changes requested):
In the Somfy module the frequency variable should also be used to reset frequency after transmision. Also do a search with an text editor to find all usages of `CC1101_FREQUENCY`.

main/ZgatewayRF.ino
NorthernMan54: I found that when changing the receive frequency the module needed to be re-initialized in order for the change to take effect.  Checking the CC1101 data sheet and the examples, I couldn't find an example with this use case, so this is a bit brute force.

main/ZgatewayRF.ino
```diff
@@ -186,21 +191,63 @@ void MQTTtoRF(char* topicOri, JsonObject& RFdata) { // json object decoding
       Log.notice(F("RF Protocol:%d" CR), valuePRT);
       Log.notice(F("RF Pulse Lgth: %d" CR), valuePLSL);
       Log.notice(F("Bits nb: %d" CR), valueBITS);
+#    ifdef ZradioCC1101 // set Receive off and Transmitt on
+      float trMhz = RFdata["mhz"] | CC1101_FREQUENCY;
+      if (validFrequency(trMhz)) {
+        ELECHOUSE_cc1101.SetTx(trMhz);
+        Log.notice(F("RF mhz: %d" CR), (int)trMhz); // this is a float but no easy way to convert
```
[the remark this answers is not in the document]
NorthernMan54: When I read the ArduinoLog docs float wasn't an option, but after looking at the platformio version, found the float option.

NorthernMan54:
@Legion2 'In the Somfy module the frequency variable should also be used to reset frequency after transmission'

I was looking at this and am thinking would need a global variable for this.  Any other thoughts on how to share the value between modules?

Legion2:
Yes, the value must be shared, I think the best place is the config_RF.h, which is already shared.

NorthernMan54:
@Legion2 support for Somfy going back to the global receive frequency has been added.  I don’t have the device so I can’t do a physical test, but the build worked.  I also named the global variable receiveMhz as that made the most sense.

Any other comments or feedback?

main/ZactuatorSomfy.ino
```diff
@@ -71,7 +71,9 @@ void MQTTtoSomfy(char* topicOri, JsonObject& jsonData) {
     SomfyRemote somfyRemote(RF_EMITTER_GPIO, somfyRemotes[remoteIndex], &rollingCodeStorage);
     somfyRemote.sendCommand(command);
 #    ifdef ZradioCC1101
-    ELECHOUSE_cc1101.SetRx(CC1101_FREQUENCY); // set Receive on
+    if (CC1101_FREQUENCY_SOMFY != receiveMhz)
+      ELECHOUSE_cc1101.Init();
```
Legion2, line 75: why is `Init()` called here?

NorthernMan54: During testing I found that the frequency was not taking effect, and initialing the transceiver resolved the issue.  Reviewed the CC1101 data sheets and examples and did not see this scenario covered.  So it is a bit of hack.

Legion2: I'm using the Somfy module and it sends with 433.42 MHz and then switches back to 433.92 MHz to receive PiLight signals. And it works without calling `ELECHOUSE_cc1101.Init();`.

NorthernMan54: During my testing I was jumping between 433 MHz and 303 MHz, and I found that when just doing a SetRx() that it was not picking up transmissions.  After adding the Init() it started working.  Am thinking that small frequency jumps may be okay ( your experience ) but large jumps may not be.  But without knowing a valid range, its a crap shot.

Legion2: Ok, `Init()` fixed the symptoms, but I don't think this is the best solution to the problem. Calling Init restarts the SPI connection and the CC1101 module https://github.com/LSatan/SmartRC-CC1101-Driver-Lib/blob/master/ELECHOUSE_CC1101_SRC_DRV.cpp#L141

NorthernMan54: After reviewing the data sheet further and looking into why having different transmit and receive frequencies worked, deduced that the receiver needed to be idled when changing receive frequencies.  This line here [here](https://github.com/LSatan/SmartRC-CC1101-Driver-Lib/blob/f860157b17b2449768d18ca46c6cb5744bad1971/ELECHOUSE_CC1101_SRC_DRV.cpp#L975)

After a few hours of testing, this resolves it, and removes the kludge approach

NorthernMan54: PS Raised this a question against https://github.com/LSatan/SmartRC-CC1101-Driver-Lib/issues/65

Legion2: Nice, thanks for looking into this and raising the issue in the library project. We can merge it with this workaround but should keep in mind to update to the new library version when available.

Legion2 (approved):

1technophile:
Thanks guys

NorthernMan54:
Tks very much gentlemen

PR #855, opened 2021-01-30
[Breaking] Publish state to same channel as received signals

qistoph:
Received signals are published to the subjectRF2toMQTT topic (/RF2toMQTT). However, echos of mqtt messages are published to the topic subjectGTWRF2toMQTT (/433toMQTT).

That seems quite inconsistent and breaks functionality in Home Assistant that expects confirming messages, like this echo, on the same topic as receive messages.

This PR fixes this.

**Breaking change**
This PR changes the topic of echo messages. Configurations expecting the previous topic `433toMQTT`, will break.

Alternatively it might be an option to publish echos to both topics or make the topic configurable. Please let me know if such changes are preferred.


1technophile:
That's coherent, thanks for the proposal

PR #861, opened 2021-02-06
Update readme for better understanding in English

anarchking:
Not sure if these changes affect other translations... But I made a few corrections for better understanding in english


1technophile:
Thanks !

PR #863, opened 2021-02-06
Added TTGO-T-Beam board + option to send ADC values in volt 

hallard:
- option to send ADC values in volt (for battery voltage for example) 
- added TTGO Beam board

For voltage conversion to work, you  just need to define ADC_DIVIDER=2 (example if Resistor divider is 2 because of 100K+100K on beam


1technophile:
Thanks for the PR!

hallard:
You're welcome, some more to come :-)

PR #869, opened 2021-02-13
added room presence yaml to docs

anarchking:
Found how to set this up in the issues, so I figured I'd add it to the docs because it works great for me


1technophile:
Thanks!

PR #872, opened 2021-02-20
LoRa packets lost - BME280 sensor fix

clabnet:
Set syncword and spreadfactor on receiving to avoid lost packets. 
Used with TTGO LORA v1.

Add custom I2C pins for settings BME280 sensor connection.


1technophile:
Thanks for the PR, would it be possible to divide it into 2 PR, one for the BME280, the other for LORA?

PR #882, opened 2021-03-03
Remove wifi protocol auto change

1technophile:
#575
so as to see if in conjunction with the platforms update we get a better stability of wifi.
We are using this function since v0.9.3 but in my point of view, it should be handled at a lower level. 
I'm afraid that keeping it at the application level is creating more instability than stability. So let's remove it and test this branch and maybe the v0.9.6beta like that.
We will see if we keep it removed for the v0.9.6


1technophile:
@maxi1134 @wimpie007 @meylux would it be possible to test this branch to see if you get better or keep stability in your wifi reconnection?

PR #897, opened 2021-03-14
[BREAKING] Change uptime unit from seconds to minutes

1technophile:
to prevent going to the max of unsigned long

### DIFF
--- a/main/ZmqttDiscovery.ino
+++ b/main/ZmqttDiscovery.ino
@@ -174,7 +174,7 @@ void pubMqttDiscovery() {
   createDiscovery("sensor", //set Type
                   subjectSYStoMQTT, "SYS: Uptime", (char*)getUniqueId("uptime", "").c_str(), //set state_topic,name,uniqueId
                   "", "", "{{ value_json.uptime }}", //set availability_topic,device_class,value_template,
-                  "", "", "m", //set,payload_on,payload_off,unit_of_meas,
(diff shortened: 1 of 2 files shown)


csiki2:
Even the signed long (2^31 sec) can hold more than 68 years, do I miss something here? :)

1technophile:
Thanks for pointing this out, you are right the cause of the issue is not the variable type, it is the limit of millis(). I will find another way.

csiki2:
Something like below?
<pre>
static unsigned long lastUptime = 0;
static unsigned long uptimeAdd = 0;
unsigned long uptime = millis() / 1000 + uptimeAdd;
if (uptime &lt; lastUptime) {
  uptime += 4294967;
  uptimeAdd +=4294967;
}
lastUptime = uptime;
</pre>
I don't know where millis() roll over, I expect to be unsigned long here.

csiki2:
Hmm. millis() seem to use
int64_t esp_timer_get_time();
So actually, we have a proper uptime source, but loose during millis():
<pre>
unsigned long IRAM_ATTR millis()
{
    return (unsigned long) (esp_timer_get_time() / 1000ULL);
}
</pre>
The propes code still should work though.

1technophile:
Thanks, your first code will work also with Arduino boards, so it will avoid adding more board-specific code. I will push a PR with it.

PR #918, opened 2021-04-06
ZGatewayRF white/blacklisting

mtiutiu:
Basic whitelisting/blacklisting support for the **ZGatewayRF**. Nothing fancy or complicated just a **simple filtering** based on the **unique RC switch code** being sent. This avoids creating in Home Assistant for example redundant entities because the RF receiver "catches signals" from other devices around the house or neighbors.

I tested both whitelisting/blacklisting and seems to work as expected. Disabling whitelisting/blacklisting is also supported. 

Hope it helps.


mtiutiu:
Too many files changed..something is wrong with the PR. Will create another one.

Issue #568, opened 2020-03-13
Filter 433 commands so that not all are published to MQTT.


wimpie007:
I get a lot of 'rubbisch' 433 commands on my gateway.
Polution you might call it. 
All these events are published to MQTT, wich leads to a lot of activity in my nodered/ HA-automations wich actually do nothing.
Also, for this reason, HA-autodiscovery is unusable: i used it a week, and got HUNDREDS of unused switches in HA.

**Describe the solution you'd like**
a filter-mechanism. (whitelist for example protocol-length(s). 24 would - in my case - work in almost all cases)

**Describe alternatives you've considered**
*dirty*-hack the code myself. :) You don't like that!:)

1technophile:
Indeed we could replicate the mechanism used for white & black list management of BLE devices

jumika:
+1 for this. My mqtt boker is flooded with all the unknown 433mhz device signals.

DigiH:
Hi all, I have already implemented a flag for suppressing unparsed 433/868 signals to the MQTT publish queue in NothernMan's rtl_433_ESP a while back. It's documented on https://github.com/NorthernMan54/rtl_433_ESP, but obviously didn't make it into the OMG documentation.

https://github.com/NorthernMan54/rtl_433_ESP/pull/7

While not a proper white/black-list implementation, it might help remedy your issues as well.

The build flag is

'-DSUPPRESS_UNPARSED=true'

Even with a black-list, this might still be needed, to avoid the MQTT publishing of unparsed/unknown devices.

Hope this helps.

PR #919, opened 2021-04-06
Basic whitelisting/blacklisting support for ZgatewayRF

mtiutiu:
Basic whitelisting/blacklisting support for the ZGatewayRF. Nothing fancy or complicated just a simple filtering based on the unique RC switch code being sent. This avoids creating in Home Assistant for example redundant entities because the RF receiver "catches signals" from other devices around the house or neighbors.

I tested both whitelisting/blacklisting and seems to work as expected. Disabling whitelisting/blacklisting is also supported.

Hope it helps.

### DIFF
--- a/main/ZgatewayRF.ino
+++ b/main/ZgatewayRF.ino
@@ -30,18 +30,18 @@
 #ifdef ZgatewayRF
 
 #  ifndef ARDUINO_AVR_UNO // no whitelist/blacklist support for Arduino UNO - doesn't fit
-#  include <vector>
+#    include <vector>
 using namespace std;
 
 static vector<SIGNAL_SIZE_UL_ULL> whitelisted_devices;
 static vector<SIGNAL_SIZE_UL_ULL> blacklisted_devices;
 
-#define isWhitelistedDevice(id) checkIfDevicePresentInList(id, whitelisted_devices)
-#define isBlacklistedDevice(id) checkIfDevicePresentInList(id, blacklisted_devices)
+#    define isWhitelistedDevice(id) checkIfDevicePresentInList(id, whitelisted_devices)
+#    define isBlacklistedDevice(id) checkIfDevicePresentInList(id, blacklisted_devices)
 #  endif
 
 #  ifdef ZradioCC1101
-#  include <ELECHOUSE_CC1101_SRC_DRV.h>
+#    include <ELECHOUSE_CC1101_SRC_DRV.h>
 #  endif
 
 #  include <RCSwitch.h> // library for controling Radio frequency switch
@@ -156,11 +156,11 @@ void RFtoMQTT() {
 
 #  ifdef simpleReceiving
 void MQTTtoRF(char* topicOri, char* datacallback) {
-#  ifdef ZradioCC1101 // set Receive off and Transmitt on
+#    ifdef ZradioCC1101 // set Receive off and Transmitt on
   ELECHOUSE_cc1101.SetTx(receiveMhz);
   mySwitch.disableReceive();
   mySwitch.enableTransmit(RF_EMITTER_GPIO);
-#  endif
+#    endif
   SIGNAL_SIZE_UL_ULL data = STRTO_UL_ULL(datacallback, NULL, 10); // we will not be able to pass values > 4294967295 on Arduino boards
 
   // RF DATA ANALYSIS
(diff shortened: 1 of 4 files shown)


mtiutiu:
Made it to compile for the AVR arch but **no ARDUINO UNO** support due to limited resources available.

mtiutiu:
Should fix #568 as well. As a future enhancement I think I can add the ability to filter based on a range like this:
```
{"white-list": ["110355-220000"]}
```
instead of writing all those id's by hand which can be tedious when you have many around. Same for blacklisting. But let's keep this PR as it is right now because as I already mentioned in the title it adds `Basic` functionality.

mtiutiu:
It's quite inefficient to send white/black list via MQTT and JSON due to RAM constraints for embedded devices. And the current implementation that I made may not be efficient as well because of the fact that one cannot send long lists of id's. I updated the readme with the details. It should be more efficient when using ranges like I mentioned in the previous comment. Yet again I will opt for a web interface that the Gateway should expose and do some configuration things there which get stored on the internal flash. This should be not that trivial to implement for the ESP32/8266 platforms.

1technophile:
Thanks for the PR, I will look at it next week

platformio.ini
```diff
@@ -82,6 +82,7 @@ extra_configs =
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 [libraries]
+arduino_stl = ciband/avr_stl @ ^1.2.3
```
1technophile, line 85: For Zgateway BT White and blacklist mechanism we use:
```c
`stl = https://github.com/mike-matera/ArduinoSTL.git#7411816`
```
The one that you are proposing seems more maintained. Would it be possible to do the replacement into platformio.ini so as to use only yours?

main/ZgatewayRF.ino
```diff
@@ -29,6 +29,17 @@
 
 #ifdef ZgatewayRF
 
+#  if defined(ESP32) || defined(ESP8266) || defined(__AVR_ATmega2560__) || defined(__AVR_ATmega1280__)
+#    include <vector>
+using namespace std;
```
1technophile, line 34: We have a difference between your macro segmentation and the one in ZgatewayBT:
```
#  if !defined(ESP32) && !defined(ESP8266)
#    include <ArduinoSTL.h>
#  endif

#  include <vector>
using namespace std;
```

It could be interesting to unify both as we are talking about the same function.

1technophile:
Thanks for the PR, we may have in a future PR to think about factoring white and blacklist functions between the different gateways.
I like also your add regarding black and white list removal, we should add it for BLE

1technophile:
> Should fix #568 as well. As a future enhancement I think I can add the ability to filter based on a range like this:

Could be a nice option

main/ZgatewayRF.ino
mtiutiu: Right, this one is more concise and shorter. Will do.

platformio.ini
mtiutiu: Sure.

josephperry:
Don't know how to submit a pull request. :-(

I solved this for me by adding in platformio.ini:
build_flags =
  ${com-esp.build_flags}
  '-DZgatewayRF="RF"'
  '-DZRFReceiveProtocol1=1'
  '-DZRFReceiveBitlength1=24'
  '-DZgatewayIR="IR"'
  '-DZactuatorONOFF="ONOFF"'
  '-DACTUATOR_ON=1'
  '-DsimplePublishing=true'
  '-DRFmqttDiscovery=true'
  '-DZmqttDiscovery="HADiscovery"'

and in ZgatewayRF.ino insert in function RFtoMQTT() after mySwitch.resetAvailable();
    //;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; 
   // Restricts reporting if ZRFReceiveProtocol1 and/or ZRFReceiveBitLength1 are defined
    //;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 #  if defined (ZRFReceiveProtocol1) && defined (ZRFReceiveBitlength1)
    // ZRFReceiveProtocol1 and ZRFReceiveBitlength1 must both match if both are defined
    if (mySwitch.getReceivedProtocol() != ZRFReceiveProtocol1 || mySwitch.getReceivedBitlength() != ZRFReceiveBitlength1) {
      return;
    }
 #  elif ZRFReceiveProtocol1     // only ZRFReceiveProtocol1 is defined
    if (mySwitch.getReceivedProtocol() != ZRFReceiveProtocol1)
      return;
 #  elif ZRFReceiveBitlength1    // only ZRFReceiveBitlength1 is defined
    if (mySwitch.getReceivedBitlength() != ZRFReceiveBitlength1)
      return;
 #  endif
    //;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

This way can restrict reporting of RF signals my specifying the RFReceiveProtocol and/or RFReceivedBitlength to avoid creating many false switches in Home Assistant.

josephperry:
Expanding on my previous suggestion.  

RF noise has caused many false switches to be created in Home Assistant.  This suggestion allows filtering of RF switches by specifying the RF protocol with corresponding RF bit length and/or by RF protocol and/or or by RF bit length.  It does NOT filter by the RF receive code. Four RF Receive Protocols and RF Receive Bit lengths are supported.

If there is a similar need for filtering IR signals, I believe a similar approach could be used in ZgatewayIR.ino.

In platform.ini:
`build_flags =
  ${com-esp.build_flags}
  '-DZgatewayRF="RF"'
  '-DZRFReceiveProtocol1=1'
  '-DZRFReceiveBitlength1=24'
  '-DZRFReceiveProtocol2=2'
  '-DZRFReceiveBitlength2=32'
  '-DZRFReceiveProtocol3=7'
;  '-DZRFReceiveBitlength3=30'
;  '-DZRFReceiveProtocol4=9'
  '-DZRFReceiveBitlength4=48'
   '-DjsonReceiving=true'
  '-DRFmqttDiscovery=true'
  '-DZmqttDiscovery="HADiscovery"'`

In file ZgatewayRF.ino insert this code in function RFtoMQTT() after mySwitch.resetAvailable();
[joe_ZgatewayRFino.txt](https://github.com/1technophile/OpenMQTTGateway/files/7636519/joe_ZgatewayRFino.txt)

PR #924, opened 2021-04-14
Typo fix

DigiH:
Typo in sending status command to MQTTtoRTL_433. "-1" should be "1" as it produces an error otherwise.


1technophile:
Thanks !

PR #934, opened 2021-04-28
Usability Improvements for esp32 platform

NorthernMan54:
I had been using these for a while with my devices, and wanted to share these usability improvements for others

1 - Ability to set **SERIAL_BAUD** via a compiler directive and platformio ini file, to allow using the higher serial speeds available with the ESP platform.

2 - In **ESPWifiManualSetup** add the ability to define a second SSID/Password via compiler directive wifi_ssid1/wifi_password1 and have the unit select the available one on startup.  For both ESP8266 and ESP32 platforms.

3 - Add support for MDNS_SD for lookup of mqtt service address on ESP32 platform.

4 - Resolve issue with MDNS_SD not working on esp8266 platform.  Arduino OTA was initializing MDNS and causing MDNS.begin to fail.  As this feature may not be widely used I'm not sure if this issue was reported.

### DIFF
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -327,9 +327,9 @@ int lowpowermode = DEFAULT_LOW_POWER_MODE;
 #ifdef ZgatewaySRFB
 #  define SERIAL_BAUD 19200
 #else
-#ifndef SERIAL_BAUD
-#  define SERIAL_BAUD 115200
-#endif
+#  ifndef SERIAL_BAUD
+#    define SERIAL_BAUD 115200
+#  endif
 #endif
 /*--------------MQTT general topics-----------------*/
 // global MQTT subject listened by the gateway to execute commands (send RF, IR or others)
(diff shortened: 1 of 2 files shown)


1technophile:
Thanks for the PR!

1technophile:
Did you test it in different configurations, with/without MDNS, with/without advanced network config?

NorthernMan54:
Yes had tested both configurations on ESP32 and ESP8266

NorthernMan54:
Tks, now onto the Active receiver switching feature

My backlog

[ ] Active switching between signal receivers ( RF, PiLight and RTL_433 )
[ ] Home Assistant Discovery for RTL_433
[ ] Black/White listing for RTL_433
[ ] FSK Modulation support for rtl_433_ESP ( waiting for feedback on the feature )

Legion2:
This broke the `ESPWifiManualSetup` feature in combination with the `SECURE_CONNECTION` feature, because `WiFiMulti wifiMulti;` is not defined when `SECURE_CONNECTION` is enabled but it is used in `ESPWifiManualSetup`.

Issue #939, opened 2021-05-07
Missing "val_tpl" on create discovery switch provided by RF 


Odyno:
**Describe the bug**
on Home Assistant the switch created by discovery process is present **but the state is blocked**

**Environment:**
- I have a Digoo Pir sersor that send messages to OMG
- OMG is built with `'-DZgatewayRF="RF"'` and  ` '-DZmqttDiscovery="HADiscovery"'` on development branch
- the board is esp32dev
- the message is received by OMG and sent to MQTT broker
- on Home Assistant the switch is present **but the state is blocked**
 
**Expected behavior**
Change the state to "on" in Home Assistant foreach revelation

**Bug Analysis**
The published message has the format "C" but the configuration provided by OMG (A) has no `val_tpl` field (B), so the `pl_on` can never match 

![Screenshot 2021-05-07 at 02 04 11](https://user-images.githubusercontent.com/1727138/117381549-9a517580-aedc-11eb-862f-4bfbd6b8aa51.png)

PR #941, opened 2021-05-07
Add val_tpl on HA discovery configuration - fix issue #939

Odyno:
fix issue #939


1technophile:
thanks for the proposal, 
I would like to know how the switch goes back to OFF in this case?

Odyno:
This is an interesting point, it all started here. 

**In short:** 
I created one Automation on HA that reset the switch... below the code
```
alias: Reset All Pirs
description: ''
trigger:
  - platform: state
    entity_id:
      - switch.2741642
      - switch.11031942
      - switch.....
    to: 'on'
    for: '00:00:05'
condition: [ ... ... ... ]
action:
  - delay:
      hours: 0
      minutes: 0
      seconds: 5
      milliseconds: 0
  - service: switch.turn_off
    target:
      entity_id: '{{ trigger.entity_id }}'
```

**With more context:**
Considerations that led me to this PR:

without the "val_tpl" no automatism will change the state of the switch, even if a "pl_on" is defined (and it is equal to the value) and even if on the MQTT topic we receive the update from "the real switch". So, in the end, we have a switch created by discovery that forces me into constraints (the "pl_on") and it is always stationary ( so... in my opinion just confusing and it is useless).

On the other hand, there is no info on pl_off, but is present the cmd_t that indicate how to change the state... so finally all are free to switch manually the sensor to something that is "on" (as specified on "pl_on") or not "on" ( so.... it can be just off).

If I were to look at the configuration without the "val_tpl"  (according to the point of view) it is useless. If I try to interpret this "sensor", by giving the value "val_tpl", it clearly becomes the modelling of the sensor that provide triggers. 

The sensor provide just the "event on" and is up to listener reset the trigger.

Open to discuss it!

1technophile:
ok, I see, maybe it could be nice to add to the docs (integrate/home assistant) your script that you are using with this PR so as to set the status to OFF?

Odyno:
ok!

1technophile:
Thanks, could you rebase from development branch please

Odyno:
done

Odyno:
@1technophile let me stop this PR, I just discovered that this configuration create a lot of error on home assistant, because sometime the value is not present and on home assistant is raised a lot of warning and error

```
2021-05-13 18:37:56 ERROR (MainThread) [homeassistant.helpers.template] Template variable error: 'value_json' is undefined when rendering '{{ value_json.value | is_defined }}'
2021-05-13 18:37:56 ERROR (MainThread) [homeassistant.helpers.template] Error parsing value: 'value_json' is undefined (value: OFF, template: {{ value_json.value | is_defined }}) 
```

PR #942, opened 2021-05-07
Enable the BME280 options to be settable using PlatformIO build flags, add and organize documentation

### LINKED ISSUE #937: Platform.io build flag for TimeBetweenReadingbme280 not working, and special characters in Wi-Fi password
**Describe the bug**
I want to set the "TimeBetweenReadingbme280" value in config_BME280.h using build flags instead of directly editing the config file. I can set User_config.h and config_RF.h variables using build flags, but having e.g. `'-DTimeBetweenReadingbme280=60000'` has no effect; it still uses the default 30000 (30 seconds) value.

Now that I look at it, I guess the problem is that `config_BME280.h` doesn't surround the variable definitions with `#ifndef TimeBetweenReadingbme280` like everything else. So the default value overwrites the `build_flags` setting. The fix would be just add that `#ifndef` wrapper. I just tried doing that, and it actually worked, so I might actually just put in a PR myself to do this to close out this issue.

Also, it was difficult to put special characters for `-Dwifi_password=`, like dollar sign, ampersand, single quotes, etc. I was getting errors like this no matter what I did:
```
ValueError: No closing quotation:`
```
or
```
sh: -c: line 0: unexpected EOF while looking for matching `''
```
I finally solved it by triple escaping the characters. Like this for the password `BlahBlah'sBro$Bro`:
```
-Dwifi_password=\"BlahBlah\\\'sBro\\\$Bro\"
```
This could probably be useful to add into the docs. It also works for setting the `Gateway_AnnouncementMsg` variable in the build flags, and probably all of them. My PR could also include this.

By the way, what's the difference between using `build_flags` like `-Dkey=value` or `-D key=value` or `'-Dkey=value'` or `'-D key=value'`? The spacing and the single quotes-wise. I can't seem to find any kind of information on the internet about how Platform.io deals with this. Let me know if you guys know anything.

**To Reproduce**
1. Create an .ini file with build flag `'-DTimeBetweenReadingbme280=60000'`
2. Also set `-Dwifi_password` to something with a dollar sign and single quotes and other special characters

**Expected behavior**
`TimeBetweenReadingbme280` is set to the specified value in the build flags.

* OpenMQTTGateway version used: Latest `development` branch pulled a few minutes ago.
* Library version related to the problem you have (if you have troubles with RF provide the version of RCSwitch library): BME280

melyux:
I've enabled the BME280 options to be set by PlatformIO build flags. Also, I've done some minor documentation reorganization for the section on compiling your own builds, specifically with PlatformIO. Some English grammar and spelling and link fixes as well.

Should resolve #937


1technophile (approved):
Thanks!

PR #943, opened 2021-05-07
Automate setting OMG_VERSION

NorthernMan54:
This script will set the OMG_VERSION to a string containing the build
env name, OMG Release number, gihthub release tag and git branch.  And
to invoke it needs to be added as an extra script in the env section of
platform.ini

extra_scripts = pre:scripts/omg_firmware_version.py

I.e.

esp32dev_cccccc-v0.9.6-12-gdea42b3[development-switching]

Tested with MacOS, will need validation with a windows environment.  If
it fails validation on windows am thinking to just comment it out in
platformio.ini and add a comment indicating the feature and optionally
enable.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -136,6 +136,7 @@ build_flags =
   '-DjsonReceiving=true'
 ;  '-DLOG_LEVEL=LOG_LEVEL_TRACE'  ; Enable trace level logging
 monitor_speed = 115200
+extra_scripts = pre:scripts/omg_firmware_version.py
 
 [com]
 esp8266_platform = espressif8266@2.6.3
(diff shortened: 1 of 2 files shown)


1technophile:
There is a command that does it at each release below:
https://github.com/1technophile/OpenMQTTGateway/blob/9fb3d90da9d5a84d8da0e68d9096c144d2a10c3e/.github/workflows/release.yml#L22

I understand that yours is built-dependent and not release-dependent isn't it?
I think this is nice to populate this on each build, but we should update the yml script above.

NorthernMan54:
Oh, I wasn't aware of that feature as I had been doing builds within platformio and OMG_VERSION in my devices was not being set.  As you already have something, I will just close this and keep the script for use in my builds.

PR #945, opened 2021-05-15
Cleanup compiler warnings in strict environment.

h2zero:
This clears up many compilation issues when using -Wall -Werror=all -Wextra options.

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -277,15 +277,15 @@ void dumpDevices() {
 }
 
 void strupp(char* beg) {
-  while (*beg = toupper(*beg))
+  while ((*beg = toupper(*beg)))
     ++beg;
 }
 
 #  ifdef ZmqttDiscovery
-void MiFloraDiscovery(char* mac, char* sensorModel) {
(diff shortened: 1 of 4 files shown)


main/ZgatewayBT.ino
```diff
@@ -1029,7 +1035,7 @@ void launchBTDiscovery() {
       String macWOdots = String(p->macAdr);
       macWOdots.replace(":", "");
       Log.trace(F("Launching discovery of %s" CR), p->macAdr);
-      if (p->sensorModel == HHCCJCY01HHCC) MiFloraDiscovery((char*)macWOdots.c_str(), "HHCCJCY01HHCC");
```
1technophile, line 1032: I'm not sure why the miflora doesn't need the (char *) here, should it be removed for all discovery calls?

h2zero: Yes, that does seem odd doesn't it. The reason was I simply did a quick compile check with just one and forgot to go back and change the rest. I'll remove the others and push the change.

h2zero: Done

1technophile:
Thanks!

PR #946, opened 2021-05-15
Remove unsubscribe and disconnect calls in BLEconnect.

h2zero:
These calls were unnecessary as the client object swill be deleted when finished with the connection.


h2zero:
Not sure what happened here, closing and opening new

PR #947, opened 2021-05-15
Remove reinitializing the BLE stack on each scan.

h2zero:
Using NimBLE requires less resources than bluedroid so it is much less beneficial to deinitialize the stack. Instead this initializes it once in initial setup and only de-initializes before entering deep sleep.


h2zero:
Not sure what happened here, closing and opening new

PR #953, opened 2021-05-21
Trigger firmware update via MQTT. (ESP Only)

h2zero:
If enabled in user_config.h this will trigger the OMG to update it's firmware from the link
provided in the MQTT message.

Example message:
```
{"version": "test ",
 "password": "OMGPASSWORD",
 "url": "https://github.com/1technophile/OpenMQTTGateway/releases/download/v0.9.6/esp32dev-ble-cont-firmware.bin"
}
```
If the firmware version in the message is not the same as the current OMG version the device will self update using the link in the url of the message.


main/User_config.h
```diff
@@ -74,11 +74,24 @@ const byte subnet[] = {255, 255, 255, 0};
...
+//#  define MQTT_HTTPS_FW_UPDATE //uncomment to enable updating via mqtt message.
 #else // for arduino boards
 const byte ip[] = {192, 168, 1, 99};
 const byte mac[] = {0xDE, 0xED, 0xBA, 0xFE, 0x54, 0x95}; //W5100 ethernet shield mac adress
 #endif
 
+#ifdef MQTT_HTTPS_FW_UPDATE
+//If used, this should be set to the root CA certificate of the server hosting the firmware.
+// The certificate must be in PEM ascii format
+const char* https_fw_server_cert = CERT_ATTRIBUTE = R"EOF("
```
1technophile: I think CERT_ATTRIBUTE  macro should be defined before if MQTT_HTTPS_FW_UPDATE || SECURE_CONNECTION are defined

h2zero: Good timing, was fixing this right now (pushed before all commits). Will be updated in a minute.

h2zero: While I'm at it, maybe that macro should be removed? Esp32 and 8266 code compile fine with just PROGMEM being there.

1technophile: ok!

main/main.ino
```diff
@@ -1613,6 +1622,65 @@ void receivingMQTT(char* topicOri, char* datacallback) {
...
+  if (strstr(topicOri, subjectFWUpdate) != NULL) {
+    if (strcmp(HttpsFwUpdateData["version"], OMG_VERSION) != 0) {
+      const char* url = HttpsFwUpdateData["url"];
+      if (!strstr((url + strlen(url - 5)), ".bin")) {
+        Log.error(F("Invalid firmware extension" CR));
+        return;
+      }
+
+      Log.warning(F("Starting firmware update" CR));
+      WiFiClientSecure update_client;
```
h2zero: TODO: Check if link is https or http and use WiFiClientSecure or WiFiClient respectively.

main/main.ino
```diff
@@ -1613,6 +1622,65 @@ void receivingMQTT(char* topicOri, char* datacallback) {
...
+#  endif
+void MQTTHttpsFWUpdate(char* topicOri, JsonObject& HttpsFwUpdateData) {
+  if (strstr(topicOri, subjectFWUpdate) != NULL) {
+    if (strcmp(HttpsFwUpdateData["version"], OMG_VERSION) != 0) {
+      const char* url = HttpsFwUpdateData["url"];
+      if (!strstr((url + strlen(url - 5)), ".bin")) {
+        Log.error(F("Invalid firmware extension" CR));
+        return;
+      }
+
```
1technophile, line 1650: If you are interested there is a function to stop BLE processing, stopProcessing()
so as avoid antenna usage during the OTA process

1technophile:
Would be nice to update the docs also:
https://docs.openmqttgateway.com/use/gateway.html

1technophile:
I think defining 2 test environments (ESP32 & ESP8266) could help for the CI of this, here is an example:
https://github.com/1technophile/OpenMQTTGateway/blob/4706d75adc4bf09e3a7616c90911e8ff2ffb8dfd/platformio.ini#L825
Or we may update these ones so that they cover also OTA

main/main.ino
h2zero: Yep, I was thinking about this as well, adding to TODO list.

main/User_config.h
h2zero: Done, macro removed and I just pushed the changes.

h2zero:
> 
> 
> Would be nice to update the docs also:
> https://docs.openmqttgateway.com/use/gateway.html

Yes that can be done as well, once the process is finalized. I think adding a port parameter to the JSON message should be added and therefore would need to be documented as well.

h2zero:
> 
> 
> I think defining 2 test environments (ESP32 & ESP8266) could help for the CI of this, here is an example:
> 
> https://github.com/1technophile/OpenMQTTGateway/blob/4706d75adc4bf09e3a7616c90911e8ff2ffb8dfd/platformio.ini#L825
> 
> Or we may update these ones so that they cover also OTA

Good plan, will do.

main/main.ino
```diff
@@ -1613,6 +1622,65 @@ void receivingMQTT(char* topicOri, char* datacallback) {
...
+#ifdef MQTT_HTTPS_FW_UPDATE
+#  include <WiFiClientSecure.h>
+#  ifdef ESP32
+#    include <HTTPUpdate.h>
+#  elif ESP8266
+#    include <ESP8266httpUpdate.h>
+#  endif
+void MQTTHttpsFWUpdate(char* topicOri, JsonObject& HttpsFwUpdateData) {
+  if (strstr(topicOri, subjectFWUpdate) != NULL) {
+    if (strcmp(HttpsFwUpdateData["version"], OMG_VERSION) != 0) {
```
1technophile: Maybe we could add also a key and a check against the ota_password for security purposes.

h2zero: Ok, will do!

main/main.ino
```diff
@@ -1613,6 +1622,65 @@ void receivingMQTT(char* topicOri, char* datacallback) {
...
+      if (strlen(https_fw_server_cert) < 200) {
+        Log.error(F("Invalid certificate - using insecure" CR));
+        update_client.setInsecure();
+      } else {
+        update_client.setCACert(https_fw_server_cert);
+      }
+#    endif
+      update_client.setTimeout(12);
+      /****** For Future use, does not work with current releases of arduino core ******/
+      // httpUpdate.setFollowRedirects(HTTPC_STRICT_FOLLOW_REDIRECTS);
```
h2zero: To work around this issue for older core (fix is only a month old) I could derive a class to incorporate the fix. Would need extra files or a separate library.

docs/use/gateway.md
```diff
@@ -33,6 +33,19 @@ If you want the settings to be kept upon gateway restart, you can publish the co
...
+The gateway can be updated through an MQTT message by providing a JSON formatted message with a version number, OTA password (optional, see below), and URL to fetch the update from.  
+
+To enable this functionality, `MQTT_HTTPS_FW_UPDATE` will need to be defined or the line that defines in in user_config.h will need to be uncommented.
+
+::: tip
+If using an unsecure MQTT broker it is **highly recommended** to disable the password checking by setting the macro `MQTT_HTTPS_FW_UPDATE_USE_PASSWORD` to 0 (default is 1 (enabled)), otherwise a clear text password may be sent over the network.
+:::
+
+### Example firmware update message:
+`mosquitto_pub -t firmware_update -m "{\"version\":\"version_tag\",\"password\":\"OTAPASSWORD\",\"url\":\"https://github.com/1technophile/OpenMQTTGateway/releases/download/v0.9.6/esp32-m5stack-ble-firmware.bin\"}"`
```
1technophile: I would remove the backslash from the MQTT message I'm not sure we need them and replace version_tag by test, 
```json
{
  "version": "test",
  "password": "OTAPASSWORD",
  "url": "https://github.com/1technophile/OpenMQTTGateway/releases/download/v0.9.6/esp32-m5stack-ble-firmware.bin"
}
```

The example topic could be presented as:
```
home/<gateway_name>/commands/firmware_update
```

h2zero: Ok, will do.

h2zero: done

PR #956, opened 2021-05-24
Fixed WiFiMulti when secure connection is enabled

Legion2:
fix bug introduced in  #934


NorthernMan54:
@Legion2 Tks for finding and resolving this

PR #957, opened 2021-05-24
Added somfy repeat option

Legion2:
update somfy dependency to fix bug with multiple somfy remotes


1technophile:
Thanks for the PR

PR #961, opened 2021-05-27
Remove Secure Connection macro and add the config to wifimanager.

h2zero:
This adds the ability to specify if the MQTT broker connection is secure via a checkbox on the config page of WiFiManager.
Also added to the WiFiManager config page is a text box to enter the brokers TLS certificate.

In the case of using manual configuration a new macro is created `MQTT_SECURE_DEFAULT`; when defined as `true` a secure connection will be requested and the certificate defined in user_config.h will be used.

### DIFF
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -80,6 +80,10 @@ const byte ip[] = {192, 168, 1, 99};
 const byte mac[] = {0xDE, 0xED, 0xBA, 0xFE, 0x54, 0x95}; //W5100 ethernet shield mac adress
 #endif
 
+#ifndef NTP_SERVER
+#  define NTP_SERVER "pool.ntp.org"
+#endif
+
 #ifdef MQTT_HTTPS_FW_UPDATE
 #  if defined(ESP8266) || defined(ESP32)
 //If used, this should be set to the root CA certificate of the server hosting the firmware.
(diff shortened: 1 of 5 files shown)


h2zero:
Build fails on env:nodemcuv2-2g

platformio/packages/framework-arduinoespressif8266/tools/sdk/libc/xtensa-lx106-elf/lib/libc.a(lib_a-sscanf.o): in function `sscanf':
/home/earle/src/esp-quick-toolchain/repo/newlib/newlib/libc/stdio/sscanf.c:411: multiple definition of `sscanf'; .pio/build/nodemcuv2-2g/lib621/libA6lib.a(sscanf.cpp.o):sscanf.cpp:(.text.sscanf+0x4): first defined here

Cause is the A6lib sscanf.cpp is redefining newlib functions. Seems that the file could be removed from the lib as that was included before newlib was in the 8266 framework I believe.

1technophile:
> Cause is the A6lib sscanf.cpp is redefining newlib functions. Seems that the file could be removed from the lib as that was included before newlib was in the 8266 framework I believe.

I would suggest to fork it, modify and point to the fork temporarily.
You may submit a PR to the A6lib repo.

main/User_config.h
```diff
@@ -80,6 +80,10 @@ const byte ip[] = {192, 168, 1, 99};
 const byte mac[] = {0xDE, 0xED, 0xBA, 0xFE, 0x54, 0x95}; //W5100 ethernet shield mac adress
 #endif
 
+#ifndef NTP_SERVER
+#  define NTP_SERVER "pool.ntp.org"
+#endif
```
1technophile, line 85: Should we exclude Arduino UNO and Mega here?

h2zero: I didn't think it necessary since it only defines a constant. It can be moved inside a conditional check if you prefer.

h2zero:
> 
> 
> > Cause is the A6lib sscanf.cpp is redefining newlib functions. Seems that the file could be removed from the lib as that was included before newlib was in the 8266 framework I believe.
> 
> I would suggest to fork it, modify and point to the fork temporarily.
> You may submit a PR to the A6lib repo.

done

main/User_config.h
1technophile: ok good point

Issue #951, opened 2021-05-17
bme280 not working


pauldogg:
Before submitting a problem please check the troubleshooting section
https://docs.openmqttgateway.com/upload/troubleshoot.html

**Describe the bug**
Using a heltec esp32 v2  with lora and oled, i've added a bme280 and the serial reports:
N: Bosch BME280 Initialized - Result of .begin(): 0x 

**To Reproduce**
1. connect bme280 (i tried 3 different sensors)
2. flash module with omg with BME280 enabled (i tried address 0x77 and 0x76)
3. See error in serial console

**Expected behavior**
Address identified, values reported

**Screenshots**
N: Bosch BME280 Initialized - Result of .begin(): 0x

**Environment (please complete the following information):**
* OpenMQTTGateway version used V0.9.6
* Library version related to the problem you have (if you have troubles with RF provide the version of RCSwitch library) bme280.h

**Additional context**

NorthernMan54:
To assist in troubleshooting, can you share your wiring diagram

pauldogg:
Sure.
i'm using these: https://www.amazon.com/gp/product/B01N4EHIW6

mcu      bme
3.3v -->vcc
21----->SDA
22----->SCL

I also have a TSL2561 on the same bus and that one works fine.
Thanks

pauldogg:
Wrong sensor bmp instead of bme derp

spacemanspiff2007:
I'm having a similar issue

Connection
```
3.3v -> VIN
21   -> SDA
22   -> SCL
```

Output on start
```text
************* WELCOME TO OpenMQTTGateway **************
N: OpenMQTTGateway Version: version_tag
[E][WiFiMulti.cpp:184] run(): [WIFI] Connecting Failed.
N: WiFi ok with manual config credentials
N: Setup BME280 on adress:
N: Bosch BME280 Initialized - Result of .begin(): 0x
N: IR_EMITTER_GPIO: 14
N: IR_RECEIVER_GPIO: 26
N: OpenMQTTGateway modules: ["BME280","IR"]
N: ************** Setup OpenMQTTGateway end **************
```

What I don't understand is why the adress of the BME does not get printed on startup.
Since it's a constant it should always work.

This is what gets published:
```
N: Subject: /CLIMAtoMQTT/bme
N: Received json : {"tempc":0,"tempf":32,"hum":0,"pa":0,"altim":44330.77,"altift":145442.2}
```

[...]

PR #962, opened 2021-05-28
Fix BME280

### LINKED ISSUE #963: Sensor BME280 is not working
Sensor BME280 is not working (GY-BME280)

Connection
```
3.3v -> VIN
21   -> SDA
22   -> SCL
```
I tried two different sensors I bought 6 month apart
I've tried both sensor with adress 0x77 and 0x76.
I've tried both times with switched SCL/SDA pins

Output on start after the log fix:
```text
N: OpenMQTTGateway Version: version_tag
[E][WiFiMulti.cpp:184] run(): [WIFI] Connecting Failed.
N: WiFi ok with manual config credentials
N: Setup BME280 on adress: 0x76
N: Bosch BME280 Initialized - Result of .begin(): 0x0
N: IR_EMITTER_GPIO: 14
N: IR_RECEIVER_GPIO: 26
N: OpenMQTTGateway modules: ["BME280","IR"]
N: ************** Setup OpenMQTTGateway end **************


N: OpenMQTTGateway Version: version_tag
[E][WiFiMulti.cpp:184] run(): [WIFI] Connecting Failed.
N: WiFi ok with manual config credentials
N: Setup BME280 on adress: 0x77
N: Bosch BME280 Initialized - Result of .begin(): 0x0
N: IR_EMITTER_GPIO: 14
N: IR_RECEIVER_GPIO: 26
N: OpenMQTTGateway modules: ["BME280","IR"]
N: ************** Setup OpenMQTTGateway end **************

```

spacemanspiff2007:
Fix log output, since %H never prints anything

### DIFF
--- a/main/ZsensorBME280.ino
+++ b/main/ZsensorBME280.ino
@@ -49,7 +49,13 @@
 BME280 mySensor;
 
 void setupZsensorBME280() {
-  Wire.begin(BME280_PIN_SDA, BME280_PIN_SCL);
+
+  // Allow custom pins on ESP Platforms
+  #if defined(ESP8266) || defined(ESP32)
+    Wire.setPins(BME280_PIN_SDA, BME280_PIN_SCL);
+  #endif
+
+  Wire.begin();
 
   mySensor.settings.commInterface = I2C_MODE;
   mySensor.settings.I2CAddress = BME280_i2c_addr;
--- a/main/ZsensorBME280.ino
+++ b/main/ZsensorBME280.ino
@@ -50,7 +50,7 @@ BME280 mySensor;
 
 void setupZsensorBME280() {
   Wire.begin(BME280_PIN_SDA, BME280_PIN_SCL);
-  
+
   mySensor.settings.commInterface = I2C_MODE;
   mySensor.settings.I2CAddress = BME280_i2c_addr;
   Log.notice(F("Setup BME280 on adress: %X" CR), BME280_i2c_addr);
@@ -107,8 +107,7 @@ void setupZsensorBME280() {
   int ret = mySensor.begin();
   if (ret == 0x60) {
     Log.notice(F("Bosch BME280 successfully Initialized: %X" CR), ret);
-  }
-  else {
+  } else {
     Log.notice(F("Bosch BME280 failed: %X" CR), ret);
   }
 }
--- a/main/ZsensorBME280.ino
+++ b/main/ZsensorBME280.ino
@@ -50,12 +50,12 @@ BME280 mySensor;
 
 void setupZsensorBME280() {
 
-# if defined(ESP8266) || defined(ESP32)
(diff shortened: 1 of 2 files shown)


spacemanspiff2007:
#951 951

spacemanspiff2007:
#963

spacemanspiff2007:
Initialized BME280, otherwise it will not work

spacemanspiff2007:
I've updated the files so lint should pass now.
The other check seems to fail because of some unrelated reason but I am not sure.

h2zero:
Build error notes:
```
error: no matching function for call to 'TwoWire::begin(int&, int&)'
framework-arduino-avr\libraries\Wire\src/Wire.h:53:10: note: candidate: void TwoWire::begin()
framework-arduino-avr\libraries\Wire\src/Wire.h:54:10: note: candidate: void TwoWire::begin(uint8_t)
framework-arduino-avr\libraries\Wire\src/Wire.h:55:10: note: candidate: void TwoWire::begin(int)
```
AVR  wire library does not accept pins for parameters in the call to `begin()`

I recommend using a conditional check such as:
```
#Ifdef ESP32
  Wire.setPins(BME280_PIN_SDA, BME280_PIN_SCL);
#endif
```
Instead of the `Wire.begin` call that was added.

spacemanspiff2007:
@h2zero 
Thank you for your quick analysis! Do you think it makes sense to just disable this for the AVRs?
I think the esp8266 supports it, too.
What would be the def for that?

h2zero:
@spacemanspiff2007 
>Do you think it makes sense to just disable this for the AVRs?
I think the esp8266 supports it, too.
What would be the def for that?

I do not have the means to test the functionality but you could add the check as:
```
#if defined (ESP8266) || defined(ESP32)
  Wire.setPins(BME280_PIN_SDA, BME280_PIN_SCL);
#endif
```

That should at least compile and I don't see any reason for it to affect the function of all involved.

spacemanspiff2007:
I've done exactly that!
Thank you very much for you help!

h2zero:
No worries 😄 lets hope it works out.

h2zero:
@spacemanspiff2007, sorry I guess 8266 does not have `setPins()`.

You'll need to change the check to this:
```
#  if defined(ESP8266) || defined(ESP32)
  Wire.begin(BME280_PIN_SDA, BME280_PIN_SCL);
#  else
  Wire.begin();
#  endif
```

1technophile:
Thanks for the PR, for the code format you can take a look below:
`So as to format automaticaly your document you have to add the "clang-Format" extension to VSCode, once done, you can format the code by doing a right click into the code file window and clicking "Format document".`

spacemanspiff2007:
@1technophile 
Thanks for the hint - I'll do this the next time from the beginning.

PR #975, opened 2021-06-02
DT24 bluetooth Voltmeter

NorthernMan54:
Support for DT24 Bluetooth digital display DC Power APP Voltmeter
Ammeter Battery Capacity Tester Fuel Gauge voltage detector Meter

https://www.aliexpress.com/item/4000835324129.html?spm=a2g0o.productlist
.0.0.5aa96393c4qWIg&algo_pvid=dd8f4348-fad0-43e3-b3c0-e711180ed13e&algo_
expid=dd8f4348-fad0-43e3-b3c0-e711180ed13e-0&btsid=0bb062441622503466970
1271e3fb6&ws_ab_test=searchweb0_0,searchweb201602_,searchweb201603_

For reference, these were the compiler definitions I used

```
; *** ble flags ***
  '-DZgatewayBT="BT"'
;  '-DLED_RECEIVE=2'
;  '-DLED_RECEIVE_ON=0'
  '-DTimeBtwRead=0'
  '-DScan_duration=20000'
  '-DAttemptBLECOnnect=true'
  '-DPublishOnlySensors=true'
;  '-DLOG_LEVEL=LOG_LEVEL_TRACE'
```

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -716,6 +716,8 @@ void notifyCB(
   if (!ProcessLock) {
     Log.trace(F("Callback from %s characteristic" CR), pBLERemoteCharacteristic->getUUID().toString().c_str());
 
+    // This logic needs refactoring and should be changed to a device/model approach based on the sensor macAdr lookup.  Growing this beyond 2 devices will get too cumbersome
+
     if (length == 5) {
       Log.trace(F("Device identified creating BLE buffer" CR));
       JsonObject& BLEdata = getBTJsonObject();
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -550,6 +550,21 @@ void WS02Discovery(const char* mac, const char* sensorModel) {
   createDiscoveryFromList(mac, WS02sensor, WS02parametersCount, "WS02", "SensorBlue", sensorModel);
 }
 
+void DT24Discovery(const char* mac, const char* sensorModel) {
+#    define DT24parametersCount 5
+  Log.trace(F("DT24Discovery" CR));
+  const char* DT24sensor[DT24parametersCount][8] = {
+      {"sensor", "DT24-volt", mac, "power", jsonVolt, "", "", "V"},
+      {"sensor", "DT24-amp", mac, "power", jsonCurrent, "", "", "A"},
+      {"sensor", "DT24-watt", mac, "power", jsonPower, "", "", "W"},
+      {"sensor", "DT24-watt-hour", mac, "power", jsonEnergy, "", "", "kWh"},
+      {"sensor", "DT24-price", mac, "", jsonMsg, "", "", ""}
+      //component type,name,availability topic,device class,value template,payload on, payload off, unit of measurement
+  };
+
+  createDiscoveryFromList(mac, DT24sensor, DT24parametersCount, "DT24", "ATorch", sensorModel);
+}
+
 #  else
 void MiFloraDiscovery(const char* mac, const char* sensorModel) {}
 void VegTrugDiscovery(const char* mac, const char* sensorModel) {}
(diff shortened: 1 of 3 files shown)


1technophile:
Thanks, one last thing can you add also the model here please
https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/prerequisites/devices.md#for-ble-devices

main/ZgatewayBT.ino
```diff
@@ -722,6 +722,34 @@ void notifyCB(
...
+      JsonObject& BLEdata = getBTJsonObject();
+      String mac_adress = pBLERemoteCharacteristic->getRemoteService()->getClient()->getPeerAddress().toString().c_str();
+      mac_adress.toUpperCase();
+      for (vector<BLEdevice*>::iterator it = devices.begin(); it != devices.end(); ++it) {
+        BLEdevice* p = *it;
+        if ((strcmp(p->macAdr, (char*)mac_adress.c_str()) == 0)) {
+          if (p->sensorModel == LYWSD03MMC)
+            BLEdata.set("model", "LYWSD03MMC");
+          else if (p->sensorModel == MHO_C401)
+            BLEdata.set("model", "MHO_C401");
```
1technophile, line 758: Not sure you need to do these checks, as they are already done before

1technophile:
Could you add also the auto-discovery functions for this sensor please, here is an example:
https://github.com/1technophile/OpenMQTTGateway/pull/972

main/ZgatewayBT.ino
NorthernMan54: I was looking at the logic in the bluetooth notification callback and was thinking it needs a refactor to handle more than a few devices as the usage of message length to determine the message parser to be used is going to be problematic with more devices.  My thought was that it would make more sense to do the MAC address to sensor model mapping first, then parse based on sensor model.  Or should the logic be length and model ?  Am thinking that this should be done with the next addition.

That IF block is a duplicate of the one when the length is '5' that I augmented with DT24 for length '20'.  Its a little bit redundant but would correctly identify in the case those devices sent a message with length '20'.

1technophile: > My thought was that it would make more sense to do the MAC address to sensor model mapping first

Agreed, it can be done with another PR indeed!

> would correctly identify in the case those devices sent a message with length '20'.

I don't think they do, you can remove

1technophile:
@NorthernMan54 we are planning to do some modifications on the BLE parts, do you think we could go further with this PR or should I integrate it like this?

NorthernMan54:
I have the home assistant discovery bits mostly complete and am just wrapping up my testing.  I had been running things without discovery enabled, as my testing devices kept appearing in my production environment.  And once I enabled discovery and the gatewaybt module caused my home automation to crash.

I’m not running home assistant but homebridge with a plugin ( homebridge-tasmota) that leverages home assistant mqtt discovery messages to automatically integrate tasmota and other devices into Homebridge.  And once I enabled openmqttgateway discovery with Bluetooth devices everything crashed on homebridge.  And I’m adding some defensive code ( my homebridge code didn’t like the topic wild cards in the discovery message) to stabilize my setup and ensure device discovery is working.

I’m about 90% complete

NorthernMan54:
This is my WIP which is 90% complete, and likely just needs some minor tweaks before publishing.  I would recommend merging, and I will rebase myself for any further tweaks required.

main/ZgatewayBT.ino
NorthernMan54: Added in comment into the code base to reflect this recommendation

h2zero:
> I would recommend merging, and I will rebase myself for any further tweaks required.

Agreed, the upcoming BLE changes will change a lot of this. If this is working as is, then merging now and updating it against the new base would make the most sense.

1technophile:
Thanks

PR #991, opened 2021-06-18
DT24 - Update messages to support temp

NorthernMan54:
With the additional functionality added by @h2zero I’m able to extract
the temperate data from the 2nd part of the bluetooth message.


1technophile:
Thanks!

NorthernMan54:
Tks @1technophile

PR #988, opened 2021-06-14
Ble read/write characteristics over MQTT.

h2zero:
This allows reading and writing BLE characteristics from an MQTT message.

Example MQTT message format:
```
mosquitto_pub -t home/OpenMQTTGateway/commands/MQTTtoBT/config -m '{
  "ble_write_address":"AA:BB:CC:DD:EE:FF", 
  "ble_write_service":"cba20d00-224d-11e6-9fb8-0002a5d5c51b",
  "ble_write_char":"cba20002-224d-11e6-9fb8-0002a5d5c51b", 
  "ble_write_value":"TEST",
  "value_type":"STRING",
  "ttl":4 }'
```
```
mosquitto_pub -t home/OpenMQTTGateway/commands/MQTTtoBT/config -m '{
  "ble_read_address":"AA:BB:CC:DD:EE:FF", 
  "ble_read_service":"cba20d00-224d-11e6-9fb8-0002a5d5c51b",
  "ble_read_char":"cba20002-224d-11e6-9fb8-0002a5d5c51b", 
  "value_type":"STRING",
  "ttl": 2 }'
```
The `ttl` parameter is the number of attempts to connect (defaults to 1), which occur after the BLE scan completes.
`value_type` can be one of: STRING, HEX, INT, FLOAT. Default is STRING if omitted in the message.

A response is provided over MQTT in the format:
write : 
```
{
  "id":"AA:BB:CC:DD:EE:FF",
  "service":"cba20d00-224d-11e6-9fb8-0002a5d5c51b",
  "characteristic":"cba20002-224d-11e6-9fb8-0002a5d5c51b",
  "write":"TEST",
  "success":true
}
```
read :
```
{
  "id":"AA:BB:CC:DD:EE:FF",
  "service":"cba20d00-224d-11e6-9fb8-0002a5d5c51b",
  "characteristic":"cba20002-224d-11e6-9fb8-0002a5d5c51b",
  "read":"TEST",
  "success":true
}
```

PR #992, opened 2021-06-19
Add Miflora battery reading

### LINKED ISSUE #217: Battery/Firmware reading for Xiaomi flora & temperature/humidity sensor
Hi, I've been using the Openmqttgateway with Fhem for a few days now and it works fine. 

For Fhem there is a Perl module which excludes the readings of the Xiaomi BLE sensors. 
Among other things, it is also possible to read the battery and firmware version. 

gatttool -i $hci -b $mac --char-read -a 0x38";   # Firmware and Battery
Source: https://svn.fhem.de/trac/browser/trunk/fhem/FHEM/74_XiaomiFlowerSens.pm?rev=15916

It would be great if you could put that in too Openmqttgateway. 
So that maybe every 24 hours the battery values are read out.

1technophile:


1technophile:
Prerequisite #988

PR #993, opened 2021-06-20
Add switching between MQTT brokers with self signed client certificates.

h2zero:
WIP: Allows defining multiple MQTT broker SSL certificates and client certificates with the capability to switch between them.


1technophile:
Remaining, switching between certificates and doc

PR #996, opened 2021-06-24
Use correct SDA+SCL pins

qistoph:
ESP8266 uses D1 (5) and D2 (4), ESP32 uses 21 and 22. Both provide defines for SDA and SCL pin numbers though.

This PR fixes the pin numbers for SDA and SCL on ESP8266 and keeps them unchanged for ESP32.


1technophile:
Thanks!

PR #997, opened 2021-06-24
Allow inversion of input GPIO

qistoph:
This PR enables users of the input GPIO to switch the on/off values.

Rationale:
The ESP8266 uses internal pull-up for input GPIO. A common use case is to use a switch to pull the input to ground. So the value is `LOW` while the button is pressed and `HIGH` otherwise.

Swapping the values for `INPUT_GPIO_ON_VALUE` and `INPUT_GPIO_OFF_VALUE` introduced in this PR, makes a `LOW` button report as `On`.


1technophile:
thanks!

PR #1000, opened 2021-06-27
Add Meater BBQ sensor

1technophile:


bailout00:
`
Scan begin

D NimBLEScan: ">> start(duration=10)"

D NimBLEScan: "<< start()"

I NimBLEScan: "New advertiser: d0:d9:4f:81:41:21"

D NimBLEScan: "discovery complete; reason=0"

N: Device detected: D0:D9:4F:81:41:21

N: Filtered connectable device

N: Found 2 devices, scan number 18 end

N: BLE Connect begin

D NimBLEClient: ">> connect(d0:d9:4f:81:41:21)"

D NimBLEClient: "Got Client event "

I NimBLEClient: "Connected event"

**lld_pdu_get_tx_flush_nb HCI packet count mismatch (1, 2)**

**D NimBLEClient: "Got Client event "**

**E NimBLEClient: "Connection failed; status=574 "**

**E: Connect to: d0:d9:4f:81:41:21 failed**

N: Failed getting characteristic

D NimBLEClient: ">> deleteServices"

D NimBLEClient: "<< deleteServices"

N: BLE Connect end

`

This is what I'm getting currently, running dev code with your modified files, on an ESP32 DevKit C v4. Would love to help. Also found this: 
https://github.com/nathanfaber/meaterble

But it seems like you already have it broken down in the code.

1technophile:
I found also this error from time to time, do you have it at each connection ?  (note that your meater should not be connected to your smartphone or to a meater plus)

bailout00:
I just randomly started working now, I was messing with the antenna, maybe it's a connection issue? Seems to be working better now, this is what I'm seeing now:
This is with tip temp of ~80 and ambient of ~100
{"model":"MEATER","id":"D0:D9:4F:81:41:21","tempc":27.5625,"tempc2":2.7375,"tempc3":0.5}
Maybe a precision/decimal issue on the tempc2. Tempc3 never seems to change, not sure if that is a model-based reading, my MEATER is very, very old, one of the first and well worn.

I didn't get anything at all when it was connected to my phone, BT is off on that now

1technophile:
And if you connect your phone just after getting data can you compare what you get in the phone exactly and what you get in the broker.
So as to identify what values needs to be sum up.

h2zero:
For reference; in the log `E NimBLEClient: "Connection failed; status=574 "` means this:
0x023e | 0x3e | BLE_ERR_CONN_ESTABLISHMENT | Connection Failed to be Established
-- | -- | -- | --

So from that we can see that either the devices are too far apart or possibly the client is trying to connect with parameters that the peripheral doesn't work well with. If it's the former the troubleshooting/solution is obvious, get the devices closer together and test. If it's the latter it will be more involved in the gateway code to iteratively change the connection parameters to find the "sweet spot".

1technophile:
I will come back later on this

PR #1013, opened 2021-07-10
Add a macro to use the MAC address as the gateway name.

h2zero:
## Description:
Adds the macro USE_MAC_AS_GATEWAY_NAME, when defined the gateway name will default to: 
OMG_XXXXXXXXXXXX, where the X's are the mac address characters.
This can be changed in WiFiManager.

## Checklist:

### DIFF
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -49,6 +49,10 @@
  * to the access point with your password (SSID: WifiManager_ssid, password: WifiManager_password)
  */
 /*-------------DEFINE GATEWAY NAME BELOW IT CAN ALSO BE DEFINED IN platformio.ini----------------*/
+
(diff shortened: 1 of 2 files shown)


NorthernMan54:
@h2zero Tks very much for this, I had been doing it manually for a while now.

NorthernMan54:
@h2zero I just updated my devices with the latest beta, and the OTA device name is not the expected Mac address but 'OpenMQTTGateway', shouldn't this use the Mac address as the OTA Device name as well ?

h2zero:
@NorthernMan54 Thanks for pointing this out, it had not occurred to me to change this as well. I will submit a PR to rectify.

PR #1029, opened 2021-07-18
fix: add code for miscale_v2 lbs

rayslinky:
## Description:
Currently, if the scale is set to imperial units, the incorrect weight is reported (by half). This patch duplicates the code from process_scale_v1 to account for kg or lbs. I have tested the patch with both.

## Checklist:


1technophile:
Thanks!

PR #1032, opened 2021-07-21
Advertise discovery of the battery sensor for MiFlora devices

Kernald:
Follow-up on #992, #217

## Checklist:

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -279,9 +279,10 @@ void strupp(char* beg) {
 
 #  ifdef ZmqttDiscovery
 void MiFloraDiscovery(const char* mac, const char* sensorModel) {
-#    define MiFloraparametersCount 4
+#    define MiFloraparametersCount 5
   Log.trace(F("MiFloraDiscovery" CR));
   const char* MiFlorasensor[MiFloraparametersCount][8] = {
+      {"sensor", "MiFlora-batt", mac, "battery", jsonBatt, "", "", "%"},
       {"sensor", "MiFlora-lux", mac, "illuminance", jsonLux, "", "", "lx"},
       {"sensor", "MiFlora-temp", mac, "temperature", jsonTempc, "", "", "°C"},
       {"sensor", "MiFlora-fer", mac, "", jsonFer, "", "", "µS/cm"},


1technophile:
Thank you!

tomcat-xx:
Could someone please explain about it? Should I see "batt" for MiFlora sensors now? I thought sensors are not advertising it, so how can something be changed on gateway side?

Kernald:
@tomcat-xx #992 added some logic where the gateway actively connect to the MiFlora sensors on a regular basis (like the official app would do), which uses more power but gives the battery level and firmware version (the app also uses this to download the history off the sensor). #988 has more details on this active connection. While before this PR, the sensor was already created and data published (since #992), this PR adds the auto-discovery settings for things like Home Assistant to discover the sensor automatically.

tomcat-xx:
@Kernald Thanks for quick answer and explanation. Now I understand. I know about active connection as I'm using it once a day to get battery value (into fhem with other module) in addition to other, advertised values (via OpenMQTTGateway). Seems that I then can integrate it now. But have to read more about it to get to know what I have to publish to which topic to place a request.

tomcat-xx:
@Kernald Sorry, me gain. Because you wrote "some logic where the gateway actively connect to the MiFlora sensors on a _regular_ basis", I have to ask: Do I have to trigger it or am I even able to do it? Or does regular in this context mean "from time to time, automatically".

By the way or maybe related: Could it be caused by MiFlora vs. vegtrug, it looks as if there is different coding? I bought some of the sensors as MiFLora and some as vegtrug while all are publishing HHCCJCY01HHCC as model.

Kernald:
@tomcat-xx I'm not sure exactly, I'm not familiar with the Bluetooth stack. I think I remember reading something about a counter that can be changed in the build options, but I'm really not sure. I was just wondering why the battery sensor wasn't showing despite having been added previously by someone else and seeing traces of it on MQTT and figured this was missing.

1technophile:
The ble connection to elligible devices is done every 10 scans per default. It can be changed [here](https://github.com/1technophile/OpenMQTTGateway/blob/0ede075843c71c7438db86296abae443f6d1e4d4/main/config_BT.h#L70) For the mi flora a BLE connection will be initiated every 10 scans so as to retrieve battery level. If you scan every 55s (default) you will get the battery level every 550s.

tomcat-xx:
Thanks, but then I don't understand. Maybe I have to clarify: It is not my intention to change something. I upgraded to 9.8.0 on some ESP two weeks ago, all others over one week ago, just to keep it up-to-date. Only now I came across that new functionality. As I could not find the battery value somewhere I thought I actively have to do something.

1technophile:
Did you check if you get a new entity for miflora battery?
Also you can take a look to the broker by using mqtt explorer and let it connected during 1h. Once done you can check the history of messages and see if the battery is published.

tomcat-xx:
That's what I did (using mqtt explorer). Already last time and now again for over 2 hours.

Back to my question above: There is no difference between MiFlora and VegTrug? (as I now in addition to coding saw older pull request "Add VegTrug (Mi Flora like) sensor", which made me thinking it is not the same).

And I also have activated module ZgatewayBT only. But when looking into coding that should cover ZgatewayBLEConnect.

1technophile:
> There is no difference between MiFlora and VegTrug?

Yes there is, they don't have the same[ identification criteria](https://github.com/1technophile/OpenMQTTGateway/blob/c5f2c1ca6f48d7468ff7e366e780917210a80918/main/ZgatewayBT.ino#L1136), currently, we don't connect to Vegtrug to retrieve battery. If you want we can do a short test to see if we can retrieve battery from these.

tomcat-xx:
Now I double checked bills and packages. I bought different sensors as MiFlora, VegTrug and also Royal Gardineer. But as written above all are sending HHCCJCY01HHCC as model. Is something else used for identification? What's about that "service_data"? Can I see it somewere (did not in serial monitor of Arduino IDE)?

Additional info: All are working with the Android app "Flower care" and in fhem with module 74_XiaomiBTLESens.pm which I want to replace by MQTT by using OpenMQTTGateway.

1technophile:
Could you share the configuration used (config files for arduino ide, env for platformio or binary name)  and version?

tomcat-xx:
First let me say, that this is my first "ESP32 project". With (new installed) Platformio I could not get it running, so I used Ardunino IDE (first 1.8.13,  now 1.8.15). Started with v0.9.5 on 6 devices, upgraded to v0.9.6 and now to v0.9.8. Downloaded OpenMQTTGateway_sources.zip and esp32dev-ble-libraries.zip, edited User_config.h. Flashed first time with devices attached to USB, now some also and some via OTA.

[preferences.txt](https://github.com/1technophile/OpenMQTTGateway/files/7046558/preferences.txt)

1technophile:
You don't need to mess with a dev environment anymore now, we have added an easier method so as to upload from chrome directly:
https://docs.openmqttgateway.com/upload/web-install.html

You can choose esp32dev-ble for your use cases.

This will use a standard configuration.

tomcat-xx:
Did it for a test, but it is not my preferred way. Now I have the dev environment and with web upload I have to re-enter the settings every time. Also some of the ESPs are "integrated" now and I first have to pull them out. Anyway it was not the solution for the issue.

After I saw the "erase flash" option I tried same with IDE. But there is no such option. There are screen shots of it in the web but maybe from older version or other modules. Could be that there is no need for it and it is done every time, but I used the espressif flash download tool for it. Both ways without result.

Then I played around a bit and found the solution. I'm using white lists (as I had issues with stable operations in v0.9.5) and after removing the white list it worked. In the received batt messages the "id" is not at the beginning it starts with "model". Could that be the root cause?

1technophile:
> Did it for a test, but it is not my preferred way. Now I have the dev environment and with web upload I have to re-enter the settings every time. Also some of the ESPs are "integrated" now and I first have to pull them out. Anyway it was not the solution for the issue.

Which settings are you referring to ?

> Then I played around a bit and found the solution. I'm using white lists (as I had issues with stable operations in v0.9.5) and after removing the white list it worked. In the received batt messages the "id" is not at the beginning it starts with "model". Could that be the root cause?

That's a very good point we may have an issue here, @h2zero maybe we don't publish the BLE connection messages when the white list is activated due to this missing id?

tomcat-xx:
> Which settings are you referring to ?

Wifi and MQTT settings, which are gone when you do a flash erase. That's a difference I saw between using IDE and web upload.  But as we now know that there is no issue with something not erased it is not relevant anymore.

1technophile:
> Wifi and MQTT settings, which are gone when you do a flash erase. That's a difference I saw between using IDE and web upload. But as we now know that there is no issue with something not erased it is not relevant anymore.

If you untick the box "Erase flash" on web flasher they should be kept

h2zero:
>That's a very good point we may have an issue here, @h2zero maybe we don't publish the BLE connection messages when the white list is activated due to this missing id?

Hmm, I'll have a look.

h2zero:
@1technophile, I had a quick look at this and I do not see any issue regarding the white list and connections. Only if the device is not on the white list, when used, could I see it not connecting/publishing the data.  

>In the received batt messages the "id" is not at the beginning it starts with "model". Could that be the root cause?

The "id" is there but it has been placed as the second parameter in the message as seen here:
https://github.com/1technophile/OpenMQTTGateway/blob/149712b57356fd90f0bb2f11e5b8fc5bda618529/main/ZgatewayBLEConnect.ino#L277

For consistency I suggest those 2 lines be swapped. Otherwise I see no issues so far. Later today I will look to emulating this device and try to reproduce this.

tomcat-xx:
Really strange. I tried to reproduce it for further tests but even with set white list I'm getting batt messages now. Could I have "initiated" or "triggered" something else with removing the white list and restart of the OMGs? There are 3 of my ESPs near to the 8 flower sensors. Over weeks I did not get batt value, only the others that I got before the upgrade to v9.8.0. After discussions here I re-flashed 2 of them, one with web update and erase flag set, other with IDE and erased before - without success. After removing the white list (and maybe something else happened at the same time) all 3 are working.

Will bring one of my "untouched" ESPs now nearer to the flower sensors to see what happens.

tomcat-xx:
> Will bring one of my "untouched" ESPs now nearer to the flower sensors to see what happens.

Not any batt message on that "new" one after a few hours. Do I need to know more about "deduplication" to understand and for further analysis? How does it work related to opening connection to flower sensors for batt request? Shouldn't I see batt messages on all OMGs?

h2zero:
If you disable the white list on the "new" device does it get battery info?

tomcat-xx:
That's what I planned to do, but because of deduplication I was not sure how long I have to wait before and just did it today.

Result: Let me say it that way: I could not see a batt message before and afterwards there were some. Could still be some kind of coincidence. 

But also after re-setting white list to the other gateways there are batt messages. So it is not directly caused by a set white list but somehow influenced by removing the old (set before upgrade to v0.9.8) white list and restart the gateway. Could also be caused by an update of the broker in between. In MQTT explorer I could not see any difference between old and new white list.

tomcat-xx:
Actually I wanted to write that it is ok for me now, but if someone likes I have still 2 gateways were I did change nothing so far.

But then I realized that I did not get battery values for a week. Tried different things, re-set the white list, stopped all gateways in reach except one, ... without success. Now I again removed white list on one gateway, restarted it and after a minute there were a few battery values. So don't know if my above statement is wrong (also battery values after re-setting the white list) or it just takes some time. Anyway I'm sure now it has to do with the white list.

1technophile:
I reproduce your issue with the following conditions:
* setting up a whitelist with a miflora on it (retain flag)
* the battery is still advertized
* restart the esp32
* the battery is not anymore advertized

1technophile:
maybe related to #1033

tomcat-xx:
> maybe related to #1033

Don't know. Don't have issues with my LYWSD03MMC but I'm not using original firmware. And battery value is advertised as all other values (different as for MiFlora).

tomcat-xx:
> I reproduce your issue with the following conditions:
> ...

Good to hear that you have a clear picture, I did not have that. Would have said that there were also messages after re-setting the white list. Also when looking into logs now: Last battery values are between 4 and 5 am on the day they stopped. It was a Saturday,  I'm sure I was in bed that time and did not restart the gateways ;-)

h2zero:
@tomcat-xx Could you try applying the patch posted in #1033. I believe it will fix your issue as well.

tomcat-xx:
Yes, looks good! Changed, restarted, new battery values within a minute, also from gateways with white list. Re-set white list on the one without, restarted, still battery values.

h2zero:
Glad to hear it!

PR #1037, opened 2021-07-26
Suggestion to also allow BLE WRITE procedures for WriteWithoutResponse characteristics

DigiH:
While playing with the new BLE READ/WRITE feature for my Playbulb Candles I found that writing was not possible for characteristics which are WriteWithoutResponse only.

I implemented this quick fix which might be useful for others with devices for which they want to write to WriteWithoutResponse characteristics.

Alternatively a more slective option could be possible for a future release, i.e. adding an additional WRITE tag like "ble_write_withoutresponse": true/false

## Checklist:


h2zero (approved):
LGTM

PR #1043, opened 2021-07-29
Add "server_cert" parameter to firmware_update command.

h2zero:
When updating from the MQTT command from a SSL enabled server this adds an option to provide
the server SSL certificate with the command.

This is useful if connecting to different servers or if the configured certificate has expired
and requires updating to not lose the ability to update via OTA.

The certificate provided will be saved to flash memory for use on subsequent restarts.

## Checklist:


1technophile:
Thanks

Issue #763, opened 2020-09-10
Any chance for OTA update possibility for the AVATTO S06 IR gateways?


DigiH:
Having seen the 0.9.5-beta release I thought I'd test it with my ESP8266 RF gateway, ESP32 BLE gateway and my IR gateway, for which I very much enjoy the AVATTO S06, after reading your blog entry.

For both the RF and BLE gateways the OTA updates worked flawlessly, but with the AVATTO S06 it failed, due to the lack of enough space.

Current from SYStoMQTT:
"freemem":35728

OTA upload error:
13:15:57 [INFO]: Upload size: 492608
13:15:57 [INFO]: Sending invitation to: 192.168.0.50
Authenticating...FAIL
13:16:00 [ERROR]: ERR: ERROR[4]: Not Enough Space

Since the Avatto would benefit the most from OTA updates - due to soldering required in contrast to the ESP8266/ESP32 - is there any leeway in reducing the Avatto specific build size so that it, and any future updates might fit into the 1mb flash for OTA updating? 

Probably just wishful thinking on my side, but I tought it might be worth asking.

Thanks

1technophile:
That's a good point, maybe playing with the partition sizes will enable to make the OTA works.

DigiH:
Where as previously with the default partition sizes I got

RAM:   [====      ]  38.3% (used 31392 bytes from 81920 bytes)
Flash: [=====     ]  49.5% (used 376980 bytes from 761840 bytes)

still failing with
17:01:47 [INFO]: Upload size: 381136
17:01:47 [INFO]: Sending invitation to: 192.168.0.50
Authenticating...FAIL
17:01:47 [ERROR]: ERR: ERROR[4]: Not Enough Space

now with adding the build_flag

-Wl,-Teagle.flash.1m64.ld
or 
board_build.ldscript = eagle.flash.1m64.ld

it's

RAM:   [====      ]  38.3% (used 31392 bytes from 81920 bytes)
Flash: [====      ]  39.3% (used 376980 bytes from 958448 bytes)

with a much better available total Flash size, but still failing

17:31:04 [INFO]: Upload size: 381136
17:31:04 [INFO]: Sending invitation to: 192.168.0.50
Authenticating...FAIL
17:31:04 [ERROR]: ERR: ERROR[4]: Not Enough Space

[...]

PR #1046, opened 2021-08-01
Disable MQTTsetMQTT and MQTT_HTTPS_FW_UPDATE to allow OTA for Avatto

DigiH:
Trying to update my Avattos with the currently released 0.9.8 I found that OTA wasn't possible any longer as the build was now too large

0.9.8
Checking size .pio/build/avatto-bakeey-ir/firmware.elf
RAM: [==== ] 42.4% (used 34772 bytes from 81920 bytes)
Flash: [====== ] 55.2% (used 528716 bytes from 958448 bytes)

this goes back to 0.9.7, where WiFi, MQTT and firmware update was introduced over MQTT commands

0.9.7
Checking size .pio/build/avatto-bakeey-ir/firmware.elf
RAM: [==== ] 42.4% (used 34772 bytes from 81920 bytes)
Flash: [====== ] 55.2% (used 528716 bytes from 958448 bytes)

Going back to building 0.9.6 I was able to OTA update again with the much smaller build

0.9.6
Checking size .pio/build/avatto-bakeey-ir/firmware.elf
RAM: [==== ] 39.7% (used 32512 bytes from 81920 bytes)
Flash: [==== ] 41.4% (used 396804 bytes from 958448 bytes)

Building 0.9.8 with the flags brings the build back to OTAable size

0.9.8 with
'-UMQTTsetMQTT'
'-UMQTT_HTTPS_FW_UPDATE'
Checking size .pio/build/avatto-bakeey-ir/firmware.elf
RAM: [==== ] 39.9% (used 32684 bytes from 81920 bytes)
Flash: [==== ] 42.3% (used 405024 bytes from 958448 bytes)

Similar to the flag setting in [env:rfbridge]

I'm not even sure if MQTT_HTTPS_FW_UPDATE would even work on the Avatto, with the limited flash memory.

Should we also add

board_build.ldscript = eagle.flash.1m64.ld

as a default, as it's required for OTA updates on the Avattos as discussed in 1technophile#763 (comment) ?

### DIFF
--- a/docs/use/gateway.md
+++ b/docs/use/gateway.md
@@ -135,6 +135,11 @@ vEsXCS+0yx5DaMkHJ8HSXPfqIbloEpw8nL+e/IBcm2PN7EeqJSdnoDfzAIJ9VNep
 
 A bash script is available [here](ota_command_cert.zip) to simplify the use of the `server_cert` parameter.  
 
+::: warning
+The pre-built binaries for **rfbridge** and **avatto-bakeey-ir** have the above WiFi and MQTT broker credentials and the Firmware update via MQTT options disabled. This is due to the restricted available flash, so as to still be able to use OTA firmware updates for these boards.
+:::
+
+
 # State LED usage
 
 The gateway can support up to 3 LED to display its operating state:
--- a/platformio.ini
+++ b/platformio.ini
@@ -756,6 +756,7 @@ board_build.flash_mode = dout
 [env:avatto-bakeey-ir]
 platform = ${com.esp8266_platform}
 board = esp01_1m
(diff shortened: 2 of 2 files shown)


1technophile:
Thanks for pointing this

1technophile:
> as a default, as it's required for OTA updates on the Avattos as discussed in #763 (comment) ?

I think so, could you add into the [docs ](https://docs.openmqttgateway.com/use/gateway.html#switching-brokers-and-using-self-signed-and-client-certificates) that the 2 boards doesn't have these functions on the pre built binaries please

DigiH:
Added a warning in the docs.

1technophile:
Thanks

PR #1062, opened 2021-08-16
Support for BME280 Temperature Correction

NorthernMan54:
## Description:

The Spark Fun BME 280 library has support for Temperature correction, and this pull request implements the ability to apply a correction via a compiler directive ( BME280Correction ).

 Correction is in celcius of temperature reported by bme280 sensor.  Both Celcius and Farenheit tempaeratures are adjusted.
 Value is a float
 ie Compiler Directive '-DBME280Correction=-3.4'

## Checklist:

### DIFF
--- a/main/ZsensorBME280.ino
+++ b/main/ZsensorBME280.ino
@@ -107,13 +107,11 @@ void setupZsensorBME280() {
   //  1 through 5, oversampling *1, *2, *4, *8, *16 respectively
   mySensor.settings.humidOverSample = 1;
 
-#  ifdef BME280Correction
-  // tempCorrection - Correction in celcius of temperature reported by bme280 sensor.  Both Celcius and Farenheit tempaeratures are adjusted.
+  // tempCorrection - Correction in celcius of temperature reported by bme280 sensor.  Both Celcius and Farenheit temperatures are adjusted.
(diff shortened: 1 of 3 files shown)


NorthernMan54:
I found that with my esp32's, cc1101 and bme280 all in the same case the temperature reading was always high.

1technophile:
Nice, could you add this macro into the .h config file for the BME280 please, so that the users know that this is available? Also, you could activate it into some *-test environments to verify that the build works with it.

NorthernMan54:
I changed the macro logic based on your recommendation and define it in the BME280 config file.  And have added `BME280Correction` to the esp32dev-all-test environment.

PS Any tips on flattening the commits?

h2zero:
@NorthernMan54 git rebase -i then mark the extra commits with `s` or squash. Then force push to the branch. However, it can be done by the github interface anyway so there isn't really a need unless you want to.

1technophile:
Thanks. Don't worry I will squash them.
For your knowledge you can also take the first commit sha and rebase from it
`git rebase -i <sha>`, a text editor will open, put an s in front of the commit you want to squash, save
and after do a `git push -f`

NorthernMan54:
Tks

PR #1090, opened 2021-10-14
Add support for discovery device trigger

Odyno:
## In short

- Add support for [MQTT Device Trigger](https://www.home-assistant.io/integrations/device_trigger.mqtt/). 
- Add Example on Documentation
- Used on RF Discovery

## More
The MQTT Device Trigger platform uses an MQTT message payload to generate device trigger events in to home assistant, so it's not create a flag/state/etc on the device (like what produced by binary sensor) but just declare where the event will be raise and provide information about the event.

I used this feature in the auto discovery process for the RF module, so we can reactivate what has been deactivated here this PR https://github.com/1technophile/OpenMQTTGateway/pull/1026 

It will also solve the problem https://github.com/1technophile/OpenMQTTGateway/issues/939 because the Device Trigger already has all the reset and event management logic, which by definition has no state.

## Checklist:

### DIFF
--- a/docs/integrate/home_assistant.md
+++ b/docs/integrate/home_assistant.md
@@ -36,26 +36,11 @@ OMG will use the auto discovery functionality of home assistant to create gatewa
 
 ## MQTT Device Trigger and RF
 
-With OpenMqttGateway [configured to receive RF signals](./setitup/rf.html) the messages are transmitted as indicated by [RCSwitch based gateway](./use/rf.html#rcswitch-based-gateway), so it is possible to receive a pulse every time the sensor discover a movement. 
+With OpenMqttGateway [configured to receive RF signals](./setitup/rf.html) the messages are transmitted as indicated by [RCSwitch based gateway](./use/rf.html#rcswitch-based-gateway), so it is possible to receive a pulse every time the sensor discover a signal. 
 
 With autodiscovery enabled, HomeAssistant will discover a [MQTT Device Trigger](https://www.home-assistant.io/integrations/device_trigger.mqtt/) identified by the value field given in the mqtt argument. 
 
-### Example with "DIGOO DG-HOSA 433MHz PIR Detector"
 
-With this configuration we receive from the broker only the close signal, which produces the change of state of the switch to "on". 
-
-```yaml
-binary_sensor:
-  - platform: mqtt
-    unique_id: pir.15961350
-    name: "Pir.Bagno"
-    device_class: motion
-    state_topic: "home/+/433toMQTT/15961350"
-    value_template: "{{ value_json.value }}"
-    payload_on: "15961350"
-    off_delay: 30
-
-```
 
 
 ## Manual integration examples
--- a/docs/integrate/home_assistant.md
+++ b/docs/integrate/home_assistant.md
@@ -1,8 +1,19 @@
 # Integrate Home Assistant
-## Auto discovery
-Home Assistant discovery is enabled by default on all binaries and platformio configurations except for UNO. With Arduino IDE please read the [advanced configuration section](../upload/advanced-configuration#auto-discovery) of the documentation.
 
-First enable discovery on your MQTT integration in HASS.
+Home Assistant provide the [MQTT integration](https://www.home-assistant.io/integrations/mqtt/) and through this integration it is possible to exploit and manage the messages published by OpenMqttGateway.
+
+Once this integration on home assistant is configured with the same mqtt broker, it is possible to create devices manually or through the autodiscovery function.
+
+
+## Auto discovery
+
+From Home Assistant site 
+
+> The discovery of MQTT devices will enable one to use MQTT devices with only minimal configuration effort on the side of Home Assistant. The configuration is done on the device itself and the topic used by the device.
+
+On OpenMqttGateway the Home Assistant discovery is enabled by default on all binaries and platformio configurations except for UNO. With Arduino IDE please read the [advanced configuration section](../upload/advanced-configuration#auto-discovery) of the documentation. Here are a few tips for activating discovery on Home Assistant, but for detailed configuration please refer to the Home Assistant website. 
+
+Enable discovery on your MQTT integration in HASS.
 
 ![](../img/OpenMQTTGateway-Configuration-Home-Assistant-Discovery-Integration.png)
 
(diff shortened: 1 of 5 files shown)


1technophile:
Thanks for the PR,
> I used this feature in the auto-discovery process for the RF module, so we can reactivate what has been deactivated here this PR #1026

The auto-discovery process of RF has been deactivated because it was creating a lot of unnecessary entities into Home Assistant, in particular when there is a lot of RF devices in the neighborhood. Does this new method prevent this?

Odyno:
If I understand what you mean, yes I confirm that no entities appear, this is my configuration: I have 3 gateways made with the [az-delivery esp32 board](https://www.az-delivery.de/en/products/esp32-developmentboard) each one with SRX882 and STX882. In my house there are about 30 RF devices and a lot of Bluetooth devices... that's my current  board on HA:
Screenshot 2021-10-17 at 15 15 56

To date I have intercepted over 350 RF signals and have none of these entities

1technophile:
To explain better, let's suppose that a device that doesn't belong to you emit a set of RF signals, 1145896, 1458765, 3555489.
This is the weather station of your neighbor. Do you think this is going to create one entity for each signal?

Odyno:
Confirmed, no entity for each signal. In HA it is part of the triggers, more specifically the [device trigger](https://www.home-assistant.io/docs/automation/trigger#device-triggers)

docs/integrate/home_assistant.md
```diff
@@ -22,6 +33,31 @@ OMG will use the auto discovery functionality of home assistant to create gatewa
...
+
+With this configuration we receive from the broker only the close signal, which produces the change of state of the switch to "on". 
+
+```yaml
+binary_sensor:
+  - platform: mqtt
+    unique_id: pir.15961350
+    name: "Pir.Bagno"
+    device_class: motion
+    state_topic: "home/+/433toMQTT/15961350"
```
1technophile: I'm not sure we should have the value at the topic end, as this is not the default parameter,
shouldn't it be:
```
state_topic: "home/+/433toMQTT"
```

main/ZmqttDiscovery.ino
```diff
@@ -90,7 +90,119 @@ void createDiscoveryFromList(const char* mac,
...
+  /* Set The topic */
+  if (topic[0]) {
+    char state_topic[mqtt_topic_max_size];
+
+    if (use_gateway_info) {
+      strcpy(state_topic, mqtt_topic);
+      strcat(state_topic, gateway_name);
+    } else {
+      strcpy(state_topic, "+/+");
+    }
```
1technophile: Should we not always use the wildcard so that the triggers are not gateway dependents?
Especially for multiple gateways setups.

Odyno: I just used the same policy already present on the code... here the same code on createDiscovery function
https://github.com/1technophile/OpenMQTTGateway/blob/e14a78e9c002def7780d07dfb409899fe799fffc/main/ZmqttDiscovery.ino#L144 

I was conservative and went with the same behaviour... anyway I agree with you, maybe a review is needed, what do you propose in case the gateway info is false? I ask this because in the code when posting messages those prefix  https://github.com/1technophile/OpenMQTTGateway/blob/e14a78e9c002def7780d07dfb409899fe799fffc/main/main.ino#L290 and https://github.com/1technophile/OpenMQTTGateway/blob/e14a78e9c002def7780d07dfb409899fe799fffc/main/main.ino#L299  and https://github.com/1technophile/OpenMQTTGateway/blob/e14a78e9c002def7780d07dfb409899fe799fffc/main/main.ino#L355   is always present and no matter what

do you think that "if" is unnecessary?

docs/integrate/home_assistant.md
Odyno: correct! in my code it is active by default, but it is not the default for all :)

Do you prefer that I remove the example or that I indicate the parameter as a precondition of the example?

main/ZmqttDiscovery.ino
1technophile: We are using the wildcards when the entity is gateway independent, which is the case with RF devices. The non-discovery messages can still be posted with the gateway_name. For the discovery message in your case I would simply set `use_gateway_infos `to `false` when calling the function.
You could also keep the same parameter name as the other function `gateway_entity` to be consistent

docs/integrate/home_assistant.md
1technophile: Would an example without valueAsAsubject activated works? If yes you could add this example type.

main/ZmqttDiscovery.ino
Odyno: OK, let's summarise to see if I've understood correctly, have the patience to answer me:

> We are using the wildcards when the entity is gateway independent, which is the case with RF devices. The non-discovery messages can still be posted with the gateway_name. 

So the action is to remove at all the else lines in order to always notify in the discovery message that the topic is always a child of the gateway_name.  **OK**

> For the discovery message in your case I would simply set use_gateway_infos to false when calling the function.

let me just notify you that in the previous implementation it was set to true https://github.com/1technophile/OpenMQTTGateway/blob/f84a8784000f4f201842e419aa2cbb13908e2dbd/main/ZgatewayRF.ino#L53 but I agree with you it's a bit strange... so **OK** fine with me to use false!

> You could also keep the same parameter name as the other function gateway_entity to be consistent

are we sure about this? this variable is very sneaky, according to my experience  I spent more than 20 minutes to understand what this parameter meant, which fields it made useless in the function signature...  if I may say so, I would suggest that we do not propagate this way of doing, but you/we can plan a small refactory so that these two functions become more expressive, perhaps transforming them into 4 (2 with gateway info as defaults and 2 where all the info can be specified)... and reusing the code they have in common with private sub-functions... etc
I didn't want to do all this work that would take the attention away from the focus of the PR, *so if you agree* I am willing to do this refactory in another PR.

docs/integrate/home_assistant.md
Odyno: I have no way of testing the example without destroying my home alarm :) I prefer not to distribute an example that I have not tested.

main/ZmqttDiscovery.ino
1technophile: > let me just notify you that in the previous implementation it was set to true

It should have been set to false.

> I didn't want to do all this work that would take the attention away from the focus of the PR, so if you agree I am willing to do this refactory in another PR.

Ok for another PR on this one

docs/integrate/home_assistant.md
```diff
@@ -36,26 +36,11 @@ OMG will use the auto discovery functionality of home assistant to create gatewa
 
 ## MQTT Device Trigger and RF
 
-With OpenMqttGateway [configured to receive RF signals](./setitup/rf.html) the messages are transmitted as indicated by [RCSwitch based gateway](./use/rf.html#rcswitch-based-gateway), so it is possible to receive a pulse every time the sensor discover a movement. 
+With OpenMqttGateway [configured to receive RF signals](./setitup/rf.html) the messages are transmitted as indicated by [RCSwitch based gateway](./use/rf.html#rcswitch-based-gateway), so it is possible to receive a pulse every time the sensor discover a signal. 
 
 With autodiscovery enabled, HomeAssistant will discover a [MQTT Device Trigger](https://www.home-assistant.io/integrations/device_trigger.mqtt/) identified by the value field given in the mqtt argument. 
 
-### Example with "DIGOO DG-HOSA 433MHz PIR Detector"
```
1technophile: Do you think we can still use the device trigger without `ValueAsaSubject`?

Odyno: I think so, it simply becomes more complicated because we have to introduce a concept of filtering the event by value... An attempt should be made to be sure of the result... but it is feasible.

1technophile: Could you add an example of this, so that people taking the pre-built versions can leverage this function?

Odyno: mmh... ok... as I told you before I put an example of real trigger made by my pir that I use as home burglar alarm. If you don't like this example because I use the flag, then for a different example you have to wait until I find the time to disassemble my home assistant, create the example that will have +0.5 added value compared to what is there now and then send it on the pr. 
you are the code manager, if you want to block for that... ok... but i just need to find the time to do it.

1technophile: The goal is more to have a consistent PR with docs rather than blocking it. I have no problem merging it like this if you think users will understand easily how to leverage this new function.

Odyno:
🥳  Time to review the code and I provider you the pr as promised!

Issue #931, opened 2021-04-25
Wrong log message for received json


Legion2:
**Describe the bug**
The `logJson` function is used by `receivingMQTT` and `pub`, to log the json which is send and received.
However, the log message text is always `Received json : %s`.
This means also when sending json, in the logs it shows as received json.

**To Reproduce**
1. Open the serial monitor
2. Wait for `SYStoMQTT`
3. See wrong log message

**Expected behavior**
The log message should either indicate the correct action (sending, receiving) or exclude this information from the log.

**Screenshots**
```
N: Subject: /SYStoMQTT
N: Received json : {"uptime":1803,"version":"v0.9.6","freemem":206120,...}
T: jsonPublishing
```

**Environment:**
* OpenMQTTGateway version: development

**Additional context**
I use log level `LOG_LEVEL_VERBOSE`.

PR #1093, opened 2021-10-27
rationalisation of logs and removal of duplicate codes (it should fix also #931) 

Odyno:
## Description:
Rationalisation of logs with removal of duplicate codes (it should fix also #931) 

## Checklist:

### DIFF
--- a/main/ZsensorDS1820.ino
+++ b/main/ZsensorDS1820.ino
@@ -102,7 +102,6 @@ void pubOneWire_HADiscovery() {
 #  ifdef ZmqttDiscovery
   // If zmqtt discovery is enabled, create a sensor topic for each DS18b20 sensor found on the bus, using addr as uniqueID
   if (disc) {
-    //Log.notice(F("CreateDiscoverySensor - Found %d" CR), ds1820_count);
     for (int index = 0; index < ds1820_count; index++) {
       createDiscovery("sensor",
                       (char*)(String(OW_TOPIC) + "/" + ds1820_addr[index]).c_str(),
--- a/main/ZsensorDS1820.ino
+++ b/main/ZsensorDS1820.ino
@@ -102,7 +102,7 @@ void pubOneWire_HADiscovery() {
 #  ifdef ZmqttDiscovery
   // If zmqtt discovery is enabled, create a sensor topic for each DS18b20 sensor found on the bus, using addr as uniqueID
   if (disc) {
-    Log.notice(F("CreateDiscoverySensor - Found %d" CR), ds1820_count);
+    //Log.notice(F("CreateDiscoverySensor - Found %d" CR), ds1820_count);
     for (int index = 0; index < ds1820_count; index++) {
       createDiscovery("sensor",
                       (char*)(String(OW_TOPIC) + "/" + ds1820_addr[index]).c_str(),
--- a/main/config_mqttDiscovery.h
+++ b/main/config_mqttDiscovery.h
@@ -94,7 +94,7 @@ void announceDeviceTrigger(bool use_gateway_info,
 #define DEVICEMANUFACTURER "OMG_community"
 
 /*-------------- Auto discovery macros-----------------*/
-// Set the line below to true so as to have autodiscovery working with OpenHAB
+// Set the line below to true so as to have auto working with OpenHAB
 #define OpenHABDiscovery false
 
 #if OpenHABDiscovery // OpenHAB autodiscovery value key definition (is defined command is not supported by OpenHAB)
--- a/main/config_mqttDiscovery.h
+++ b/main/config_mqttDiscovery.h
@@ -29,8 +29,31 @@
 
 extern String getUniqueId(String name, String sufix);
 extern void pubMqttDiscovery();
-extern void createDiscoveryFromList(const char* mac, const char* sensorList[][8], int sensorCount,
-                                    const char* device_name, const char* device_manufacturer, const char* device_model);
+
+/**
+ * Create a discover messages form a list of attribute
+ * 
+ * @param mac the mac adres
+ * @param sensorList[][0] = component type
+ * @param sensorList[][1] = name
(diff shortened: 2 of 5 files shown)


main/main.ino
```diff
@@ -286,15 +286,123 @@ bool to_bool(String const& s) { // thanks Chris Jester-Young from stackoverflow
...
+ * @return char* the string
+ */
+const char* to_char(JsonObject& jsondata) {
+  String out = "";
+  serializeJson(jsondata, out);
+  return out.c_str();
+}
+
+/**
+ * @brief from unsigned long to chart *
```
1technophile: small typo chart -> char

Odyno: Hi! I have discovered that this code is not thread-safe and contain a variable scope problem, I am updating the pr just now

1technophile:
One next step could be to rework the pubMQTT calls leveraging the toChar functions added.

Odyno:
Unfortunately the problem is in the to_char because: if I put the prefix "static char xxx[yy]" they are not thread safe, but if I don't put it they are subject to the stack clean when the function exits.

Odyno:
@1technophile  I've updated the code but had to put it back inline. I agree with you to make a review of the 30 or so versions of pubMQTT, maybe it would be useful to create an ad hoc class (maybe an extension of "client") that also manages the connection.

main/config_mqttDiscovery.h
```diff
@@ -94,7 +94,7 @@ void announceDeviceTrigger(bool use_gateway_info,
 #define DEVICEMANUFACTURER "OMG_community"
 
 /*-------------- Auto discovery macros-----------------*/
-// Set the line below to true so as to have autodiscovery working with OpenHAB
+// Set the line below to true so as to have auto working with OpenHAB
```
1technophile: Not sure this removal is necessary?

main/ZsensorDS1820.ino
```diff
@@ -102,7 +102,7 @@ void pubOneWire_HADiscovery() {
 #  ifdef ZmqttDiscovery
   // If zmqtt discovery is enabled, create a sensor topic for each DS18b20 sensor found on the bus, using addr as uniqueID
   if (disc) {
-    Log.notice(F("CreateDiscoverySensor - Found %d" CR), ds1820_count);
+    //Log.notice(F("CreateDiscoverySensor - Found %d" CR), ds1820_count);
```
1technophile: Maybe we can remove the line if we want to be consistent with the others removals

1technophile:
> maybe it would be useful to create an ad hoc class (maybe an extension of "client") that also manages the connection.

Indeed It would be interesting to have something more modular and that can be extended to further protocols

main/ZsensorDS1820.ino
Odyno: ok

main/config_mqttDiscovery.h
Odyno: true

1technophile:
Could you rebase it from development please, and I think we are good to go

Odyno:
@1technophile  I have a problem now... the current *development*  branch for me is out of size

```
....
Linking .pio/build/az-delivery/firmware.elf
Retrieving maximum program size .pio/build/az-delivery/firmware.elf
Checking size .pio/build/az-delivery/firmware.elf
Error: The program size (1479456 bytes) is greater than maximum allowed (1310720 bytes)
RAM:   [=         ]  11.7% (used 62276 bytes from 532480 bytes)
Flash: [===*** [checkprogsize] Explicit exit, status 1
=======]  112.9% (used 1479456 bytes from 1310720 bytes)
============================================================ [FAILED] Took 51.47 seconds ============================================================

Environment    Status    Duration
-------------  --------  ------------
az-delivery    FAILED    00:00:51.469
======================================================= 1 failed, 0 succeeded in 00:00:51.469 =======================================================
The terminal process "platformio 'run'" terminated with exit code: 1.
```

I have this configuration
```
[platformio]
default_envs = az-delivery

[env:az-delivery]
lib_ldf_mode =  deep+
platform = espressif32
board = az-delivery-devkit-v4
check_tool = cppcheck, clangtidy
check_flags =
  cppcheck: -v
  clangtidy: -fix-errors -format-style=mozilla

lib_deps =
  ${env:esp32dev-all-test.lib_deps}
build_flags =
  ${env:esp32dev-all-test.build_flags}
  '-DZgatewayBT="BT"'
  '-DZgatewayRF="RF"'
  '-DLED_RECEIVE=2'
  '-DLED_RECEIVE_ON=0'
  '-DpubUnknownBLEServiceData=false'
  '-DsimpleReceiving=true'
  '-DZmqttDiscovery="HADiscovery"'
  '-DGateway_Name="OMG_BTRF_Latest"'
  '-DESPWifiManualSetup=true'
  '-Dwifi_ssid="+++++"'
  '-Dwifi_password="+++++"'
  '-DMQTT_PORT="+++++"'
  '-DMQTT_USER="+++++"'
  '-DMQTT_PASS="+++++"'
  '-DMQTT_SERVER="+++++"'
  #'-DOMG_VERSION="0.9.6-4"'
  '-DLOG_LEVEL=LOG_LEVEL_VERBOSE'  ; Enable trace level logging
  '-DvalueAsASubject=true'

```

do you have suggestion?

1technophile:
>do you have suggestion?

Maybe by identifying the difference with this [one](https://github.com/1technophile/OpenMQTTGateway/blob/10f0d4cff488ef68e9e874c2bee9518ea5174d35/platformio.ini#L202),
one issue also could be that Arduino_json is duplicated into the binary due to the call from [decoder library](https://github.com/theengs/decoder/tree/development/src) and the direct call to it.
I have tried to avoid it but it may be the case

Odyno:
@1technophile  fixed, just used   ${com-esp.build_flags} instead of    ${env:esp32dev-all-test.build_flags}

1technophile:
Thanks

PR #1126, opened 2022-01-07
Behaviour when MQTT and/or Wifi is not available

NorthernMan54:
## Description:

I was looking into why all my OpenMQTTGateway devices were not responding after a recent power failure and am speculating that the ESP32 was able to restart faster than my Wifi and MQTT Server was able to restart and resume operations.  And was thinking that a generic ESP.restart may be a reasonable approach with repeated Wifi and/or MQTT connection failures.

For this I created a new variable `maxRetryWatchDog` which is set to 2 x maxConnectionRetryWifi + 1.

For the placement of the restart in the WiFi connection logic I did not want to affect the existing low power mode functionality, but would like a second opinion on this.  Also I looked at the Arduino spec and could not identify something similar. Also I don't have any in my test environment.

Am that this could be leveraged further for other retry loops that need a failsafe exit.

Also included a fix for the valueAsASubject mode when using multiple receivers.  The field `value` is null when the decoder is not PiLight.

As mentioned here https://community.openmqttgateway.com/t/behaviour-when-mqtt-and-or-wifi-is-not-available/1747

## Checklist:

### DIFF
--- a/main/main.ino
+++ b/main/main.ino
@@ -985,16 +985,19 @@ void setup_wifi() {
     Log.trace(F("." CR));
     failure_number_ntwk++;
 #  if defined(ESP32) && defined(ZgatewayBT)
-    if (failure_number_ntwk > maxConnectionRetryWifi && lowpowermode)
-      lowPowerESP32();
-#  elif defined(ESP32)
-    if (failure_number_ntwk > maxRetryWatchDog && !lowpowermode) {
-      watchdogReboot(2);
-    }
-#  else
+    if (lowpowermode) {
+      if (failure_number_ntwk > maxConnectionRetryWifi) {
+        lowPowerESP32();
+      }
+    } else {
       if (failure_number_ntwk > maxRetryWatchDog) {
         watchdogReboot(2);
       }
+    }
+#  else
+    if (failure_number_ntwk > maxRetryWatchDog) {
+      watchdogReboot(2);
+    }
 #  endif
   }
   Log.notice(F("WiFi ok with manual config credentials" CR));
--- a/main/main.ino
+++ b/main/main.ino
@@ -2120,7 +2120,7 @@ String toString(uint32_t input) {
   Reason Codes
   1 - Repeated MQTT Connection Failure
   2 - Repeated WiFi Connection Failure
-  3 - Failed WiFi configuration portal
+  3 - Failed WiFiManager configuration portal
 */
 void watchdogReboot(byte reason) {
   Log.warning(F("Rebooting for reason code %d" CR), reason);
--- a/main/main.ino
+++ b/main/main.ino
@@ -561,7 +561,7 @@ void connectMQTT() {
     delay(2000);
     digitalWrite(LED_ERROR, !LED_ERROR_ON);
     delay(5000);
-    if (failure_number_mqtt > maxConnectionRetryWifi) {
+    if (failure_number_mqtt > maxRetryWatchDog) {
       watchdogReboot(1);
     }
   }
@@ -988,7 +988,7 @@ void setup_wifi() {
     if (failure_number_ntwk > maxConnectionRetryWifi && lowpowermode)
       lowPowerESP32();
 #  else
-    if (failure_number_ntwk > maxConnectionRetryWifi) {
+    if (failure_number_ntwk > maxRetryWatchDog) {
       watchdogReboot(2);
     }
 #  endif
--- a/main/main.ino
+++ b/main/main.ino
@@ -987,10 +987,14 @@ void setup_wifi() {
 #  if defined(ESP32) && defined(ZgatewayBT)
     if (failure_number_ntwk > maxConnectionRetryWifi && lowpowermode)
       lowPowerESP32();
-#  else
+#  elif defined(ESP32)
     if (failure_number_ntwk > maxRetryWatchDog && !lowpowermode) {
       watchdogReboot(2);
     }
+#  else
+      if (failure_number_ntwk > maxRetryWatchDog) {
+        watchdogReboot(2);
+      }
 #  endif
   }
   Log.notice(F("WiFi ok with manual config credentials" CR));
--- a/main/main.ino
+++ b/main/main.ino
@@ -317,7 +317,7 @@ void pub(const char* topicori, JsonObject& data) {
 #  ifdef ZgatewayPilight
   String value = data["value"];
   String protocol = data["protocol"];
-  if (value != 0) {
+  if (value != "null" && value != 0) {
     topic = topic + "/" + protocol + "/" + value;
   }
 #  else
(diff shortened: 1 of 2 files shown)


1technophile:
Thanks for the PR, we use to have such a mechanism in the past but I was not a fan of it as "restarting" can be considered as the last resort option.
Nevertheless, I get your point, we may need to do some testing in particular regarding the behavior of the wifi manager when restarting without the wifi available. If it triggers the configuration portal this is not expected as it may be the opportunity to access the configuration by someone else.

Also, we use to differentiate the restart between ESP32 and ESP8266:
```
#  if defined(ESP8266)
        ESP.reset();
#  else
        ESP.restart();
#  endif
```
but as this seems to build fine, we can confirm that it works as expected with testing.
If working we will need to remove this condition to leverage only `ESP.restart()`

NorthernMan54:
I went and looked at the Tasmota code base, they use both [ESP8266](https://github.com/arendst/Tasmota/blob/e5d576b5076874def7c990ff4ee1c3db8cf8207e/tasmota/support_esp.ino#L58) and [ESP32](https://github.com/arendst/Tasmota/blob/e5d576b5076874def7c990ff4ee1c3db8cf8207e/tasmota/support_esp.ino#L421)

So I will make the update

In regards to WiFi Manager, I personally don't use it, but will take some time to look into the scenario of WiFi manager being invoked, but no response after a period of time ( 5 to 10 minutes ).  Needs to be long enough to allow people to slowly go thru the process.

1technophile:
I tested the behavior and I have the following results (ESP32 + BT):
1) OMG running on an ESP32, stop the MQTT broker, the gateway restart after `maxConnectionRetryWifi`
2) OMG running on an ESP32, stop the WIFI, the gateway keep attempting to connect to the wifi
3) Stop the wifi, Reset OMG on an ESP32, the gateway keep attempting to connect to the wifi

I'm ok with 1 and 2, but I don't think 3 is the expected behavior

main/main.ino
```diff
@@ -984,6 +987,10 @@ void setup_wifi() {
 #  if defined(ESP32) && defined(ZgatewayBT)
     if (failure_number_ntwk > maxConnectionRetryWifi && lowpowermode)
       lowPowerESP32();
+#  else
```
1technophile, line 997: Rather than an #else maybe we should just condition this function execution to not being in lowpowermode.
This would correct the result of 3) from my comment

NorthernMan54: Let me recreate this and submit an update.  Tks for taking a moment to review

1technophile: Maybe we could have also a different macro for the max connection retry than `maxConnectionRetryWifi`, with a bigger number like 30 to avoid repetitive restart

NorthernMan54: Something like a maximumRetryWatchdog, and set it to be 2 x maxConnectionRetryWifi ?

NorthernMan54: 2 x + 1

1technophile: yes would be cool

1technophile:
@NorthernMan54 I would like to close the v0.9.9 and this modification could be part of it.
Do you think you could take into account the comments?

NorthernMan54:
Yes, just need a few hours

NorthernMan54:
Once the build check completes, this should be good to include.  I tested both the missing MQTT server scenario and the missing wifi hotspot scenario and it worked as expected.  I did not test the scenario with BT ( no local devices with this config available ).

main/main.ino
```diff
@@ -984,6 +987,10 @@ void setup_wifi() {
...
+    if (failure_number_ntwk > maxRetryWatchDog) {
+      watchdogReboot(2);
+    }
 #  endif
   }
```
1technophile, line 1002: I think
```
    if (failure_number_ntwk > maxRetryWatchDog && !lowpowermode) {
      watchdogReboot(2);
    }
```

Should be after  the endif at line 994, so that this function is used when having an ESP32 + BT, we can just add `!lowpowermode` into it. But this will be only for comfort due to the fact that the ESP32 should not go this far if it has lowpowermode activated

NorthernMan54: Thinking this thru

Without BT enabled

1 - If low power mode is enabled (1 or 2 ), and Wifi is down, the unit will never reboot

2 - If low power mode is not enabled (0), and Wifi is down, the unit will reboot after maxRetryWatchDog

With BT Enabled

1 - If low power mode is enabled (1 or 2 ), and Wifi is down, the unit will not reboot, but will go to sleep after maxConnectionRetryWifi

2 - If low power mode is not enabled (0), and Wifi is down, the unit will reboot after maxRetryWatchDog

With lowpowermode mode the unit will never reboot due it bad wifi, but this patch does not change the existing behaviour in this situation.  So makes sense

NorthernMan54:
Well that didn't work out too well, needed a minor change

```
/Code/OpenMQTTGateway/main/main.ino:991:52: error: 'lowpowermode' was not declared in this scope; did you mean 'changelowpowermode'?
  991 |     if (failure_number_ntwk > maxRetryWatchDog && !lowpowermode) {
      |                                                    ^~~~~~~~~~~~
      |                                                    changelowpowermode
Compiling .pio/build/manual-wifi-test/libe6b/ESP8266WiFi/ESP8266WiFiGeneric.cpp.o
```

Promoting again

main/main.ino
```diff
@@ -984,6 +987,14 @@ void setup_wifi() {
...
     if (failure_number_ntwk > maxConnectionRetryWifi && lowpowermode)
       lowPowerESP32();
+#  elif defined(ESP32)
+    if (failure_number_ntwk > maxRetryWatchDog && !lowpowermode) {
+      watchdogReboot(2);
+    }
+#  else
+      if (failure_number_ntwk > maxRetryWatchDog) {
+        watchdogReboot(2);
+      }
```
1technophile: here is a proposal to cover the use of BT on ESP32:

```
#  if defined(ESP32)
    if (lowpowermode) {
      if (failure_number_ntwk > maxConnectionRetryWifi) {
        lowPowerESP32();
      }
    } else {
      if (failure_number_ntwk > maxRetryWatchDog) {
        watchdogReboot(2);
      }
    }
#  else
    if (failure_number_ntwk > maxRetryWatchDog) {
      watchdogReboot(2);
    }
#  endif
```

NorthernMan54: Made one final tweak as lowPowerESP32() is part of the BT module, and changed `#  if defined(ESP32)` to `#  if defined(ESP32) && defined(ZgatewayBT)`

```
#  if defined(ESP32) && defined(ZgatewayBT)
    if (lowpowermode) {
      if (failure_number_ntwk > maxConnectionRetryWifi) {
        lowPowerESP32();
      }
    } else {
      if (failure_number_ntwk > maxRetryWatchDog) {
        watchdogReboot(2);
      }
    }
#  else
    if (failure_number_ntwk > maxRetryWatchDog) {
      watchdogReboot(2);
    }
#  endif
```

1technophile: that works!

NorthernMan54:
Tks very much, now to start on my next feature, using cc1101 carrier sense to detect start and end of signal

1technophile:
With pleasure!

PR #1136, opened 2022-01-18
Use write with response when non-response write is not available.

h2zero:
## Description
If the BLE characteristic does not have write without response available this will default to using write with response.

## Checklist:


DigiH:
@h2zero - somehow this created constant  **"success":false** for my Playbulb Candles, as discussed in https://github.com/1technophile/OpenMQTTGateway/pull/1137

PR #1153, opened 2022-02-02
Fix OpenHAB discovery

kabili207:
## Description:
Enabling OpenHAB discovery breaks the build as it expects defines only present in the HASS define set

## Checklist:


1technophile:
Thanks for this, could you add also to the PR a test environment with OpenHAB auto-discovery. This will enable us to verify that the build is monitored for this configuration at each change.

kabili207:
@1technophile Test env has been added

1technophile:
Perfect!

PR #1154, opened 2022-02-04
Add TinyPICO board

kabili207:
## Description:
Adds support for the [TinyPICO](https://www.tinypico.com/) dev board. The board does have a built-in RGB LED, however RAM is at a premium on this device and I wasn't able to get it working correctly with the fastled actuator, so I've left that out of the config for now.

## Checklist:


1technophile:
Thanks!

PR #1183, opened 2022-03-10
Remove auto discovery for Ibeacon, GAEN and MS CDP

1technophile:
## Description:
to avoid creating hundred of entities and devices in the controller due to their changing MAC addresses

## Checklist:


1technophile:
@koenvervloesem are you ok with this PR? Or do you think there is another solution?
Note that this only impacts auto-discovery not the regular publication of values.

koenvervloesem:
This looks fine! No need to create hundreds of devices indeed.

Issue #1179, opened 2022-03-07
Publish BSSID as a SYS sensor


chinyaolin:
I have a card in Homeassistant to show how my wifi end devices connect to.
I use BSSID to determine whitch "AP" they connected.
IT would be nice to get OMG BSSID.
![BSSID](https://user-images.githubusercontent.com/52078139/157006712-5dcfce9d-f2b3-415c-b6b3-b14a4c245eac.jpg)

1technophile:
Hello,

Seems that a method exists for this `WiFi.BSSID()`, would you like to submit a PR to add it after [here](https://github.com/1technophile/OpenMQTTGateway/blob/a1846faee9d54f8e64fe97096f08bad37ff350ee/main/main.ino#L1558)?
```cpp
  String BSSID = WiFi.BSSID();
  SYSdata["BSSID"] = BSSID;
```

chinyaolin:
Its my first time to create a PR.
Let me know if anything wrong.

PR #1194, opened 2022-03-30
Add Wifi BSSID as SYSdata

chinyaolin:
Publish BSSID as a SYS sensor.

## Description:
[Issue 1179](https://github.com/1technophile/OpenMQTTGateway/issues/1179)

## Checklist:


1technophile:
`String BSSID = WiFi.BSSID();` returns an array of bytes, this is why the builds are failing,

Could you replace by
`  String BSSID = WiFi.BSSIDstr();`
so that the function will return a string instead.

1technophile:
Thanks!

PR #1200, opened 2022-04-14
Add ability to specify pulse length in ZactuatorONOFF

josephperry:
Add an additional parameter to the MQTT message to control the actuatorONOFF pulse length e.g. `mosquitto_pub -t home/OpenMQTTGateway_MEGA/commands/MQTTtoONOFF -m '{"gpio":15,"cmd":"high_pulse","pulse_length":25}'`

## Description:
Currently the actuatorONOFF pulse length is restricted to 500 ms. I'd like to be able to control the pulse length down to 10's of ms. I have a buzzer connected to a ESP32 port and would like to have a very short duration pulse so the buzzer produces only a short click sound. Sending a MQTT command to turn the port ON, delaying a short time, then sending a MQTT command to turn the port OFF doesn't work consistently for very short delays.

## Checklist:
  - [x ] The pull request is done against the latest development branch
  - [x ] Only one feature/fix was added per PR and the code change compiles without warnings
  - [ x] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).

### DIFF
--- a/main/ZactuatorONOFF.ino
+++ b/main/ZactuatorONOFF.ino
@@ -47,16 +47,24 @@ void MQTTtoONOFF(char* topicOri, JsonObject& ONOFFdata) {
       if (ONOFFdata["cmd"] == "high_pulse") {
         Log.notice(F("MQTTtoONOFF high_pulse ok" CR));
         Log.notice(F("GPIO number: %d" CR), gpio);
+        int pulselength = ONOFFdata["pulse_length"];
+        if (!pulselength)
+          pulselength = 500;
+        Log.notice(F("Pulse length: %d ms" CR), pulselength);
         pinMode(gpio, OUTPUT);
         digitalWrite(gpio, HIGH);
-        delay(500);
+        delay(pulselength);
         digitalWrite(gpio, LOW);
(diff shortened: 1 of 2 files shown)


main/ZactuatorONOFF.ino
```diff
@@ -47,16 +47,24 @@ void MQTTtoONOFF(char* topicOri, JsonObject& ONOFFdata) {
       if (ONOFFdata["cmd"] == "high_pulse") {
         Log.notice(F("MQTTtoONOFF high_pulse ok" CR));
         Log.notice(F("GPIO number: %d" CR), gpio);
+        int pulselength = ONOFFdata["pulse_length"];
```
1technophile: one tip with Arduino json is to define the default value like below:
` int pulselength = ONOFFdata["pulse_length"]|500;`

This will give pulselength = 500 if pulse_length is not found in the json

josephperry: Thanks, will revise and resubmit.

1technophile:
Thanks!

PR #1216, opened 2022-05-15
Add option to offload decoding to an external gateway over MQTT.

h2zero:
## Description:
This adds an option to offload the decoding of BLE data to a TheengsGateway device. This uses the functionality provided by https://github.com/theengs/gateway/pull/25.

To test:
- Make sure you have set a subscribe topic in the TheengsGateway config and that you're using the code from the PR above.
- Flash your esp with this PR and set the build flag '-DMQTTDecodeTopic="YOUR_TOPIC"' in your PIO environment.

## Checklist:

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -66,7 +66,9 @@ using namespace std;
 #  define device_flags_isBlackL 1 << 2
 #  define device_flags_connect  1 << 3
 
+#  ifndef MQTTDecodeTopic
 TheengsDecoder decoder;
+#  endif
 
 struct decompose {
   int start;
@@ -95,7 +97,16 @@ void pubBTMainCore(JsonObject& data, bool haPresenceEnabled = true) {
   if (abs((int)data["rssi"] | 0) < minRssi && data.containsKey("id")) {
     String mac_address = data["id"].as<const char*>();
     mac_address.replace(":", "");
+#  ifdef MQTTDecodeTopic
+    String mactopic;
+    if (data.containsKey("model")) {
+      mactopic = subjectBTtoMQTT + String("/") + mac_address;
+    } else {
+      mactopic = String("/") + MQTTDecodeTopic;
+    }
+#  else
     String mactopic = subjectBTtoMQTT + String("/") + mac_address;
+#  endif
     pub((char*)mactopic.c_str(), data);
   }
   if (haPresenceEnabled && data.containsKey("distance")) {
@@ -870,6 +881,7 @@ void launchBTDiscovery() {
         p->sensorModel_id != TheengsDecoder::BLE_ID_NUM::GAEN) {
       String macWOdots = String(p->macAdr);
       macWOdots.replace(":", "");
+#    ifndef MQTTDecodeTopic
       if (p->sensorModel_id > TheengsDecoder::BLE_ID_NUM::UNKNOWN_MODEL &&
           p->sensorModel_id < TheengsDecoder::BLE_ID_NUM::BLE_ID_MAX &&
           p->sensorModel_id != TheengsDecoder::BLE_ID_NUM::HHCCJCY01HHCC) { // Exception on HHCCJCY01HHCC as this one is discoverable and connectable for battery retrieving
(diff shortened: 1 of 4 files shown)


1technophile:
Works great!

platformio.ini
```diff
@@ -408,6 +408,23 @@ build_flags =
...
+  ${libraries.wifimanager32}
+  ${libraries.ble}
+  ${libraries.decoder}
+build_flags =
+  ${com-esp.build_flags}
+  '-DZgatewayBT="BT"'
+  '-DLED_SEND_RECEIVE=2'
+  '-DLED_SEND_RECEIVE_ON=0'
+  '-DGateway_Name="OpenMQTTGateway_ESP32_BLE"'
+  '-DMQTTDecodeTopic="MQTTDecode"'
```
1technophile: @h2zero could you replace `MQTTDecode` by `BTtoMQTT/undecoded`
And add also this macro into config_BT.h

h2zero: Done

1technophile:
Thanks

main/ZgatewayBT.ino
```diff
@@ -95,7 +97,16 @@ void pubBTMainCore(JsonObject& data, bool haPresenceEnabled = true) {
   if (abs((int)data["rssi"] | 0) < minRssi && data.containsKey("id")) {
     String mac_address = data["id"].as<const char*>();
     mac_address.replace(":", "");
+#  ifdef MQTTDecodeTopic
+    String mactopic;
+    if (data.containsKey("model")) {
+      mactopic = subjectBTtoMQTT + String("/") + mac_address;
+    } else {
+      mactopic = String("/") + MQTTDecodeTopic;
```
BadWolf42, line 105: Is it intended that `mactopic` begins with `/` or is `subjectBTtoMQTT` is missing ?
(I'm reviewing this code to merge PR #1220 with it and this seems strange)

DigiH: Hi @BadWolf42, for the new undecoded messages isn't it then **/** + **MQTTDecodeTopic**, which in this case is
```
// Uncomment to send undecoded device data to another gateway device for decoding
// #define MQTTDecodeTopic    "BTtoMQTT/undecoded"
```

`'-DMQTTDecodeTopic="BTtoMQTT/undecoded"'`

-> '/BTtoMQTT/undecoded'

so the `subjectBTtoMQTT` part (/BTtoMQTT) is already defined.

BadWolf42: Ok, so you confirm that L105 is OK and should **NOT** be:
`mactopic = subjectBTtoMQTT + String("/") + MQTTDecodeTopic;` 
to publish in `/BTtoMQTT/undecoded` with `'-DMQTTDecodeTopic="undecoded"'`
Right?

DigiH: I'll better leave the confirmation to @h2zero or @1technophile why it was decided to define it this way ;) and yes, your variation

mactopic = subjectBTtoMQTT + String("/") + MQTTDecodeTopic;

would also be valid with MQTTDecodeTopic just being

```
// Uncomment to send undecoded device data to another gateway device for decoding
// #define MQTTDecodeTopic    "undecoded"
```

`'-DMQTTDecodeTopic="undecoded"'`

> (Forgive me but a heading `/` is not very frequent in mqtt)

but either way, even **subjectBTtoMQTT** has a heading `/`, and in the end `mactopic` will always be a trailing substring after `home/OpenMQTTGateway_ESP32_BLE`

I'm very much looking forward to your iBeacon UUID PR as well, but assume that since it also won't decoder if the flag is set, a similar implementation would also be needed in the receiving Theengs Gateway end when it receives the raw manufacturerdata.

1technophile: I think L105 should be :
`mactopic = subjectBTtoMQTT + String("/") + MQTTDecodeTopic;`

And L56 of config_BT.h should be:
`// #define MQTTDecodeTopic    "undecode"`

As @DigiH stated.
@badWolf42 feel free to introduce this into your PR

BadWolf42: Ok thank you both for the confirmation, I'll update my PR accordingly 👍

Issue #1139, opened 2022-01-18
Ibeacons with random MAC addresses


scribblrsam:
**Is your feature request related to a problem? Please describe.**

I would like to use the home assistant companion app to publish beacons over ble, collect these and track presence of people in the house like this. As far as I can see this is not currently possible as modern phones randomise the Bluetooth MAC address used and presence events are published with the mac as the device ID.

**Describe the solution you'd like**

Publish ibeacon presence events to HA/MQTT using the beacon uuid not the MAC or perhaps both to allow either event to be tracked. I am happy to look at this, would you consider a PR for this?

**Describe alternatives you've considered**

Carry a hardware ibeacon, clearly less practical and I don't always have the same set of keys!

I have also considered using esphome as this has functionally to track beacons by uuid, I would rather use OMG as I use this for other sensors. 

**Additional context**

1technophile:
>Publish ibeacon presence events to HA/MQTT using the beacon uuid not the MAC or perhaps both to allow either event to be tracked. I am happy to look at this, would you consider a PR for this?

You should [get the UUID](https://github.com/theengs/decoder/blob/3abed86931ffec8890ced8a3b12a05a55f92bd05/src/devices/iBeacon_json.h#L13) already with an ibeacon

With the v0.9.9 beta

Which version are you using?

scribblrsam:
> Publish ibeacon presence events to HA/MQTT using the beacon uuid not the MAC or perhaps both to allow either event to be tracked. I am happy to look at this, would you consider a PR for this?

Sorry this was meant to say: 
Publish ibeacon presence events to HA/MQTT using the beacon uuid **as the device ID and** not the MAC or perhaps both to allow either event to be tracked. I am happy to look at this, would you consider a PR for this?

The messages are published to the `home/home_presence/<room name>` topics
`{"id":"50:54:99:B4:00:18","rssi":-82,"distance":11.4949,"brand":"GENERIC","model":"IBEACON","model_id":"IBEACON","mfid":"4c00","uuid":"e1e4448b54ea4a7eafc833522f901b6c","major":100,"minor":1,"power":-59}`

The mac addresses used for the ID are ever changing and therefore cannot be used to track a device according to [the HA docs for MQTT room presence](https://www.home-assistant.io/integrations/mqtt_room/). Using the uuid as the ID would allow these beacons to be tracked. 

[...]

Issue #1141, opened 2022-01-20
Deactivate jsonPublishing from `build_flags` in  `*_env.ini`


debuglevel:
**Describe the bug**
If I see it correctly `jsonPublishing` cannot be deactivated without actually editing `main/User_config.h`:

If you would try to undefine it via 
```ini
build_flags =
  ${env:esp32-lolin32lite-ble.build_flags}
  '-UjsonPublishing'
```
it will be set again in `main/User_config.h`:

```c
#ifndef jsonPublishing
#  define jsonPublishing true //comment if you don't want to use Json  publishing  (one topic for all the parameters)
#endif
```` 

Setting
```ini
build_flags =
  ${env:esp32-lolin32lite-ble.build_flags}
  '-DjsonPublishing=false'
```
probably also has no effect, as the code in `main.ino` just checks if it is defined, but not for its value:
```c
#ifdef jsonPublishing
    Log.trace(F("jsonPublishing" CR));
#  if defined(ESP8266) || defined(ESP32) || defined(__AVR_ATmega2560__) || defined(__AVR_ATmega1280__)
    char JSONmessageBuffer[data.measureLength() + 1];
#  else
    char JSONmessageBuffer[JSON_MSG_BUFFER];
#  endif
    data.printTo(JSONmessageBuffer, sizeof(JSONmessageBuffer));
    pubMQTT(topic, JSONmessageBuffer);
#endif
```

Unless I'm mistaken (sorry then!), one cannot deactivate `jsonPublishing` from in  `*_env.ini` (btw: which is great for the other configurations!)

I think `#if jsonPublishing == false` in `main.ino` could do the trick. But I'm a lousy C programmer and would rather leave the solution up to someone who know's what he's doing :)

**Environment (please complete the following information):**
* tag `v0.9.8` from git
* compiled via `platformio`

1technophile:
Good catch,
what is the use case for deactivating it? For my curiosity.

I'm planning from a long time to remove the `simplePublishing` and to have permanently `jsonPublishing`

debuglevel:
To be honest, I just wanted to start with the "plain MQTT" thing for now - and I like the "hierarchical topic approach" of MQTT because it is very simple and easy to understand, even though JSON is probably also great for backend processing.

[...]

PR #1220, opened 2022-05-29
iBeacon: Flag to use UUID in topic or presence

BadWolf42:
## Description:
Inspired by issue https://github.com/1technophile/OpenMQTTGateway/issues/1139

Add compilation flags to use UUID in topic, instead of sender (random) mac address:
FLAG `useBeaconUuidForTopic` to use UUID in topic
FLAG `useBeaconUuidForPresence` to use UUID in presence
This features is disabled by default, usual behavior is maintained

The question of whitelisting/blacklisting by UUID for a iBeacon, as exposed in the issue, is not approached in this commit.
(WhtL/BlkL is filtered prior to `pubBTMainCore()` function, so it does not seems to be as simple as this PR)

## Checklist:

Thanks, Bad

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -94,7 +94,7 @@ int minRssi = abs(MinimumRSSI); //minimum rssi value
 void pubBTMainCore(JsonObject& data, bool haPresenceEnabled = true) {
   if (abs((int)data["rssi"] | 0) < minRssi && data.containsKey("id")) {
     String topic = data["id"].as<const char*>();
-    topic.replace(":", "");
+    topic.replace(":", "");
 #  ifdef useBeaconUuidForTopic
     if (data.containsKey("model_id") && data["model_id"].as<String>() == "IBEACON") {
       topic = data["uuid"].as<const char*>();
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -95,20 +95,20 @@ void pubBTMainCore(JsonObject& data, bool haPresenceEnabled = true) {
   if (abs((int)data["rssi"] | 0) < minRssi && data.containsKey("id")) {
     String topic = data["id"].as<const char*>();
     topic.replace(":", "");
-#  if useBeaconUuidForTopic
-    if (data.containsKey("model_id") && data["model_id"].as<String>() == "IBEACON") {
-      topic = data["uuid"].as<const char*>();
+#  ifdef useBeaconUuidForTopic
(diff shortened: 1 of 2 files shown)


1technophile:
Thanks for the PR, could you add this macro into [config_BT](https://github.com/1technophile/OpenMQTTGateway/blob/development/main/config_BT.h) please

1technophile:
@BadWolf42

BadWolf42:
Yes sure, I will tomorrow

BadWolf42:
Hello, it's done.

Please let me know if I implemented this feature in it the "correct way".
Especially regarding the use of `#ifdef` directives, and not `#if`, as I spotted in `config_BT.h`. I can quickly change to `#if`, but in my logic there is no need to test the "content" value of a flag, just if this here or not.

DigiH:
Hi @BadWolf42 , as a personal preference and suggestion for clearer definition I would define the flags with
```

#ifndef useBeaconUuidForTopic
#  define useBeaconUuidForTopic false // define true to use iBeacon UUID as topic, instead of sender (random) mac address
#endif

…

#ifndef useBeaconUuidForPresence
#  define useBeaconUuidForPresence false // //define true to use iBeacon UUID as for presence, instead of sender mac (random) address
#endif
```

similar to the BLEServiceData/ManufacturerData publishing definitions, and then use **if** directives for evaluation.

This would allow for easily adjustable custom environment build flag definitions

`'-DuseBeaconUuidForTopic=true'`

being able to true or false quickly, instead of a more cumbersome commenting and uncommenting, defining/undefining

`'-DuseBeaconUuidForTopic'`

`'-UuseBeaconUuidForTopic'`

P.S.: Just realised that the jsonPublishing issue has already been logged ;)

Just a thought :)

BadWolf42:
Ok, this PR is getting way too messy (already was with the white space issues).

I'll start a new PR from dev branch with a sequence of 4 changes:
- Add useBeaconUuidForPresence feature
- [Change regarding MQTTDecodeTopic](https://github.com/1technophile/OpenMQTTGateway/pull/1216/files#r897361158) and code refactor
- Add useBeaconUuidForTopic feature
- Fix jsonPublishing (regarding [Issue 1141](https://github.com/1technophile/OpenMQTTGateway/issues/1141)?) there is an `#ifdef` left [here](https://github.com/1technophile/OpenMQTTGateway/blob/e43f7b3c15c9cbb751bdd9160bdf82540341bc59/main/main.ino#L331) that should be an `#if`

PR #1226, opened 2022-06-15
iBeacon: Flag to use UUID in topic or presence (and fixes)

BadWolf42:
## Description:
Inspired by issue https://github.com/1technophile/OpenMQTTGateway/issues/1139
Replaces PR #1220

Add compilation flags to use UUID in topic, instead of sender (random) mac address:
FLAG `useBeaconUuidForTopic` to use UUID in topic
FLAG `useBeaconUuidForPresence` to use UUID in presence
This features is disabled by default, usual behavior is maintained
_The question of whitelisting/blacklisting by UUID for a iBeacon, as exposed in the issue, is not approached in this commit.
(WhtL/BlkL is filtered prior to `pubBTMainCore()` function, so it does not seems to be as simple as this PR)_

[Change introduced regarding MQTTDecodeTopic](https://github.com/1technophile/OpenMQTTGateway/pull/1216/files#r897361158) and code refactor
Fix jsonPublishing (regarding https://github.com/1technophile/OpenMQTTGateway/issues/1141?) there is an #ifdef left [here](https://github.com/1technophile/OpenMQTTGateway/blob/e43f7b3c15c9cbb751bdd9160bdf82540341bc59/main/main.ino#L331) that should be an #if

## Checklist:

Thanks, Bad

### DIFF
--- a/main/main.ino
+++ b/main/main.ino
@@ -328,7 +328,7 @@ void pub(const char* topicori, JsonObject& data) {
 #  endif
 #endif
 
-#ifdef jsonPublishing
+#if jsonPublishing
   Log.trace(F("jsonPubl - ON" CR));
   pubMQTT(topic, dataAsString.c_str());
 #endif
--- a/main/main.ino
+++ b/main/main.ino
@@ -328,7 +328,7 @@ void pub(const char* topicori, JsonObject& data) {
 #  endif
 #endif
 
-#if jsonPublishing
+#ifdef jsonPublishing
   Log.trace(F("jsonPubl - ON" CR));
   pubMQTT(topic, dataAsString.c_str());
 #endif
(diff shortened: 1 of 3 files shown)


main/main.ino
```diff
@@ -328,7 +328,7 @@ void pub(const char* topicori, JsonObject& data) {
 #  endif
 #endif
 
-#ifdef jsonPublishing
+#if jsonPublishing
```
1technophile: I would keep this as it is, as users may have defined with another value than `true` in their configuration, also for purpose of consistency with other macro definitions.

BadWolf42: Ok, sure I can revert this file
But I think this `#ifdef` has no sense here as `jsonPublishing` is always define:
https://github.com/1technophile/OpenMQTTGateway/blob/25b8123ff6471352951d3def22c815017f5f11e1/main/User_config.h#L355-L357

1technophile: If we do this I think we should replace the comment, and say to the user that it should be `false` instead of commenting.
I would rather do this in a separate pr and align also `simplePublishing` macro for consistency sake.

BadWolf42: I agree, commit 02e2d03 reverted by commit 019f9e8

1technophile:
Thanks

PR #1229, opened 2022-06-26
Publishing/receiving inconsistencies fix

DigiH:
With the inconsistencies between the definitions and implementations of
jsonPublishing
jsonReceiving
simplePublishing
simpleReceiving
and to avoid possible resulting confusion, I have changed them to all adhere to a true/false definition.

While this will break some users' current implementations for working around the inconsistencies, it should be very easily modifiable in their environments. Should be clearly communicated as a breaking change in the release notes, if you agree with these changes.

## Checklist:


1technophile:
Thanks !

PR #1230, opened 2022-06-28
Add command example to flash with esptool.py

spinpwr:
## Description:
If you are on Linux and don't want to (or can't) install Chrome based browser to upload the firmware you can use the Python based flash tool from Espressif. This can be installed in user space without special right needed.
It took me some searching and reading the docs of esptool to get the commands to work, so I thought I can save the others from this by adding the relevant info to this how-to.

## Checklist:


1technophile:
Thanks !

PR #1231, opened 2022-06-29
Lora band define

DigiH:
This should fix the LORA_BAND defines working properly in the platformio.ini environments and binary builds, not being overwritten by the default define in config_LORA.h, as pointed out in the forum

https://community.openmqttgateway.com/t/heltec-wifi-lora-32-915-option-1-upload-from-the-web-gateway-appears-to-be-set-to-the-wrong-frequency/1933/5

## Checklist:


DigiH:
Finally :) I thought my formatter was all set up to read the definition correctly, but obviously it wasn't.

1technophile:
Thanks!

PR #1234, opened 2022-07-12
Configure LoRA gateway parameters at startup

kamaradclimber:
Without this (and without sending an initial MQTT message to the gateway
to configure it) messages are not necessarily received if parameters are
different from firmware defaults.

This patch helped at least two users:
- https://community.home-assistant.io/t/awesome-lora-soil-sensor/304351/23
- https://community.home-assistant.io/t/awesome-lora-soil-sensor/304351/18

## Checklist:

### DIFF
--- a/main/ZgatewayLORA.ino
+++ b/main/ZgatewayLORA.ino
@@ -49,6 +49,12 @@ void setupLORA() {
   Log.notice(F("LORA_SS: %d" CR), LORA_SS);
   Log.notice(F("LORA_RST: %d" CR), LORA_RST);
   Log.notice(F("LORA_DI0: %d" CR), LORA_DI0);
+  LoRa.setTxPower(LORA_TX_POWER);
+  LoRa.setSpreadingFactor(LORA_SPREADING_FACTOR);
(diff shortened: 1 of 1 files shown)


main/ZgatewayLORA.ino
```diff
@@ -49,6 +49,13 @@ void setupLORA() {
   Log.notice(F("LORA_SS: %d" CR), LORA_SS);
   Log.notice(F("LORA_RST: %d" CR), LORA_RST);
   Log.notice(F("LORA_DI0: %d" CR), LORA_DI0);
+  LoRa.setTxPower(LORA_TX_POWER);
+  LoRa.setFrequency(LORA_BAND);
```
1technophile: Is this not redundant with the LORA_BAND setup during the `begin` above?

kamaradclimber: That's a good question, let me check what this does

kamaradclimber: Indeed https://github.com/sandeepmistry/arduino-LoRa/blob/master/API.md#radio-parameters is explicit about this. Thanks for the feedback.

1technophile:
Thanks

PR #1235, opened 2022-07-13
Proposal for ble-datatest-environment

DigiH:
Suggestion for adding a BLE DataTest environment which includes servicedata, manufacturerdata and service uuid even if a device has been decoded.

For easy downloads and (web)installs for users when these infos might be necessary for issue investigations. 

While this would only apply for **esp32dev-ble** environment it would probably cover many BLE cases.

## Checklist:


1technophile:
Thanks

1technophile:
Could you list it also into the upload.md page with a short explanation, for users directly looking for this ?

DigiH:
Totally forgot, thanks. Done.

PR #1236, opened 2022-07-13
Typo in doc for ESP32 config

BadWolf42:
## Description:
Typo in doc [Change the MQTT main topic and or gateway name](https://docs.openmqttgateway.com/use/gateway.html#change-the-mqtt-broker-credentials).

Missing `"` (double quote) makes the json invalid.

By the way (just writing this here):
In v0.9.13, if `mqtt_topic` and `gateway_name` are changed at the same time, then only `gateway_name` is set, 
Resulting in base topic `<gateway_name>/`, instead of `<mqtt_topic>/<gateway_name>/`.
(Sorry I had no more time to open a second issue or PR)

## Checklist:

Bad


1technophile:
Thanks

Issue #1237, opened 2022-07-15
Wrong base topic if gateway_name is changed


BadWolf42:
**Describe the bug**
Using `home/OpenMQTTGateway/commands/MQTTtoSYS/config` topic :
If `mqtt_topic` and `gateway_name` are changed at the same time, then only `gateway_name` is set,
Resulting in base topic `<gateway_name>/`, instead of `<mqtt_topic>/<gateway_name>/`.

&nbsp;
**To Reproduce**
Each time on a fresh install with base topic set to `home/OpenMQTTGateway/`:

Publish `{"mqtt_topic":"bt/"}` on `home/OpenMQTTGateway/commands/MQTTtoSYS/config`
Base topic becomes `bt/OpenMQTTGateway/` -> OK

Publish `{"gateway_name":"OMG"}` on `home/OpenMQTTGateway/commands/MQTTtoSYS/config`
Base topic becomes `home/OMG/` -> OK

Publish `{"mqtt_topic":"bt/","gateway_name":"OMG"}` on `home/OpenMQTTGateway/commands/MQTTtoSYS/config`
Base topic becomes `OMG/` -> **KO**

&nbsp;
**Expected behavior**
When publishing `{"mqtt_topic":"bt/","gateway_name":"OMG"}` on `home/OpenMQTTGateway/commands/MQTTtoSYS/config`
Base topic should becomes `bt/OMG/`

&nbsp;
**Environment (please complete the following information):**
OpenMQTTGateway version used V0.9.13, esp32-lolin32lite-ble on ESP32

&nbsp;
**Checks:**
Code responsible of this feature is really straightforward PR #1053:
https://github.com/1technophile/OpenMQTTGateway/blob/v0.9.13/main/main.ino#L2070-L2080

I don't understand why only 1 parameter works at the time.
Maybe `stopProcessing()` is required before `saveMqttConfig()` and `startProcessing()` after `client.disconnect()`...

1technophile:
Thanks for pointing that out, any idea @h2zero ?

BadWolf42:
I just re-tested with v0.9.14 to be sure.
Actually I was wrong.
I did not test in this order and can tel that each time `gateway_name` is set, base topic becomes `<gateway_name>/`.

So (Each time on a fresh install with base topic set to `home/OpenMQTTGateway/`):
1. Publish `{"mqtt_topic":"bt/"}` on `home/OpenMQTTGateway/commands/MQTTtoSYS/config`
[...]

Issue #716, opened 2020-08-02
MQTT user name


Pyromane:
**Describe the bug**
The MQTT user name entered in the web interface is not used when establishing a connection if you change it

**To Reproduce**
Here are the serial debug outputs when changing the user name:
The variable "mqtt_user": "" appears to be empty after the restart and the connection attempts
```
*WM: [2] <- HTTP WiFi save  <\r><\n>
*WM: [3] Method: POST<\r><\n>
*WM: [2] Parameters <\r><\n>
*WM: [2] -------------------- <\r><\n>
*WM: [2] server: 192.168.0.108<\r><\n>
*WM: [2] port: 1883<\r><\n>
*WM: [2] user: mqsensor<\r><\n>
*WM: [2] pass: PASSWORD<\r><\n>
*WM: [2] name: OpenMQTTGateway_ESP32_BLE<\r><\n>
*WM: [2] topic: home/<\r><\n>
*WM: [2] -------------------- 

*WM: [1] config portal exiting <\r><\n>
{"mqtt_server":"192.168.0.108","mqtt_port":"1883","mqtt_user":"","mqtt_pass":"PASSWORD","mqtt_topic":"home/","gateway_name":"OpenMQTTGateway_ESP32_BLE"}N: BLEinterval: 55555<\n>
```
Here with standard values:
```
*WM: [2] <- HTTP WiFi save  <\r><\n>
*WM: [3] Method: POST<\r><\n>
*WM: [2] Parameters <\r><\n>
*WM: [2] -------------------- <\r><\n>
*WM: [2] server: 192.168.0.108<\r><\n>
*WM: [2] port: 1883<\r><\n>
*WM: [2] user: your_username<\r><\n>
*WM: [2] pass: your_password<\r><\n>
*WM: [2] name: OpenMQTTGateway_ESP32_BLE<\r><\n>
*WM: [2] topic: home/<\r><\n>

*WM: [1] config portal exiting <\r><\n>
{"mqtt_server":"192.168.0.108","mqtt_port":"1883","mqtt_user":"your_username","mqtt_pass":"your_password","mqtt_topic":"home/","gateway_name":"OpenMQTTGateway_ESP32_BLE"}N: BLEinterval: 55555<\n>
N: Minrssi: -100<\n>
```

**Environment (please complete the following information):**
* OpenMQTTGateway v0.9.4 + OpenMQTTGateway v0.9.3
* esp32dev-ble-firmware.bin
[...]

PR #1242, opened 2022-07-17
Fix wrong base topic if gateway_name is changed

BadWolf42:
## Description:
Related to issue #1237:

- Rectify and match parameters sizes by:
Limiting `mqtt_topic` & `gateway_name` to `parameters_size` instead of `mqtt_topic_max_size`,
Declaring `mqtt_topic` & `gateway_name` with a size of `parameters_size + 1`,

- Increase `mqtt_topic_max_size` from 100 to 150 for common platforms and from 50 to 75 for others.
So that at least `<max mqtt_topic><max gateway_name>/commands/MQTTtoSYS/config` could fit in it.

- Rewrite `cmpToMainTopic()` function to be safer (and faster) when comparing C-strings.

## Checklist:


1technophile:
LGTM, referencing this for info https://github.com/1technophile/OpenMQTTGateway/issues/716, but as we are now with a `parameter_size `of 60 already, this is equal to the 30 * 2 that was set with the referenced PR

PR #1244, opened 2022-07-18
Typo fixes II

DigiH:
Ok, decided to scan through all the docs, instead of only logging when I stumble across some typos. Also upper casing of brand names like Arduino and Sony.

## Checklist:


h2zero:
Good work with these!

DigiH:
Now I'm even thinking we should have a consistent casing for MQTT Mqtt, mqtt ... but enough for today. Sort of cooled off a bit here from the blazing heat of the day, so off to bed soon ;)

1technophile:
Thanks

PR #1245, opened 2022-07-19
Casing consistency

DigiH:
Casing consistency for MQTT, MAC, IP and mDNS in running text only.

I think that should be it with typos from me for now, before it gets too nitpicky ;)

## Checklist:


1technophile:
Thanks

PR #1248, opened 2022-07-24
Bump Theengs Decoder to v0.6.2

1technophile:
## Description:
Bump Theengs Decoder from v0.4.0 to v0.6.2

## Checklist:

### DIFF
--- a/docs/prerequisites/devices.md
+++ b/docs/prerequisites/devices.md
@@ -18,10 +18,12 @@ Added to that it retrieves the measures from the devices below. By default the d
 
 |Devices|Model|Measurements|
 |-|:-:|:-:|
+| Amazfit Smart Watch(1)||steps, heart rate|
 | ATorch Battery Capacity Monitor (c)|DT24|volt/amp/watt|
 | BLE watches with fixed MAC||rssi for presence detection|
 | BLE beacons keychains||rssi for presence detection|
 | BlueMaestro|TempoDisc|temperature/humidity/duepoint/voltage|
+| BM2 Battery Monitor|BM2|battery|
 | ClearGrass |CGG1|temperature/humidity/battery|
 | ClearGrass alarm clock|CGD1|temperature/humidity|
 | ClearGrass with atmospheric pressure |CGP1W|temperature/humidity/air pressure|
@@ -34,7 +36,7 @@ Added to that it retrieves the measures from the devices below. By default the d
 | HONEYWELL |JQJCY01YM|formaldehyde/temperature/humidity/battery|
 | Ibeacon|protocol|UUID/MFID/Major/Minor/Power|
 | INKBIRD (1)|IBS-TH1|temperature/humidity/battery|
-| INKBIRD (1)|IBS-TH2|temperature/battery|
+| INKBIRD (1)|IBS-TH2/P01B|temperature/battery|
 | INKBIRD (1)|IBT-2X|temperature1/temperature2|
 | INKBIRD (1)|IBT-4XS|temperature1/temperature2/temperature3/temperature4|
 | INKBIRD (1)|IBT-6XS|temperature1/temperature2/temperature3/temperature4/temperature5/temperature6|
@@ -44,9 +46,15 @@ Added to that it retrieves the measures from the devices below. By default the d
 | Qingping |CGDK2|temperature/humidity|
 | Qingping |CGH1|open|
 | Qingping |CGPR1|presence/luminance|
+| RBaron |b-parasite|moisture, temperature, humidity, luminance (v1.1.0+), voltage|
 | RuuviTag Raw V1|RuuviTag|temperature/humidity/pressure/acceleration/volt|
 | RuuviTag Raw V2|RuuviTag|temperature/humidity/pressure/acceleration/volt/TX power/movement/counter/sequence number|
-| Switchbot(c)|S1|mode/state/battery|
+| SmartDry|Laundry Sensor|temperature/humidity/shake/voltage/wake|
+| Switchbot|Bot(c)|mode/state/battery|
+| Switchbot|Motion Sensor|movement/light level/sensing distance/led/scope tested/battery|
+| Switchbot|Contact Sensor|contact/movement/scope tested/light level/battery|
+| Switchbot|Curtain|motion state/position/light level/battery/calibration state|
+| Switchbot|Meter (Plus)|temperature/humidity/battery|
 | Thermobeacon|WS02|temperature/humidity/volt|
 | Thermobeacon|WS08|temperature/humidity/volt|
 | TPMS|TPMS|temperature/pressure/battery/alarm/count|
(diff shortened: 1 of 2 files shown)


DigiH:
Hi @1technophile, I saw these PR builds failing and had a look into it.

https://github.com/DigiH/OpenMQTTGateway/commits/build-cleanup

With always getting

```
Library Manager: Installing FastLED @ 3.3.2
Library Manager: Warning! More than one package has been found by FastLED @ 3.3.2 requirements:
Library Manager:  - fastled/FastLED@3.5.0
Library Manager:  - etidbury/FastLED@3.3.3
Library Manager: Please specify detailed REQUIREMENTS using package owner and version (shown above) to avoid name conflicts
Downloading...
Unpacking...
Library Manager: FastLED@3.3.2 has been installed!
Library Manager: Installing OneWire
Library Manager: Warning! More than one package has been found by OneWire requirements:
Library Manager:  - paulstoffregen/OneWire@2.3.7
Library Manager:  - sensesp/OneWire@2.0.0
Library Manager:  - ivankravets/OneWire@0.0.0+sha.7bd102cb73e0
Library Manager:  - mbed-hudakz/OneWire@0.0.0+sha.7f7759008807
Library Manager:  - mbed-alpov/OneWire@0.0.0+sha.445fe6e6bd68
Library Manager:  - mbed-fblanc/OneWire@0.0.0+sha.8c4e1841eb30
Library Manager:  - mbed-jack-zen/OneWire@0.0.0+sha.a47e8ec71a4e
Library Manager:  - mbed-waf/OneWire@0.0.0+sha.79a1b7700cf2
Library Manager:  - mbed-sroy/OneWire@0.0.0+sha.7c6dd6bc20e4
Library Manager:  - mbed-simonbarker/Onewire@0.0.0+sha.d961f715d82b
Library Manager: Please specify detailed REQUIREMENTS using package owner and version (shown above) to avoid name conflicts
Downloading...
Unpacking...
Library Manager: OneWire@2.3.7 has been installed!
```

I thought it would be a good idea to specify the **FastLED** and **OneWire** repos, also bumping FastLED to 3.4.

I also added the additional information for the two Xiaomo Mi scales in the **devices.md**

While the build is are still failing with

```
/home/runner/.platformio/packages/toolchain-xtensa/bin/../lib/gcc/xtensa-lx106-elf/10.3.0/../../../../xtensa-lx106-elf/bin/ld: address 0x3fffc140 of .pio/build/nodemcuv2-all-test/firmware.elf section `.bss' is not within region `dram0_0_seg'
/home/runner/.platformio/packages/toolchain-xtensa/bin/../lib/gcc/xtensa-lx106-elf/10.3.0/../../../../xtensa-lx106-elf/bin/ld: address 0x3fffc140 of .pio/build/nodemcuv2-all-test/firmware.elf section `.bss' is not within region `dram0_0_seg'
collect2: error: ld returned 1 exit status
```

I think it would be a good idea to add the above changes

Added the **devices.md** changes here, but keeping the library changes separate, waiting to see what you think.

1technophile:
@DigiH I think we need to divide the nodemcu test environment into 2 env, it may be a memory issue.

DigiH:
Shall we still define the two libraries, maybe not with updating FastLED to 3.4, but leaving it at 3.3.2, to be on the safe side for now? So

```
fastled/FastLED@3.3.2
paulstoffregen/OneWire@2.3.7
```

1technophile:
> but leaving it at 3.3.2, to be on the safe side for now?

Agreed

DigiH:
Done, will you do the environment split, as I'm not exactly sure how you'd want to divide it?

1technophile:
> fastled/FastLED@3.3.2
paulstoffregen/OneWire@2.3.7

Yes we can do that

h2zero:
Looks like the new decoder version is using up all the ram on the 8266.

DigiH:
Yes, it was coming with `nodemcuv2-all-test` with the last integrated decoder version 0.4.0 already close to the edge with

RAM:   [========= ]  90.0% (used 73700 bytes from 81920 bytes)
Flash: [=======   ]  71.2% (used 743520 bytes from 1044464 bytes)

And lots of new devices in 0.5.0, 0.6.0 and 0.6.1 

@1technophile wants to split the environment up so we should be back on track - for a while ;)

1technophile:
Removing BT from the `nodemcuv2-all-test` env now as it is already tested into `nodemcuv2-ble`

DigiH:
> now as it is already tested into `nodemcuv2-ble`

even that is getting close now though

RAM:   [==========]  95.7% (used 78360 bytes from 81920 bytes)
Flash: [======    ]  61.5% (used 641944 bytes from 1044464 bytes)

1technophile:
By the way not sure if people are still using an ESP8266 + HM10, also we could optimize the RAM footprint of the decoder

DigiH:
One additional idea I had for Decoder was to introduce a masking function, which would allow us to cut down on several multiple versions of the same property like batt/_batt to slim down things further.

1technophile:
> was to introduce a masking function

Do you have an example of what you would like to do?

h2zero:
I would start here https://github.com/theengs/decoder/tree/reduce-and-optimize

DigiH:
> Do you have an example of what you would like to do?

Not yet, probably best to start a Decoder discussion about it, if best to implement within the `"value_from_hex_data"` function already, or as a `"post_proc"` where we already have the "&" for AND though.

h2zero:
It's a good beginning but as the decoders are getting larger more needs to be done. TBH ArduinoJson will be the issue for the most part as the memory handling is not ideal, it's just what was used by OMG when this project started, and now continues. Unfortunately, the alternative options would require a major refactor and not worth it to support outdated devices IMO.

DigiH:
Well this will go through now, and I think it'll be fine for the next and likely a few more OMG releases, but realistically the bread and butter of Decoder is to have more and more devices with time, so we'll have to think about the best way forward for optimisation, and as @1technophile mentioned above, possibly even cutting the ESP8266/HM10 combo at some stage.

h2zero:
Absolutely, there are many optimizations of the decoders that can and should be done, but this is already outside the scope of the PR. Best to move to decoder discussions.

DigiH:
Why has nobody ever told me that the 

"&" Logical And the values

post_proc we have is the masking function I was talking about above?? :blush:

This will really give us a lot of saved space, when all the relevant multi properties are pulled together, removing conditions for both variations and only having one property calculation.

@1technophile - a quick test with what I meant

https://github.com/DigiH/decoder/commit/37ba06182341bafe9be1d4ba5a2755a4ee4a2f3e

h2zero:
https://github.com/theengs/decoder/blob/b752cfcf7c078867dc408d71b53a669187b6c46f/docs/participate/adding-decoders.md?plain=1#L185

DigiH:
Yes, now I know ;) just quickly changing all the relevant decoders.

https://github.com/theengs/decoder/releases/tag/v0.6.2

PR #1253, opened 2022-08-09
Add missing macType param to NO_DEVICE_FOUND

kabili207:
## Description:
This addresses an issue that was created back when the `mac_type` was first added back in January with f8b4b3ca21dc73bcb9deb06e466ed31cb14232fa, which was causing every BLE device to be marked as connectable when first detected. Combine this with `BLE_FILTER_CONNECTABLE 1` and suddenly every device is getting filtered out before OMG has the chance to initialize anything.

## Checklist:

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -84,6 +84,7 @@ vector<BLEdevice*> devices;
 int newDevices = 0;
 
 static BLEdevice NO_DEVICE_FOUND = {{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0},
+                                    0,
                                     false,
                                     false,
                                     false,
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -83,7 +83,7 @@ vector<BLEAction> BLEactions;
 vector<BLEdevice*> devices;
 int newDevices = 0;
 
-static BLEdevice NO_DEVICE_FOUND = {{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0},
+static BLEdevice NO_DEVICE_FOUND = {{0},
                                     0,
                                     false,
                                     false,


kabili207:
I should point out that there is also a mismatch regarding `macAddr` as well: it's defined as `char[18]`, but we're only initializing 12 values. It's not really a functional issue though, since the missing array values will still get initialized to `0` at compile time.

h2zero:
Good catch!

> I should point out that there is also a mismatch regarding macAddr as well: it's defined as char[18], but we're only initializing 12 values.

I don't agree here though, the mac (ID) when creating the new BLEDevice uses the string value of the mac returned, which contains 12 chars of the address + 5* ":" separators, and of course the null char at the end. The ":"s are then later removed when publishing, which is what you're seeing in the logs/mqtt messages. 
Just wanted to clarify, certainly an area for optimization though. Again, good catch!

kabili207:
It was driving me up the wall trying to trace the issue since I wanted to enable that flag. These little quirks in C/C++ are always fun to find.

The mismatch I was referring to was in the NO_DEVICE_FOUND definition, where we're setting it to `{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}`, which is only 12 zeroes instead of the expected 18. It's still technically valid since the missing 6 array values will get defaulted to 0, but it did stick out to me since I was already counting parameters in the same area.

h2zero:
> It was driving me up the wall trying to trace the issue since I wanted to enable that flag. These little quirks in C/C++ are always fun to find.

Keeps me entertained lol, things like this are a real pain to find though 👍 

> The mismatch I was referring to was in the NO_DEVICE_FOUND definition, where we're setting it to {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}, which is only 12 zeroes instead of the expected 18.

Yes, you're correct, the whole thing could just be changed to {0} for this case. If you're inclined to do so you could add that to this PR.

kabili207:
> Yes, you're correct, the whole thing could just be changed to {0} for this case. If you're inclined to do so you could add that to this PR.

Added! Better to fix it now rather than later.

h2zero:
Perfect, thanks!

h2zero (approved):
👍

1technophile:
Thanks!

PR #1287, opened 2022-10-16
Spi connectivity check for c1101

crasu:
## Description:

I had messed up the wiring from the c1101 to my esp. This will do a basic connectivity check from the esp to the c1101 and print out a warning/success message.

### DIFF
--- a/main/ZgatewayRF.ino
+++ b/main/ZgatewayRF.ino
@@ -111,6 +111,12 @@ void setupRF() {
   Log.notice(F("RF_EMITTER_GPIO: %d " CR), RF_EMITTER_GPIO);
   Log.notice(F("RF_RECEIVER_GPIO: %d " CR), RF_RECEIVER_GPIO);
 #  ifdef ZradioCC1101 //receiving with CC1101
+  if (ELECHOUSE_cc1101.getCC1101()) {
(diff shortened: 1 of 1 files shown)


main/ZgatewayRF.ino
```diff
@@ -111,6 +111,12 @@ void setupRF() {
   Log.notice(F("RF_EMITTER_GPIO: %d " CR), RF_EMITTER_GPIO);
   Log.notice(F("RF_RECEIVER_GPIO: %d " CR), RF_RECEIVER_GPIO);
 #  ifdef ZradioCC1101 //receiving with CC1101
+  if (ELECHOUSE_cc1101.getCC1101()) {
+    Serial.println("C1101 spi Connection OK");
```
1technophile: Could you use `Log.notice` and `Log.error` instead, please?

crasu: Obsolete due to change to radiolib

crasu:
Lib is still used for rf gateway

1technophile:
Thanks

rds76:
You should use Flash strings for static Logs and end them with CR
`Log.notice(F("C1101 spi Connection OK" CR));`

1technophile:
Good catch

PR #1307, opened 2022-10-30
Handle OTA actively when MQTT is not connected

### LINKED ISSUE #1292: OMG is not responding to OTA invite when MQTT configuration is invalid
Before submitting a problem please check the troubleshooting section
https://docs.openmqttgateway.com/upload/troubleshoot.html

**Describe the bug**

I just flashed my ESP32 with OpenMQTTGateway with correct (manual config) WiFi but invalid (wrong IP) MQTT configuration.

`python3 espota.py         --ip=192.168.99.99         --port=8266         --host_port=10001         --auth=OTAPASSWORD         --file=${firmware}_firmware.bin         --debug         --progress`  gives the following error:

```
[INFO]: Starting on 0.0.0.0:10001
[INFO]: Upload size: 1333936
[INFO]: Sending invitation to: 192.168.99.99
[ERROR]: No Answer
```

**To Reproduce**

See above
**Expected behavior**

OTA update should be possible as long as WiFi is up.

**Screenshots**

* OpenMQTTGateway version used (V0.9.3, 0.8, development)
* Library version related to the problem you have (if you have troubles with RF provide the version of RCSwitch library)
* For IR and RF clarify if you tested with the basic examples given with these libraries

development version

**Additional context**

I suspect `connectMQTT` enters retry loop, without giving opportunity to call `ArduinoOTA.handle();` This is just speculation though, not confirmed

* You should not have a compilation error if you use the versions of the libraries linked into the libraries folder, this badges show you the state of the compilation 
[Build Status](https://github.com/1technophile/OpenMQTTGateway/actions)
* If you are not sure this is a bug or an enhancement post your question to the forum below
[Community forum](https://community.openmqttgateway.com)

ssalonen:
Resolves #1292

## Checklist:

### DIFF
--- a/main/main.ino
+++ b/main/main.ino
@@ -510,12 +510,16 @@ bool cmpToMainTopic(const char* topicOri, const char* toAdd) {
   return true;
 }
 
-void delayWithOTA(long millis) {
+void delayWithOTA(long waitMillis) {
+#if defined(ESP8266) || defined(ESP32)
   long waitStep = 100;
-  for (long wait = 0; wait < millis; wait += waitStep) {
+  for (long waitedMillis = 0; waitedMillis < waitMillis; waitedMillis += waitStep) {
     ArduinoOTA.handle();
     delay(waitStep);
   }
+#else
+  delay(waitMillis);
+#endif
 }
 
 void connectMQTT() {
--- a/main/main.ino
+++ b/main/main.ino
@@ -188,6 +188,7 @@ bool disc = true; // Auto discovery with Home Assistant convention
 #endif
 unsigned long timer_led_measures = 0;
 static void* eClient = nullptr;
+static unsigned long last_ota_activity_millis = 0;
 #if defined(ESP8266) || defined(ESP32)
 static bool mqtt_secure = MQTT_SECURE_DEFAULT;
 static uint8_t mqtt_ss_index = MQTT_SECURE_SELF_SIGNED_INDEX_DEFAULT;
@@ -509,6 +510,14 @@ bool cmpToMainTopic(const char* topicOri, const char* toAdd) {
   return true;
 }
 
+void delayWithOTA(long millis) {  
+  long waitStep = 100;
+  for(long wait=0; wait < millis; wait += waitStep){
+    ArduinoOTA.handle();
+    delay(waitStep);
+  }    
+}
+
 void connectMQTT() {
 #ifndef ESPWifiManualSetup
 #  if defined(ESP8266) || defined(ESP32)
@@ -564,10 +573,17 @@ void connectMQTT() {
       Log.warning(F("failed, ssl error code=%d" CR), ((WiFiClientSecure*)eClient)->getLastSSLError());
 #endif
     digitalWrite(LED_ERROR, LED_ERROR_ON);
-    delay(2000);
+    delayWithOTA(2000);
     digitalWrite(LED_ERROR, !LED_ERROR_ON);
-    delay(5000);
+    delayWithOTA(5000);
     if (failure_number_mqtt > maxRetryWatchDog) {
+      unsigned long millis_since_last_ota;
+      while((millis_since_last_ota != 0) && ((millis_since_last_ota = millis() - last_ota_activity_millis) < ota_timeout_millis)) {
+        // OTA might be still active, we sleep for a while
+        Log.warning(F("OTA might be still active (activity %d ms ago)" CR), millis_since_last_ota);
+        ArduinoOTA.handle();
+        delay(100);
+      }
       watchdogReboot(1);
     }
   }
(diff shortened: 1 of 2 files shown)


ssalonen:
Tested with ESP32, and it resolves the issue I was having with #1292.

Did not test with ESP8266, I would presume some sort of `yield()` might be needed there.

1technophile:
Good idea, so it enables to have OTA if we are only connected to Wifi isn't it ?

ssalonen:
Yes, let's say the system is connected via WiFi but not to mqtt (e.g. Wrong mqtt configured). With this patch, OTA updates still work

main/main.ino
```diff
@@ -564,10 +577,19 @@ void connectMQTT() {
...
+    delayWithOTA(5000);
     if (failure_number_mqtt > maxRetryWatchDog) {
+      unsigned long millis_since_last_ota;
+      while ((millis_since_last_ota != 0) && ((millis_since_last_ota = millis() - last_ota_activity_millis) < ota_timeout_millis)) {
+        // OTA might be still active, we sleep for a while
+        Log.warning(F("OTA might be still active (activity %d ms ago)" CR), millis_since_last_ota);
+#if defined(ESP8266) || defined(ESP32)
+        ArduinoOTA.handle();
+#endif
+        delay(100);
```
1technophile, line 597: Are you sure we need a `delay()` here ?

ssalonen: Well, the delay makes the waiting more CPU efficient? There is also that logging on every iteration of the loop currently...

Alternatively we could refactor the code here to log once, and then call delayWithOTA with the "remaining time". Naturally there would be still that delay, inside delayWithOTA.

```

while ((millis_since_last_ota != 0) && ((millis_since_last_ota = millis() - last_ota_activity_millis) < ota_timeout_millis)) {
  Log.warning(F("OTA might be still active (activity %d ms ago)" CR), millis_since_last_ota);
  unsigned long remaining = ota_timeout_millis - millis_since_last_ota;
// TODO: ensure delayWithOTA works properly with non-round numbers
  delayWithOTA(remaining);

}

```

ssalonen: One thing I noted, the while clause in the PR was supposed to be **(EDIT: corrected in https://github.com/1technophile/OpenMQTTGateway/pull/1307/commits/6a1b2b04867817d4627b4ceac549d2a0714c26db )**

```
while ((last_ota_activity_millis != 0) && ((millis_since_last_ota = millis() - last_ota_activity_millis) < ota_timeout_millis)) {

```

That is, break immediately/sooner when last_ota_activity_millis is zero. The zero value signifies no OTA or OTA that has finished.

1technophile: > Well, the delay makes the waiting more CPU efficient? There is also that logging on every iteration of the loop currently...

I was asking myself if this could not cause issue with `ArduinoOTA.handle();`, as I'm not seeing a `delay()` in there examples

ssalonen: Hmm, do you have a link to an example? We normally would not need any `delay()` whenever `loop()` is concerned. 

However, here I wanted to keep the changes minimal and avoid yielding control to delay. Instead, as we have retried all MQTT already, it's time to reboot (as long as there is no OTA in progress). 

I can remove the delay here and in `delayWithOTA` as well, it should not be too bad as this is an very exceptional code path, should not happen often, and thus CPU effieciency is not everything.

Edit: fwiw, the 100 ms delay seemed to work for uploading according to my tests

1technophile: I'm referring to the examples delivered with the [ArduinoOTA library ](https://github.com/espressif/arduino-esp32/tree/master/libraries/ArduinoOTA/examples/BasicOTA)

You tested it with which kind of configuration/gateways?

ssalonen: Right yeah, well in that ArduinoOTA library example the logic is in the loop. A bit trickier here IMO.

Configuration I used for tests is here: https://github.com/ssalonen/OpenMQTTGateway/blob/lerbacka/ble_scanner_cont_openhab_env.ini . ESP32 dev board

There's been some slight changes after my test. If you like, I can try it one more time to be sure.

ssalonen: @1technophile  tested with https://github.com/1technophile/OpenMQTTGateway/pull/1307/commits/6a1b2b04867817d4627b4ceac549d2a0714c26db included [my branch which is rebased on this one](https://github.com/ssalonen/OpenMQTTGateway/blob/lerbacka/ble_scanner_cont_openhab_env.ini), seems to work very reliably.

1technophile: Cool, I will test and merge if ok

1technophile: Working good!

1technophile:
Thanks

PR #1320, opened 2022-11-16
Add RGB Led management for errors, infos, and send receive events

1technophile:
## Description:
Add RGB led management for indicators, integrate it for M5 Atom and ESP32 C3 and S3 dev board Update Fastled version to 3.5.0.

The default behavior is with regular LED, to use RGB Led the macro RGB_INDICATORS needs to be define, along with the `FASTLED_IND_TYPE, FASTLED_IND_DATA_GPIO and FASTLED_IND_NUM_LEDS`
Note that some RGB led need to have a different color scheme, so in this case it can be entered along with the data PIN definition: `-DFASTLED_IND_DATA_GPIO=27,GRB`

## Checklist:


ozett:
why only for the m5atom? why not make LED-PINs signalage available on all esp32 boards?

1technophile:
This PR scope is RGB Led. This is why it impacts the boards with RGB Led only.

ozett:
fastled supports also the most common "neopixel"=ws2811/ws2812/ws2812B
https://github.com/FastLED/FastLED/wiki/Chipset-reference

why not make this on GPIOS available/compile on all normal ESP devices? 
looks not like a big-deal?

1technophile:
> why not make this on GPIOS available/compile on all normal ESP devices?
looks not like a big-deal?

When we don't know if the board has an RGB led we don't presume it by adding the library and the module to an hypothetical pin.

We could have a Tasmota like approach with a binary embedding more functions than a board needs but this is not currently implemented.

ozett:
dont know if the binary builds are done automatically,
but building an [all-sensors binary](https://github.com/arendst/Tasmota/pull/9568#issuecomment-1364513201) for the webinstaller could make this "tasmota"-like  (for risk-friendley people?) 😄 


image

1technophile:
We would need to enable pin change and module activation/deactivation at runtime level to do this.

kxn25:
I have connectet the last cable (in my case the white) to 3.3v no LED!
In `user.config` the AS608 is uncommentet.
I have uncommentet this in  `xsns_79_as608.ino`

Zwischenablage01
Still no LED!

1technophile:
This is not the OpenMQTTGateway code. I don't think you are in the right repository.

ozett:
>  This is not the OpenMQTTGateway code. I don't think you are in the right repository.



is this for me? 
my post before should illustrate that tasmota has an all-binary (for people who want to experiment and risk all consequences).
i would love to see that on the openmqttgateway web-installer too. thats why i posted that images. only as an example.
and only, if the builds were done by some automatismen and someone is willing to check all modules for an "all-sensors" build.

otherwise, if the hint was meant to someone else, my writings could be ignored

1technophile:
> otherwise, if the hint was meant to someone else, my writings could be ignored

It was for a post that has been removed ;-)

PR #1327, opened 2022-11-23
Support LM75 temperature sensor

davidfurey:
## Description:

Adds support for LM75 compatible sensors.  Mostly copy+paste from the existing HTU21 sensor. I have tested this with a ESP8266 and NCT75 temperature sensor and it publishes sensible temperature values.

## Checklist:


1technophile:
Thanks for the PR, could add the sensor into the `esp32dev-all-test` and `nodemcuv2-all-test`, so that it will be tested with the  continuous integration.

1technophile:
Thanks

PR #1331, opened 2022-11-28
Wifi Manager AP name and password extracted from MAC

1technophile:
## Description:
Set the AP name with the short name plus the 4 first digits of the ESP mac
to don't disclose the password used for WM.
Also change the wm password from MAc to be the last 8 digits, as these will change more than the first 8 ones. Also make consistent the ifndef qnd the parsing between WM_PWD_FROM_MAC and USE_MAC_AS_GATEWAY_NAME

## Checklist:

### DIFF
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -50,7 +50,7 @@
  */
 /*-------------DEFINE GATEWAY NAME BELOW IT CAN ALSO BE DEFINED IN platformio.ini----------------*/
 
-// Uncomment to use the MAC address in the format of 112233445566 as the gateway name
+// Uncomment to use the MAC address first 4 digits in the format of 5566 as the gateway name
 // Any definition of Gateway_Name will be ignored. The Gateway_Short_name _ MAC will be used as the access point name.
 //#define USE_MAC_AS_GATEWAY_NAME
 #ifndef Gateway_Name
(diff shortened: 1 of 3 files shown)


NorthernMan54:
Just updated a couple of my test boards, and identified a potentially un-intended side effect of this change.  When use `USE_MAC_AS_GATEWAY_NAME` the OTA hostname is also the first 4 digits of ESP Mac, which is not unique.

```
N: OTA Hostname: OMG_4C75.local
```

My recommendation would be to revert the OTA hostname back to earlier behaviour, which was the full ESP Mac, but I did not consider the original use case this change was designed to resolve.

NorthernMan54:
Replacing this line

```
strcpy(ota_hostname, WifiManager_ssid);
```

With 
```
snprintf(ota_hostname, MAC_NAME_MAX_LEN, "%s_%s", Gateway_Short_Name, gateway_name);
```

Will restore the original behaviour

1technophile:
I have considered this, but in this case, the ESP advertises the MAC address on the network, which also means the WifiManager Password when using `WM_PWD_FROM_MAC` will be available easily.
I was trying to find another board identifier and found that we could use the flash id, but didn't have time to implement it.

PR #1343, opened 2022-12-04
SECURITY FIX Workaround unsecure AP bug on restart.

h2zero:
## Checklist:


1technophile:
Thanks,

This is a workaround to https://github.com/espressif/arduino-esp32/issues/5038

Impact, when the board is in AP mode to display the wifi manager portal, if you wait more than 2 mins (wifi manager time out), a restart will occur, and the ESP32 may generate the Access Point without requesting the password.

PR #1344, opened 2022-12-05
Add BMP280 support, compiler directive support for oversampling

melyux:
## Description:
The BMP280 is a sister sensor to the BME280. Unlike the BME280, it doesn't measure humidity but is 33% more accurate than it for pressure. Technically, OMG already supports BMP280 with the same BME280 library ([added 2018](https://github.com/sparkfun/SparkFun_BME280_Arduino_Library/commit/6c239570ac816b5559c870fd68bbf5317fc2413a)) and it worked with OMG as it is, but this updates the documentation and comments and removes the error message when a BMP280 is recognized instead of a BME280.

I also add compiler directives for temperature/humidity/pressure oversampling settings for the BME280/BMP280. It's a minor change that mirrors the existing temperature correction, and the comments for it will mention the BMP280, so I included this in this PR too. Let me know if that's ok.

Some minor spelling and grammar fixes in the comments.

I don't know how to edit/suggest things to the Compatibility website, but we could add the [BMP-280 sensor I'm using](https://www.amazon.com/gp/product/B07VNDZ6N4/ref=ppx_yo_dt_b_search_asin_title?ie=UTF8&psc=1) to it.

## Checklist:

### DIFF
--- a/docs/prerequisites/board.md
+++ b/docs/prerequisites/board.md
@@ -3,7 +3,7 @@ OpenMQTTGateway is not closed to one board or type of board, by using the power 
 
 You can take a look to the [OpenMQTTGateway compatible website](https://compatible.openmqttgateway.com) to have a view of the [supported boards](https://compatible.openmqttgateway.com/index.php/boards/).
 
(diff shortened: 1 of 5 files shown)


1technophile:
Thanks for this, I will add the sensor to the compatibility website

PR #1364, opened 2022-12-15
[HA Discovery] Save the BLE settings and change switch to optimistic false

1technophile:
## Description:
The current way of saving settings is through the retain flag of the MQTT commands, unfortunately if several commands are issues they may overwrite each other. So as to prevent it, we now leverage the memory saving introduced earlier to store the BLE settings. Also leverage the config status publication to show the state through the button position (optimistic false)

image

This could be breaking for people that were using before `/commands/BTtoMQTT/config` so as to retrieve the status of the BT configuration.

## Checklist:

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -191,7 +191,7 @@ void BTConfig_fromJson(JsonObject& BTdata, bool startup = false) {
     Serial.println();
     return; // Do not try to erase/write/send config at startup
   }
-  pub("/commands/BTtoMQTT/config", jo);
+  pub(subjectBTtoMQTT, jo);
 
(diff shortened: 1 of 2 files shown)


1technophile:
@BadWolf42 this is changing the topic on which the BT config dump was published for information. I think it is better to have it there as this dump is more status information rather than a command. Let me know if any comments.

BadWolf42:
Hello @1technophile, 
I do not understand all the modifications you made, but I agree with the purpose of this change if I got it right: 
"Configuration information should not be published on the same topic than the one used to change configuration".
I did not change the config dump on my initial commit as was too much change at once, gade you made it :)
Bad

1technophile:
@BadWolf42, the modifications enable to save the BLE settings each time a switch or a change of configuration is triggered in HA

Issue #1380, opened 2023-01-07
CC1101 -> Support for ELECHOUSE_cc1101.setPA();


Dattel:
**Is your feature request related to a problem? Please describe.**
I'm wondering, if you can implement the ability to define the TX-Power for sending RF433 with the ELECHOUSE and the CC1101 with OpenMQTTGateway. 

Currently i have the problem, that some of my RF-Signals are too powerful and disturbing my neighbours power-sockets.. I'm switching them too with my commands :-D

The ELECHouse-Gateway already seems to support changing the TX-Power using the setPA(); function

```
ELECHOUSE_cc1101.setPA(10);  
// Set TxPower. The following settings are possible depending on the frequency band.  (-30  -20  -15  -10  -6    0    5    7    10   11   12) Default is max!
```

**Describe the solution you'd like**
I think of a compiler switch to define the default TX power (if not set, always use the maximum)
'-DZradioCC1101Power=12'

The second change would be a new optional parameter in the payload sended to the MQTTTo433 topic to change the power on a single 433 transmission

`{"value": 9084946,	  "protocol": 4,	  "length": 24,	  "delay": 350,	  "repeat":5, "CC1101_PA":10 }`

That would be the maximum flexible solution.
I'm interessted in your thoughts..
Thanks

DigiH:
Hi @Dattel, while I understand your request, and someone else will answer about this possibility, I had exactly the same situation here with my neighbour. Depending on which power-sockets the neighbour and you are using it might also be possible to change their receiving values - in our case with dip switches - to avoid having the conflicting commands. If they are however self-learning and/or multi-receiving value versions this will not work.

Dattel:
Hi @DigiH
Sadly not.... 
I'm using sockets from Lidl, and he is using intertec sockets.
Both remotes are looking the same. Buttons for A to D and 2 global Power on/off. There is no way to change frequency.

My remote sends rolling codes -> each key sends 4 different codes alternately. 
For openmqtt i just picked ONE of these Code to send with and it works for me.

But the strange thing is that my D-Off-Key switches all of my neighbours sockets to off... regardless, if they are programmed to channel A,B,C or D... I already tried another D-Off code (from my list of four), but it doesn't help.

So i already implemented my idea from above in an quick and dirty way. 
Will test the threshold for me and hopefully can give a feedback on that...

Dattel:
publishing 
[...]

PR #1382, opened 2023-01-08
Small fix on BME280 Config & CC1101-TXPower Throttle

Dattel:
## Description:
see discussion on:
https://github.com/1technophile/OpenMQTTGateway/issues/1380

Sorry, it's my first pull request and i mixed in another small change on BME280 configuration to make it possible to define BME Pins in platformio.ini

## Checklist:

### DIFF
--- a/docs/use/rf.md
+++ b/docs/use/rf.md
@@ -74,16 +74,19 @@ Arduino IDE serial data received when receiving data by 433Mhz
 
 ### Send data by MQTT with advanced RF parameters
 
-RF sending support three advanced parameters; bits length, RF protocol and RF pulselength
-if you want to use a different RCswitch protocol put inside your payload the protocol number 2, "protocol":2.
+RF sending support four advanced parameters: bits length, RF protocol, RF pulselength and RF transmission power
 
-if you want to use a pulselength 315 put inside your topic "delay":315
+-if you want to use a bits number different than 24 put inside your topic "length":24 for example
 
-if you want to use a bits number different than 24 put inside your topic "length":24 for example
+-if you want to use a different RCswitch protocol put inside your payload the protocol number 2, "protocol":2.
+
+-if you want to use a pulselength 315 put inside your topic "delay":315
+
+-if you want to send with a lower power (scale from -30 to 12; 12 is default) than put inside your topic "cc1101_pa":5
 
 Example:
-`mosquitto_pub -t "home/OpenMQTTGateway/commands/MQTTto433" -m '{"value":1315156,"protocol":2,"length":24,"delay":315}'`
-will make RCSwitch use the protocol 2 with a pulselength of 315ms and a bits number of 24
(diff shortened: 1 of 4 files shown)


NorthernMan54:
@Dattel Tks for this pull request, it looks really good

Can you make the following tweaks and update the pull request

1 - Pls fix the lint issue reported by the automated checks

2 - Can you add the new parameter to the documentation for the RF Module.  The documentation is located here https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/use/rf.md#rcswitch-based-gateway and is part of the repository, so please include the update as part of the pull request.

3 - I watching the conversation in #1380, and do we need the feature ?

Tks

Dattel:
Hi @NorthernMan54,
it's up to you to decide, if that's a usefull feature to use in OmqttG.
From my side it's a suggestion, since i was recently confronted with the problem. I like the flexibility to define a power-value for each sended RF-Code since only 2 key's of my remote seems to collide with "foreign" Devices :-D

I will have a look on the lint-error's but i have no clue, whats's really the problem. I guess the indent was not right?

NorthernMan54:
To find the lint issues, I go to the checks section of the pull request and look at the details for the failed check.  The lint check is pretty good as it shows you the fix.

In regards to the CC1101-TXPower Throttle feature, if you or @DigiH were going to make use of the feature then lets add it, but adding a feature/code that is never going to be used turns into a future code maintenance issue.

DigiH:
> In regards to the CC1101-TXPower Throttle feature, if you or @DigiH were going to make use of the feature then lets add it, but adding a feature/code that is never going to be used turns into a future code maintenance issue.

I'm afraid I'm on the give me a signal as strong as possible side - so this would not be useful for me personally.

Dattel:
Okay, i'm with you "strongest TX" will match almost 99% of all usecases. That's the reason for the paramateter to be as much optional as possible. And i'm also with you that every rare used feature is a possible source of error in case of refactoring... 

So in case, these implementation is declined, i would make me a stash to apply on every new version because i already use it to avoid neighborhood disputes :-D
But the argument "I don't use it, so we don't adapt it" would be a very weak reason - then you could certainly question some other features as well :-D

DigiH:
:thumbsup: Thanks @Dattel. 

Could you still add the documentation about the new parameter, as requested above by @NorthernMan54? 
Then this will be ready for merging.

DigiH:
Sorry for being awkward here, but since this is a CC1101 only feature could you move the information down a bit to the section
**Set Transmit and Receive Frequency of CC1101 Transceiver Module**
rename it to
**Set Transmit and Receive Frequency and Transmit Power of CC1101 Transceiver**
and include your information there.
Currently, apart from the actual key name, it is included in the general RF options also applicable to other transmitters.

Dattel:
you are right :-D

DigiH:
Looking good, thanks. 

Congratulations on your first PR :tada:

PR #1458, opened 2023-02-14
Added WiPhone LoRa support

xopr:
## Description:
This PR enables decoding [WiPhone](https://www.wiphone.io/) LoRa packets.
WiPhone uses the RadioHead library which adds some extra bytes to the header, but is currently unused (as per default `0xFFFF0000`).

The format is as followed:
```
  uint8_t rh_to; // currently 0xff
  uint8_t rh_from; // currently 0xff
  uint8_t rh_id; // currently 0x00
  uint8_t rh_flags; // currently 0x00
  uint16_t magic; // always 0x6c6d
  uint32_t to; // ESP ChipID or 0x00000000 for "broadcast"
  uint32_t from; // ESP ChipID
  char message[230];
```
I modified the LoRa to support more devices in the future, but I would like to see a mechanism similar as used in the [rtl_433_ESP](https://github.com/NorthernMan54/rtl_433_ESP#v0.1.6) library;
Then again, this library does support the `sx127x` chipset (that also does LoRa) already, it only doesn't support the RFM9xW.

Sending a message would be as simple as:
`mosquitto_pub -h 127.0.0.1 -t "home/OpenMQTTGateway_ESP32_LORA/commands/MQTTtoLORA" -m '{"type":"WiPhone","from":"C0FFEE","to":"0","message":"Hello from OMG"}'`

Incoming messages will look like:
`home/OpenMQTTGateway_ESP32_LORA/LORAtoMQTT {"rssi":-113,"snr":5.5,"pferror":18748,"packetSize":34,"from":"123456","to":"000000","message":"hi from wiphone","type":"WiPhone"}`

## Checklist:

## Notes:
To be honest, I'm not sure if people want to have the LoRa library extended to support decoding different payloads, but I thought it was the right thing to do (I came across this library which had all the fantastic features I needed, except LoRa payload decode support).

Also:
* When non-ascii data arrives, it converts to hex and will put the data in `LORAdata["hex"]` (and the other way around)
* I used the conversion methods from `Taken from main/ZgatewaySRFB.ino`, but it might be better to have a helper class that one could use to do these conversions.
* I also added invertIQ support for downstream/multiple gateway support

Let me know what you think.
Cheers.

### DIFF
--- a/main/ZgatewayLORA.ino
+++ b/main/ZgatewayLORA.ino
@@ -56,32 +56,6 @@ enum LORA_ID_NUM {
 typedef enum LORA_ID_NUM LORA_ID_NUM;
 
 /*
-Taken from main/ZgatewaySRFB.ino
-From an hexa char array ("A220EE...") to a byte array (half the size)
- */
-bool _hexToRaw(const char* in, byte* out, int rawSize) {
-  if (strlen(in) != rawSize * 2)
-    return false;
-  char tmp[3] = {0};
-  for (unsigned char p = 0; p < rawSize; p++) {
-    memcpy(tmp, &in[p * 2], 2);
-    out[p] = strtol(tmp, NULL, 16);
-  }
-  return true;
-}
-
-/*
-Taken from main/ZgatewaySRFB.ino
-From a byte array to an hexa char array ("A220EE...", double the size)
- */
-bool _rawToHex(byte* in, char* out, int rawSize) {
-  for (unsigned char p = 0; p < rawSize; p++) {
-    sprintf_P(&out[p * 2], PSTR("%02X" CR), in[p]);
-  }
-  return true;
-}
-
-/*
 Try and determine device given the payload
  */
 uint8_t _determineDevice(byte* packet, int packetSize) {
--- a/main/ZgatewayLORA.ino
+++ b/main/ZgatewayLORA.ino
@@ -33,6 +33,124 @@
 #  include <SPI.h>
 #  include <Wire.h>
 
+#define WIPHONE_MESSAGE_MAGIC 0x6c6d
+#define WIPHONE_MESSAGE_MIN_LEN sizeof(wiphone_message) - WIPHONE_MAX_MESSAGE_LEN
+#define WIPHONE_MAX_MESSAGE_LEN 230
+
+typedef struct __attribute__((packed)) {
+  // WiPhone uses RadioHead library which has additional (unused) headers
+  uint8_t rh_to;
+  uint8_t rh_from;
+  uint8_t rh_id;
+  uint8_t rh_flags;
+  uint16_t magic;
+  uint32_t to;
+  uint32_t from;
+  char     message[WIPHONE_MAX_MESSAGE_LEN];
+}
+wiphone_message;
+
+enum LORA_ID_NUM
+{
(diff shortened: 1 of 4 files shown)


1technophile:
That's great. Thanks for pushing this. I did not know about the WiPhone, it seems very interesting.

xopr:
Oops, I see that I messed up on the linting part.
I have to admit that PlatformIO support on VSCodium (which I use) is a bit flaky.

Is there something I could do? I noticed that it seems to reformat the code but the check still failed.

By the way, WiPhone also has a GitHub repo, but their firmware source is still in the process of being developed and prepared for a first public release: https://github.com/ESP32-WiPhone

1technophile:
> Is there something I could do? I noticed that it seems to reformat the code but the check still failed.

You should be able to format automatically by a right click on the document from VSC, and click "Format document"

xopr:
> You should be able to format automatically by a right click on the document from VSC, and click "Format document"

Thanks, that worked better than I'd expect from my setup :)

1technophile:
@kamaradclimber would you be interested in reviewing this PR as you were the previous commiter of this module

xopr:
@kamaradclimber let me know if you have any concerns regarding this PR; I don't want to break or hijack your code :)

I'm currently using it as a hub for my WiPhone and also monitoring binary messages in my neighbourhood.
I can probably poke a friend for testing the new features as well.

Thanks / merci

kamaradclimber:
I'm fine, if any issue I'll fix them on my side ;)

xopr:
@1technophile is this PR something you could approve / merge?

main/ZgatewayLORA.ino
```diff
@@ -33,6 +33,122 @@
...
+/*
+Taken from main/ZgatewaySRFB.ino
+From a byte array to an hexa char array ("A220EE...", double the size)
+ */
+bool _rawToHex(byte* in, char* out, int rawSize) {
+  for (unsigned char p = 0; p < rawSize; p++) {
+    sprintf_P(&out[p * 2], PSTR("%02X" CR), in[p]);
+  }
+  return true;
+}
```
1technophile: If used into ZgatewaySFRB.ino, could you add the function to main and remove it from here and from ZgatewaySRFB.ino

main/ZgatewayLORA.ino
```diff
@@ -69,16 +185,38 @@ void LORAtoMQTT() {
...
-    packet = "";
+    // Create packet and reserve null terminator space
+    byte packet[packetSize + 1];
+    boolean binary = false;
     for (int i = 0; i < packetSize; i++) {
-      packet += (char)LoRa.read();
+      packet[i] = (char)LoRa.read();
+
+      if (packet[i] < 32 || packet[i] > 127)
+        binary = true;
```
1technophile, line 195: I'm not getting this criterion about identifying a binary depending on the <32 or >127 could you explain ?

main/ZgatewayLORA.ino
xopr: I was a bit hesitant to change at several places, but you're right: I shouldn't add redundant code; fixed it :)

1technophile:
Thanks

xopr:
Thanks for the quick response!

PR #1485, opened 2023-02-24
[OTA] Enable 2 clicks OTA update / one simple MQTT command

1technophile:
## Description:
This feature will scan at the start if an update is available, if yes an MQTT command can be triggered to launch the update
For the controllers using Home Assistant convention, the option to update will be presented, and the update will be done through a 2 clicks sequence
image

## Checklist:


main/main.ino
```diff
@@ -1940,38 +1942,96 @@ void receivingMQTT(char* topicOri, char* datacallback) {
...
+    Log.trace(F("Payload %s" CR), payload.c_str());
+  } else {
+    Log.error(F("Error on HTTP request"));
+  }
+  http.end(); //Free the resources
+  if (jsondata.containsKey("latest_version")) {
+    jsondata["installed_version"] = OMG_VERSION;
+    jsondata["entity_picture"] = ENTITY_PICTURE;
+    latestVersion = jsondata["latest_version"].as<String>();
+    pub(subjectSYStoMQTT, jsondata);
```
NorthernMan54, line 1980: Should this be a separate topic ?   As the message is lost/over written with further updates.

1technophile: It is enough to be caught by HA, but for the sake of readiness I can understand the argument for a separate topic

PR #1512, opened 2023-03-08
Add environment for board Wemos Lolin C3 mini

Argafal:
## Description: New environment definition for board Wemos Lolin C3 mini v2.3.0

This is the result of a discussion about how to run OMG on this board. See https://community.openmqttgateway.com/t/watchdog-kills-omg-on-esp32-c3-within-seconds/2192/36

## Checklist:

### DIFF
--- a/environments.ini
+++ b/environments.ini
@@ -1578,7 +1578,7 @@ extends = env:lolin_c3_mini
 build_flags =
   ${env:lolin_c3_mini.build_flags}
   '-DLOG_LEVEL=LOG_LEVEL_NOTICE' ;default
-  '-DWM_DEBUG_LEVEL=DEBUG_NOTIFY' ;default
+  '-DWM_DEBUG_LEVEL=1' ;DEBUG_NOTIFY=1, default value
 ; Serial output via USB
   '-DCONFIG_ESP_CONSOLE_USB_SERIAL_JTAG=true'
   '-DCONFIG_ESP_CONSOLE_UART=ESP_CONSOLE_USB_SERIAL_JTAG'   
--- a/environments.ini
+++ b/environments.ini
@@ -1564,17 +1564,20 @@ build_flags =
   '-DTRIGGER_GPIO=9'
   '-DNO_INT_TEMP_READING=true' ; No internal temperature on ESP32 C3 or S3
   '-DGateway_Name="OpenMQTTGateway_ESP32C3_BLE"'
+; Serial output currently causes a delay on each message, if no serial listener is connected.
+; Therefore we disable serial output. Use the _with_serial environment if you
+; need serial output for debugging purposes.
+  '-DLOG_LEVEL=LOG_LEVEL_SILENT'  ; disable serial logging.
+  '-DWM_DEBUG_LEVEL=DEBUG_ERROR'  ; disable serial logging.
 custom_description = BLE gateway on ESP32C3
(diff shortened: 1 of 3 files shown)


1technophile:
@Argafal thanks for PR, do not try to fix the issue with ` 'MS_CDP' is not a member of 'TheengsDecoder::BLE_ID_NUM` we have a PR coming to fix it

environments.ini
```diff
@@ -1539,6 +1539,43 @@ build_flags =
...
+  '-DLED_ERROR=7'
+  '-DTRIGGER_GPIO=9'
+  '-DNO_INT_TEMP_READING=true' ; No internal temperature on ESP32 C3 or S3
+  '-DGateway_Name="OpenMQTTGateway_ESP32C3_BLE"'
+custom_description = BLE gateway on ESP32C3
+custom_hardware = Wemos Lolin C3 mini v2.3.0
+
+; Serial output currently causes a delay on each message, if no serial listener is connected.
+; Use the _no_serial environment unless you are actively debugging.
+[env:lolin_c3_mini_no_serial]
```
1technophile: @Argafal for the moment I would merge the no serial into the environment above to avoid users going into the same issues as you

Argafal: Fine with me. How about we do it the other way round: make no_serial the default, but also provide a target for debugging with the serial output enabled. I am still keen on understanding the random reboot...

What are the default log levels for those two parameters? -DWM_LOG_LEVEL=? and -DLOG_LEVEL=?

1technophile: Make no_serial the default is perfect

For `LOG_LEVEL=` if you don't define it, it would be NOTICE
https://github.com/Legion2/OpenMQTTGateway/blob/257b39da8136448ee10ed4e39e7b3d0c086f73a5/main/User_config.h#L487

and NOTIFY for WM
https://github.com/Legion2/OpenMQTTGateway/blob/257b39da8136448ee10ed4e39e7b3d0c086f73a5/main/User_config.h#L130

Argafal: Perfect. Thanks for pointing out where to find this part in the documentation as well. Maybe I can understand the random reboot better by increasing the log levels in my next test run.

New patch below.

environments.ini
```diff
@@ -1539,6 +1539,46 @@ build_flags =
...
+  '-DLED_INFO=6'
+  '-DLED_ERROR=7'
+  '-DTRIGGER_GPIO=9'
+  '-DNO_INT_TEMP_READING=true' ; No internal temperature on ESP32 C3 or S3
+  '-DGateway_Name="OpenMQTTGateway_ESP32C3_BLE"'
+; Serial output currently causes a delay on each message, if no serial listener is connected.
+; Therefore we disable serial output. Use the _with_serial environment if you
+; need serial output for debugging purposes.
+  '-DLOG_LEVEL=LOG_LEVEL_SILENT'  ; disable serial logging.
+  '-DWM_DEBUG_LEVEL=DEBUG_ERROR'  ; disable serial logging.
```
1technophile: You may try 0 instead of `DEBUG_ERROR`

Argafal: `#  define WM_DEBUG_LEVEL 1 // valid values are: DEBUG_ERROR = 0`

DEBUG_ERROR is the same as 0, no? But sure, i can change it to 0 instead. I will also update some comments so that the definition also matches the Seeed C3 board.

1technophile:
Thanks!

Issue #1404, opened 2023-01-19
[FR] Timestamp, Retain messages, Filter by name/prefix


ilgrank:
Hi
Again, thanks for OpenMQTTGateway!
I'm using it since some day on an ESP32 and.. it's great :)

I have some ideas to propose, that (IMHO of course) could be of interest:
- **Timestamp messages**: Since we configure an NTP anyways, it would nice to have the option to timestamp messages. E.g.: I have several types of sensors, some which report every few seconds and others that report every few hours: it would be useful to know ***when*** a message was received.. 
- **Retain messages:** The scenario above requires the ability to *retain* messages. I've read previous discussions questioning the usefulness of this, but I have a new argument to support my request: in my case, not having the option to retain messages messages means that any new client subscribing to my broker (I'm not using HA or any other sofware, just JS clients connecting to the broker) needs to wait for a sensor to advertise new data.. and it could be second or dozens of minutes. Having the last reading as a retained message would mean knowing immediately the latest known sensor reading.

I've already a work-around in place for the points above (another client that is subscribed to OMQTTG topics and re-publishes them with timestamps and retain flag), but I think it would be more practical to have these features built-in (possibly without re-compiling)

- **Device filtering**: There are (alas) dozens of BLE devices around me it seems. While I've filtered by MAC Address whitelist (thanks!) but it seems to me that adding a filtering by name/prefix would prove to be more flexible.

As always, thanks for your support!

1technophile:
Hello,

> Timestamp messages: Since we configure an NTP anyways, it would nice to have the option to timestamp messages. E.g.: I have several types of sensors, some which report every few seconds and others that report every few hours: it would be useful to know when a message was received..

This can be managed on the controller side, what do you think would be the added value of adding a timestamp at the firmware level versus adding it when the controller receives it?

ilgrank:
pardon me.. when you say "_on the controller side_", you mean on the devices connecting to the broker that subscribe to the sensors topics? (these are offline most of the time)
or you mean on the broker itself? if that's the case, I've talked some time ago with the Mosquitto devs, and they're against the broker altering the message payload in any way. (discussion [here](https://github.com/eclipse/mosquitto/issues/1225) )

(on a side note, there's a project similar in scope to OMQTTG, Zigbee2mqtt, has a _last_seen_ option that does what the name suggests.. just discovered that while I was searching for my FR to Mosquitto devs)

1technophile:
[...]

PR #1525, opened 2023-03-12
Add time (unix timestamp) to MQTTtoSYS and sensors topics

ilgrank:
## Description:
Added timestamping logic to Sensor readings, to provide the ability to know when a reading was made.
(Ad discussed [here](https://github.com/1technophile/OpenMQTTGateway/issues/1404)  )

It would be very useful to know *when* a sensor reported a value, since most sensors don't report very often and in a scenario without a controller, that means not knowing when a (retained) message has been published.

Also, adding a timestamp to MQTTtoSYS  too provides useful insight on the validity of the sensor timestamps.
## Checklist:


main/main.ino
```diff
@@ -1791,6 +1794,19 @@ void stateMeasures() {
 }
 #endif
 
+#if defined(ESP8266) || defined(ESP32)
+void SyncNTP() {
```
1technophile: Could you use this function below to avoid the block duplication Row 2110

ilgrank: Now using SyncNTP in place of the (mostly duplicated) block @ row 2110, however I was not 100% sure of the startProcessing() call, that's why I was originally just duplicating the code..

Tried in an ESP32 and it's working well so far (more than 6 days of uptime) but likely it is overkill to sync to the timeserver every 2 minutes.. (as I'm doing now).
Maybe the SyncNTP could be called.. say, every hour instead? (just a guess, as I did not check yet how reliable ESP32 timing is)

PR #1526, opened 2023-03-12
Fix for #1458: crash on sending LoRa data

xopr:
## Description:
This PR fixes the crash introduced in #1458 (I'm so sorry I introduced this).

Upon writing documentation for the new features, I discovered the device will crash when sending `message` or `hex` data.
It now verifies the presence of the `type` field, and also makes sure the correct size is used when sending in hex.

## Checklist:


1technophile:
Thanks

PR #1531, opened 2023-03-14
Added macros for ESP -  WifiGMode and WifiPower to reduce/eliminate Wifi interference on sensors like HC-SR501

diepeterpan:
## Description:

Added macros for ESP -  WifiGMode and WifiPower to reduce/eliminate Wifi interference on sensors like HC-SR501 and updated documentation.

https://community.openmqttgateway.com/t/wifi-interference-with-hc-sr501/2303

## Checklist:
  - [ X] The pull request is done against the latest development branch
  - [ X] Only one feature/fix was added per PR and the code change compiles without warnings
  - [ X] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).


1technophile:
Thanks for the PR, could you rebase and force push please

diepeterpan:
> Thanks for the PR, could you rebase and force push please

Apology for the dumb question, Rebase to?
image

diepeterpan:
Also don't see a [Update branch] here.
image

1technophile:
Oh, I apologize for the confusion! I thought the added environments were a result of a missing rebase. Just to share my perspective, I would rather not add production environments for specific sensor configurations as it may lead to a lot of unnecessary environments. However, I do agree that we still need to test any modifications that are made to ensure proper build functionality. 
As a solution, how about we keep one environment and add the suffix "-test" to differentiate it from the production environments and prevent it from adding a binary?

main/main.ino
```diff
@@ -970,6 +970,27 @@ bool wifi_reconnect_bypass() {
...
+    }
+//https://github.com/espressif/arduino-esp32/search?q=WIFI_PROTOCOL_11G
+//https://www.letscontrolit.com/forum/viewtopic.php?t=671&start=20
+#    ifdef WifiGMode
+    if (esp_wifi_set_protocol(WIFI_IF_STA, WIFI_PROTOCOL_11B | WIFI_PROTOCOL_11G) != ESP_OK) {
+      Log.error(F("Failed to change WifiMode." CR));
+    }
+#    endif
+
+#    ifdef WifiPower
```
1technophile: Could you add these macros into User_config.h (commented) with an explanation

main/main.ino
```diff
@@ -970,6 +970,27 @@ bool wifi_reconnect_bypass() {
   while (WiFi.waitForConnectResult() != WL_CONNECTED && wifi_autoreconnect_cnt < maxConnectionRetryWifi) {
 #  endif
     Log.notice(F("Attempting Wifi connection with saved AP: %d" CR), wifi_autoreconnect_cnt);
+
+    //Reduce WiFi interference when using ESP32 using custom WiFi mode and tx power
+#  ifdef ESP32
+    // https://github.com/jnogues/ESP32-Long-Range-WiFi/issues/4
+    // Reset wifi protocol
+    if (esp_wifi_set_protocol(WIFI_IF_STA, WIFI_PROTOCOL_11B | WIFI_PROTOCOL_11G | WIFI_PROTOCOL_11N | WIFI_PROTOCOL_LR) != ESP_OK) {
```
1technophile: The goal of this is to improve wifi reconnection ?

diepeterpan:
> Oh, I apologize for the confusion! I thought the added environments were a result of a missing rebase. Just to share my perspective, I would rather not add production environments for specific sensor configurations as it may lead to a lot of unnecessary environments. However, I do agree that we still need to test any modifications that are made to ensure proper build functionality. As a solution, how about we keep one environment and add the suffix "-test" to differentiate it from the production environments and prevent it from adding a binary?

Instead of "-test"; which might sound like we are not sure whether it works rather use the suffix "-custom" for custom working configuration without production binaries.

main/main.ino
diepeterpan: Done.

main/main.ino
diepeterpan: As I understand it when we start changing the Wifi band it gets stored in the ESP32 firmware and needs to be reset to default using explicit code.  Hence the research and reference to https://github.com/jnogues/ESP32-Long-Range-WiFi/issues/4 where other Platform.IO users had the same experience.

I could not understand why my ESP32 would not want to connect to Band N after lowering it to Band G even after removing my code. And this fixed it.

1technophile:
> Instead of "-test"; which might sound like we are not sure whether it works rather use the suffix "-custom" for custom working configuration without production binaries.

`-test` is the current prefix used by the scripts to differentiate the environments between the testing ones and the production ones
https://github.com/1technophile/OpenMQTTGateway/blob/3ceb19f7beb3531f933d8de250deae6a6d14fd98/scripts/prepare_deploy.sh#L20

Users can take samples from either production or test one to build their custom environment.

Also if you are not doing it already I encourage to use a separate file for custom environment like a [prod_env.ini](https://github.com/1technophile/OpenMQTTGateway/blob/development/prod_env.ini.example), this can host the custom configurations, stay local and will be ignored by git.

diepeterpan:
> > Instead of "-test"; which might sound like we are not sure whether it works rather use the suffix "-custom" for custom working configuration without production binaries.
> 
> `-test` is the current prefix used by the scripts to differentiate the environments between the testing ones and the production ones
> 
> https://github.com/1technophile/OpenMQTTGateway/blob/3ceb19f7beb3531f933d8de250deae6a6d14fd98/scripts/prepare_deploy.sh#L20
> 
> Users can take samples from either production or test one to build their custom environment.
> 
> Also if you are not doing it already I encourage to use a separate file for custom environment like a [prod_env.ini](https://github.com/1technophile/OpenMQTTGateway/blob/development/prod_env.ini.example), this can host the custom configurations, stay local and will be ignored by git.

Kewl, changed to -test ; will def. look into prod_env.ini. ; I was wondering how ones custom environment is maintained.

main/main.ino
1technophile: We will need testing of this with different router configurations, as I'm not sure of the impact of this function.

1technophile: I just tested it, and it prevents the wifi configuration portal from starting

diepeterpan: Busy testing a solution.

1technophile: If we could avoid going through this function for regular users we would save a bunch of regression testing and potential headache

diepeterpan: I got rid of that; from my testing, it should be working, incl. WiFi configuration.  I did an {"cmd":"erase"} and checked that I can connect and reconfigure.

diepeterpan: See whether the latest changes are more suitable; I have done quite a bit of testing and it seems fine.

1technophile: I'm not a fan of impacting the wifi connection for everybody. We had users struggling with their routers the last time we did that.
Can these changes be only for HC-SR501?

diepeterpan: The latest pre-processor statements prevent this code from impacting existing users.  Only when they **explicitly**  **DEFINE** or **uncomment** either **WifiGMode** or **WifiPower** and have an **ESP32** will the code be compiled which changes the WiFi protocol and transmit power.

1technophile: Great!

PR #1541, opened 2023-03-16
[HA Discovery] Auto Off timer

1technophile:
## Description:
Auto deactivate discovery after 30min from the last activation, goal is to avoid unwanted entities created

## Checklist:


NorthernMan54:
Two thumbs for this

PR #1548, opened 2023-03-17
[BT] Device tracker presence detection

1technophile:
## Description:
Add device tracker detection and automatic creation following HA discovery convention So as to trigger the away state add an offline status to the identified BLE trackers

## Checklist:


toomyem:
May I add a note here regarding this new feature?

I just tried it at home, and it looks like that for each device, more than one tracker is created. There are two in my case. One for 'bpm' and other for 'steps':

obraz

I don't think this is the expected behavior.

1technophile:
Good catch, let me fix that

DigiH:
@toomyem just for the compatibility records - would you mind sharing which Mi Band/Amazfit Watch model you have?

Thanks

toomyem:
No problem:

"name": "Amazfit Bip S"
"brand": "Xiaomi/Amazfit"
"model": "Mi Band/Smart Watch"
"model_id": "MB/SW"

toomyem:
Let me also add, that it would be nice if device's name was used for tracker name, instead of model_id.
This is a general remark. It would be nice to have all discovered devices named after their bt name rather than model_id.

1technophile:
@toomyem, here is the fix:
https://github.com/1technophile/OpenMQTTGateway/pull/1567

Regarding the name, the issue is that we do not always get the name from the advertisements, so for consistency stake, we choose to display the model instead

DigiH:
@toomyem 

Having caught a few **Amazfit Bip S** broadcasts in my area, I get the feeling that they don't match the **Mi Band/Amazfit Watch** decoder, and give incomplete and possibly wrong decoded information - like always activity heart rate, even if there is no activity going on, and far too high, near dangerous heart rates, and no servicedata/steps at all.

I think the **Bip** line can be differentiated from other Amazfit Watches/Bands, which adhere to the decoder, by their leading **570100** of the manufacturerdata, compared to  **570102** of the compatible models.

Would you be willing in helping to confirm this, and possibly creating a separate correct decoder for the **Bip** watches/bands?

Ideally in The Theengs Decoder discussions
https://github.com/theengs/decoder/discussions

toomyem:
Yes, I can take a look at this. I have also access to Amazfit GTS mini and Xiaomi Miband 5, so I can check them as well.

DigiH:
That would be great, thanks a lot.

I have already prepared a PR to restrict the current Mi Band/Amazfit decoder

https://github.com/theengs/decoder/pull/319

and think the Bip S and similar might need a separate decoder, as they seem to have a lot more information in their manufacturerdata, and possibly only use the manufacturerdata and no servicedata for their steps count as well.

Mi Band 5 and 6 are apparently still fine with the current decoder scheme, and hopefully the latest 7 as well.

PR #1550, opened 2023-03-18
[TEMT] Added Temt6000 light sensor using a d1 mini

diepeterpan:
## Description:
Added Temt6000 light sensor module using a d1 mini

image
image
image

## Checklist:

### DIFF
--- a/environments.ini
+++ b/environments.ini
@@ -1104,6 +1104,7 @@ build_flags =
   '-DZsensorBME280="BME280"'
   ;'-DZsensorHTU21="HTU21"'
   '-DZsensorAHTx0="AHTx0"'
+  '-DZsensorTEMT6000="TEMT6000"'
   '-DZsensorTSL2561="TSL2561"'
   '-DZsensorDHT="DHT"'
   '-DZsensorDS1820="DS1820"'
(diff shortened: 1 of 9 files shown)


environments.ini
```diff
@@ -1675,3 +1676,62 @@ build_flags =
...
+custom_hardware = Motion sensor first version
+
+[env:esp01-hcsr501-ms-ota-test]
+extends = env:esp01-hcsr501-ms
+upload_protocol = espota
+upload_port = 127.0.0.1
+upload_flags =
+  --auth=OTAPASSWORD
+  --port=8266
+
```
1technophile: Could you add ` '-DZsensorTEMT6000="TEMT6000"'` to `esp32dev-all-test` so that the continuous integration will check that it builds with ESP32?

We should not add three other environments. The users should store this kind of custom environment in a prod_env.ini
By adding the module into the `nodemcu-all-test` and the `esp32dev-all-test` we verify that the module builds correctly and can be used by changing the configuration of the gateway through Platformio.
I would like to avoid being in a situation on which we store all possible configuration in the web upload and make it difficult for other users to find the right binary.

diepeterpan: All good, done.

platformio.ini
```diff
@@ -101,6 +101,8 @@ extra_configs =
 ;default_envs = lolin_c3_mini
 ;default_envs = thingpulse-espgateway
 ;default_envs = thingpulse-espgateway-ethernet
+;default_envs = d1mini-temt6000-ls
```
1technophile: Just remove these 2 also and we will be good to go

diepeterpan: No problem, done.

1technophile:
Thanks

PR #1576, opened 2023-03-25
Add MQ2 Gas sensor & tweak DHT config to avoid editing source when PI…

diepeterpan:
## Description:

Add MQ2 Combustible Gas sensor & tweak DHT config to avoid editing source when PIN is different

image

This MQ-2 Methane Gas Sensor module detects a wide range of potentially dangerous gases, with an onboard potentiometer for adjusting sensitivity.

## Checklist:


1technophile:
Thanks for the PR.
For Lint, if you right-click into the code in VSC, you should have a "Format document" option. This will format the code directly according to the expectations.

diepeterpan:
> Thanks for the PR. For Lint, if you right-click into the code in VSC, you should have a "Format document" option. This will format the code directly according to the expectations.

Thanks, was not sure whether it will work, will make a note of it.

PR #1547, opened 2023-03-17
[SYS] Bump ESP32 platform to v6.1.0 and ESP8266 to v4.1.0

1technophile:
## Description:
Update ESP platforms

### Resolved Issues

* ZdisplaySSD1306 module when included causes a immediate reboot loop - #1551

### Open Issues

* Usage of GPIO pins greater than 39 causes excessive logging

```
[    41][E][esp32-hal-gpio.c:102] __pinMode(): Invalid pin selected
[    41][E][esp32-hal-gpio.c:102] __pinMode(): Invalid pin selected
E (40) gpio: gpio_set_level(226): GPIO output gpio_num error
E (45) gpio: gpio_set_level(226): GPIO output gpio_num error
E (51006) gpio: gpio_set_level(226): GPIO output gpio_num error
E (51012) gpio: gpio_set_level(226): GPIO output gpio_num error
E (51017) gpio: gpio_set_level(226): GPIO output gpio_num error
```
* Save configuration to flash triggers an exception/restart on a lily board

```
assert failed: heap_caps_free heap_caps.c:360 (heap != NULL && “free() target pointer is outside heap areas”)

Backtrace: 0x40083c69:0x3ffb2700 0x4008dcc1:0x3ffb2720 0x40093a45:0x3ffb2740 0x40084126:0x3ffb2870 0x40093a75:0x3ffb2890 0x40102e41:0x3ffb28b0 0x40102e51:0x3ffb28d0 0x400db272:0x3ffb28f0 0x400db3db:0x3ffb2c70 0x400db4f9:0x3ffb2cb0 0x400de686:0x3ffb2cd0 0x400dea63:0x3ffb3040 0x401a626d:0x3ffb30f0 0x400e641a:0x3ffb3110 0x400e64c1:0x3ffb3130 0x400e659e:0x3ffb3170 0x400e6751:0x3ffb31e0 0x400df386:0x3ffb3230 0x400df457:0x3ffb3250 0x401057e1:0x3ffb3290
  #0  0x40083c69:0x3ffb2700 in panic_abort at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/esp_system/panic.c:408
  #1  0x4008dcc1:0x3ffb2720 in esp_system_abort at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/esp_system/esp_system.c:137
[...]

PR #1580, opened 2023-03-27
Removed forced 5.3.0 platform for Lolin C3 boards

ilgrank:
## Description:
Forcing platform vesion @5.3.0 was a stopgap since the reference at the time was 5.2.0, which had issues. (the infamous `assert rwble.c 261` error)
Now it makes sense (IMHO of course) to go back to the default platform since those issues have been fixed and  [it should also be more stable](https://github.com/1technophile/OpenMQTTGateway/pull/1547#issuecomment-1481439786)

## Checklist:


1technophile:
Thanks

PR #1621, opened 2023-04-28
[OTA] Capability to disable Firmware Update on ESP32 WEB

bwynants:
## Description:

there is a compile error when '-UMQTT_HTTPS_FW_UPDATE'  is set in the build flags with an ESP32

## Checklist:

### DIFF
--- a/environments.ini
+++ b/environments.ini
@@ -116,6 +116,7 @@ build_flags =
   '-DDEEP_SLEEP_IN_US=120000000' ; Deep sleep for a period of time to conserve battery
   '-DESP32_EXT0_WAKE_PIN=GPIO_NUM_14' ;Should a water leak be detected immediatly wake the ESP32 and report the problem
   '-DESP32_EXT0_WAKE_PIN_STATE=0' ;Should a water leak be detected immediatly wake the ESP32 and report the problem, sensor state to look for wakeup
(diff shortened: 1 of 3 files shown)


1technophile:
Thanks for the PR,

Could you add the macro to this environment:
https://github.com/1technophile/OpenMQTTGateway/blob/4b8dfbea42488c464a8bdb3cc613c81a799fb4a8/environments.ini#L51
This way it will be tested by the CI

NorthernMan54:
@bwynants Question - Why do you have the MQTT_HTTPS_FW_UPDATE directive set ?  Should the scope of  MQTT_HTTPS_FW_UPDATE be changed to still enable the WebUI firmware upgrade use case ?

bwynants:
> @bwynants Question - Why do you have the MQTT_HTTPS_FW_UPDATE directive set ? Should the scope of MQTT_HTTPS_FW_UPDATE be changed to still enable the WebUI firmware upgrade use case ?

because i do not want that kind of updating? i upload my firmware updates via 'ota' saves me some space on the ESP

bwynants:
> This way it will be tested by the CI

but CI would not test updates via WEB UI anymore. do you want that?

1technophile:
> but CI would not test updates via WEB UI anymore. do you want that?

It will be tested by the other environments

1technophile:
Thanks

PR #1648, opened 2023-05-20
Compare time differences instead of timestamps

flo-x:
Compare time differences instead of timestamps

## Description:
Comparing two results of millis() can result in erroneous behavior when the value of millis() wraps around, which happens after around 25 days of uptime. Comparing time difference (by subtracting two timestamps) works well, provided timestamps are not too old (less than 25 days old).

## Checklist:
  - [ X] The pull request is done against the latest development branch
  - [ X] Only one feature/fix was added per PR and the code change compiles without warnings
  - [ X] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).

### DIFF
--- a/main/ZsensorADC.ino
+++ b/main/ZsensorADC.ino
@@ -34,6 +34,8 @@
 ADC_MODE(ADC_TOUT);
 #  endif
 
+static int persistedadc = -1024; // so that the first reading will always be published
+
 //Time used to wait for an interval before resending adc value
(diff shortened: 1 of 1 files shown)


1technophile:
Thanks for the PR, for code format, right click on the code from VSC, "Format document"

1technophile:
thanks

PR #1677, opened 2023-06-11
[SYS] Add timestamp for messages

dkneisz:
## Description:
- add timestamp when publishing
- add "GMT" as timezone for timestamp (e.g. for openHAB to use the correct time if you are in a different timezone)

## Checklist:

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -692,14 +692,6 @@ void procBLETask(void* pvParameters) {
       String mac_adress = advertisedDevice->getAddress().toString().c_str();
       mac_adress.toUpperCase();
       BLEdata["id"] = (char*)mac_adress.c_str();
-#  if defined(ESP8266) || defined(ESP32)
-#    if message_UTCtimestamp == true
-      BLEdata["UTCtime"] = UTCtimestamp();
-#    endif
-#    if message_unixtimestamp == true
-      BLEdata["unixtime"] = unixtimestamp();
-#    endif
-#  endif
       BLEdata["mac_type"] = advertisedDevice->getAddress().getType();
       BLEdata["adv_type"] = advertisedDevice->getAdvType();
       Log.notice(F("Device detected: %s" CR), (char*)mac_adress.c_str());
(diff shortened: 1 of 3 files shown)


1technophile:
Thanks for the PR. I think timestamp should be added in the joint pub function (from the main.ino) rather than adding it into each gateway.
This will be easier to maintain and will avoid code duplication.

dkneisz:
I agree. Since it was already implemented in ZgatewayBT.ino I was under the impression, that it is implemented in some other modules also. That's why I just implemented it for ZgatewayRTL_433.ino. Should I implement this here or create another pull request?

1technophile:
Up to you, I don't mind either solution

dkneisz:
I moved it now to main.ino into the function "void pub(const char* topicori, JsonObject& data)". There are also to additional functions
- void pub(const char* topicori, const char* payload, bool retainFlag)
- void pub(const char* topicori, const char* payload)
but they don't seem to do the same thing. I'm not quite sure where exactly they are used. They just publish the given payload to the given topic, so I didn't change anything there.

I tested this for the following:
- BTtoMQTT
- PilighttoMQTT
- RLStoMQTT
- SYStoMQTT
- WebUItoMQTT
- 433toMQTT
- RTL_433toMQTT
- SSD1306toMQTT

1technophile:
Thanks

DigiH:
@dkneisz @1technophile

While great to have the time stamp option for all modules, why has the naming here changed from the definitive Z(ulu Time) to an ambiguous, non-existant, ZGMT naming scheme?

What we do get is Zulu time, which is also known as Coordinated Universal Time/Temps Universel Coordonné - abbreviated nonsensically as UTC, so either the previously existing Z or alternatively UTC should be used, but definitely not GMT in any form, as GMT does observe daylight saving time, which for the Coordinated Universal (Zulu) Time we give is incorrect and unwanted for users worldwide. Just to clarify my above comment.

dkneisz:
@DigiH your are right. I was trying to fix an issue I had, that my openHAB used the wrong timezone for this date. Now that you mention it, I'm not sure why this solution worked for me. Even stranger I now reverted the change and my openHAB is now using my timezone. Not sure what happened here. I will revert this change and submit another pull request for this.

PR #1686, opened 2023-06-16
fix wrong time format for UTC timestamp

dkneisz:
## Description:
fixes wrong format of timestamp introduced with #1677 

## Checklist:

### DIFF
--- a/main/main.ino
+++ b/main/main.ino
@@ -1963,8 +1963,8 @@ int unixtimestamp() {
 String UTCtimestamp() {
   time_t now;
   time(&now);
-  char buffer[sizeof "yyyy-MM-ddThh:mm:ssZGMT"];
-  strftime(buffer, sizeof buffer, "%FT%TZGMT", gmtime(&now));
+  char buffer[sizeof "yyyy-MM-ddThh:mm:ssZ"];
+  strftime(buffer, sizeof buffer, "%FT%TZ", gmtime(&now));
   return buffer;
 }
 


DigiH:
Hi @dkneisz thanks for the fix of the time zone naming. 

Now that we have the chance, and as Z(ulu) time and UTC are identical, but UTC is generally being used to define time zones around the world with UTC +/- X, and it being less militaristic than Zulu time …

@dkneisz @1technophile 
… should we finalise it with a change to UTC? 

Would that also work with your OpenHAB set up?

dkneisz:
@DigiH I'm not sure what you mean by that. The "Z" in the UTC time just means that I's in the GMT timezone. Alternatively you could use "+0000" for that. Another option would be to just set the local timezone and use that in the UTC timestamp, but for some reason the function "gmtime" is used there.

DigiH:
The function **gmtime** gives back the UTC - to just add a bit more to the naming confusion, as GMT strictly speaking is also identical to UTC and Z, but historically has changed and therefore still has different meanings for different people.

All I thought was that instead using the formatting "yyyy-MM-ddThh:mm:ssZ", we could have "yyyy-MM-ddThh:mm:ssUTC" for easier international recognition of the time stamp we give.

Let me quickly test it here …

DigiH:
So with

```
  char buffer[sizeof "yyyy-MM-ddThh:mm:ssUTC"];
  strftime(buffer, sizeof buffer, "%FT%TUTC", gmtime(&now));
```

and then getting

`…"UTCtime":"2023-06-16T14:13:00UTC"…`

I just thought it might avoid possible confusion for some users with the Z, but with the key already saying UTCtime all is fine either way :) 

Merging with the original Z

Thanks

PR #1713, opened 2023-07-14
[RF] rain_rate_mm_h fix

DigiH:
https://community.openmqttgateway.com/t/rain-rate-mm-h-ignored/2597

as defined in several rtl_433 decoders

## Checklist:


1technophile:
Thks, original was taken from https://github.com/merbanan/rtl_433/blob/1026deab33372e4fe2e6c1d4c0f3df0c02c491be/examples/rtl_433_mqtt_hass.py#L362

Not sure if we should remove rain_mm_h or keep it added to rain_rate_mm_h

DigiH:
Initially I thought that as well, but in all the rtl_433_ESP decoders I couldn't find any **rain_mm_h** at all, only ever **rain_rate_mm_h**, so not sure where **rain_mm_h** could come from otherwise.

Same for rtl_433 actually, only ever mentioned in the example, probably an old/legacy/wrong entry.

https://github.com/merbanan/rtl_433/pull/2565

1technophile:
Make sense, maybe we should submit a PR in rtl_433 repo

DigiH:
See my link above :)

1technophile (approved):

PR #1714, opened 2023-07-14
INA226 JSON do not output leading spaces

### LINKED ISSUE #1691: INA226 incorrect JSON output
Hello,

JSON output from INA226 sensor is incorrect. Output contain space before values, e.g.:  {"volt": 3.3,...} and it should be {"volt":3.3,...} (without space before value). 

My workaround is to modify "ZsensorINA226.ino" to:
...   
//char volt_c[7];
//char current_c[7];
//char power_c[7];
//dtostrf(volt, 6, 3, volt_c);
//dtostrf(current, 6, 3, current_c);
//dtostrf(power, 6, 3, power_c);
INA226data["volt"] = (float)volt;
INA226data["current"] = (float)current;
INA226data["power"] = (float)power;
...

1. Publish INA226 to MQTT with JSON
2. Use Telegraf "mqtt_consumer" with json output to influxdb to read this topic 
3. InfluxDB/Grafana will be not able to read/graph this data

**Expected behavior**
Valid JSON output format like other sensors.

Thank you. 

Peter

NorbertHD:
## Description:
Closes #1691

## Checklist:

### DIFF
--- a/main/ZsensorINA226.ino
+++ b/main/ZsensorINA226.ino
@@ -69,15 +69,9 @@ void MeasureINA226() {
     float current = shuntvolt * 0.0000025 / rShunt; // * LSB / R
     float power = abs(volt * current);
 
-    char volt_c[7];
-    char current_c[7];
-    char power_c[7];
-    dtostrf(volt, 0, 3, volt_c);
-    dtostrf(current, 0, 3, current_c);
-    dtostrf(power, 0, 3, power_c);
-    INA226data["volt"] = (char*)volt_c;
-    INA226data["current"] = (char*)current_c;
-    INA226data["power"] = (char*)power_c;
+    INA226data["volt"] = volt;
+    INA226data["current"] = current;
+    INA226data["power"] = power;
     pub(subjectINA226toMQTT, INA226data);
   }
 }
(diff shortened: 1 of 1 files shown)


1technophile:
I think the code could be simplified to have:
```c
    INA226data["volt"] = (float)volt;
    INA226data["current"] = (float)current;
    INA226data["power"] = (float)power;
```
And get rid of the `dtostrf`

NorbertHD:
1. Why cast to float? volt, current and power are already of type float.
2. With dtostrf you have exactly 2 decimal places. But with your proposal, you have rounding artifacts.
   e.g. 0.28 gets 0.280000001 or 3.43 gets 3.430000067 in the json data.

So how should we go on?

1technophile:
1) The casting was a way to specify ArduinoJson the type, but it may not be necessary
https://arduinojson.org/v6/api/config/use_double/

2) is it the responsibility of OMG to choose the number of decimals to display rather than the controller? Still trying to figure out this one. As it is use cases related, we could stick with the behavior of ZsensorHTU21, ZsensorLM75, ZsensorSHTC3, ZsensorDHT not rounding or round it like in ZsensorRN8209
Open to discussion on this

NorbertHD:
I have updated the PR, I think not rounding is the correct way. Data should be transported and stored in full precision and only rounded when displaying.

1technophile:
Thanks

PR #1715, opened 2023-07-15
User friendly manual network configuration

vi7:
## Description:

1. Add ability for user friendly IP address override without the need to touch `User_config.h`
2. Add DNS server option for ESP WiFi STA

NOTE: Tested on Wemos D1 mini (ESP8266) **ONLY**, unfortunately I do not have a possibility to test on Arduino and ESP32

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).

PR #1716, opened 2023-07-15
Add env for Lolin Wemos D1 mini with CC1101

vi7:
## Description:

1. Add env for Lolin (Wemos) D1 Mini with CC1101 transceiver

Network related parameters (`-DNET_IP` etc.) are dependant on the feature delivered by #1715

NOTE: My editor automatically fixed some redundant spaces. Pls let me know if this is not needed

## Checklist:

### DIFF
--- a/prod_env.ini.example
+++ b/prod_env.ini.example
@@ -1,5 +1,5 @@
 [platformio]
-default_envs = 
+default_envs =
   esp32dev-ble-1
   esp32dev-ble-2
   env:nodemcuv2-ONOFFPILIGHT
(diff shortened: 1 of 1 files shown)


1technophile:
Thanks for the PR. I'm not sure of the value of adding a new environment for CC1101 and an ESP8266 based board, as we have already `nodemcuv2-rf-cc1101`

Do you know that you can store your environment into a prod_env.ini file separate from environments.ini ?

vi7:
@1technophile yep, `prod_env.ini` is exactly how I'm using it right now. The reason for the PR is that `nodemcuv2-rf-cc1101` appeared to be incompatible with Wemos D1 mini boards as-is, so I thought dedicated env could be helpful for people who are using D1 mini board.

I'm ok with any decision from your side however.

1technophile:
Let's add it as an example only in this case (commented)

vi7:
@1technophile I've moved D1 mini example to the `prod_env.ini.example` file

1technophile:
Thanks

PR #1720, opened 2023-07-20
add custom SPI config for CC1101 + added board

### LINKED ISSUE #1665: CC1101 setSpi call if SPI pins are set
Hi All

**Is your feature request related to a problem? Please describe.**
I'm using [M5 Stamp-C3](https://docs.m5stack.com/en/core/stamp_c3) (ESP32-C3) and cannot use the default pins configuration as some of the pins are not available on my board
[Pinout](https://static-cdn.m5stack.com/resource/docs/products/core/stamp_c3/stamp_c3_09.webp)

**Describe the solution you'd like**
Add `ELECHOUSE_cc1101.setSPI(SCK, MISO, MOSI, CS);` call just above `ELECHOUSE_cc1101.Init();` if SPI pins are set to non-default
[All mentions](https://github.com/search?q=repo%3A1technophile%2FOpenMQTTGateway+ELECHOUSE_cc1101.Init%28%29%3B&type=code)

**Describe alternatives you've considered**
Only other solution is not using boards with limited pinout

**Additional context**
I tested it with the library and It works fine. I plan to make a PR with the changes

efuturetoday:
## Description:
You can now configure the SPI connection pins via the following defines:

- RF_MODULE_SCK
- RF_MODULE_MISO
- RF_MODULE_MOSI
- RF_MODULE_CS

Also added a board config for WT32-ETH01 with CC11101 and RF.
fixes #1665 

## Checklist:

### DIFF
--- a/environments.ini
+++ b/environments.ini
@@ -5,48 +5,6 @@
 ;List of environments that can be build                                                ;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-
-[env:esp32dev-multi_receiver-eth01]
-platform = ${com.esp32_platform}
-board = wt32-eth01
-board_build.partitions = min_spiffs.csv
-lib_deps =
-  ${com-esp.lib_deps}
-  ${libraries.wifimanager32}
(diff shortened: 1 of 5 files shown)


1technophile:
Thanks for the PR, I'm not a fan of adding many environments combinations (as long as we test already RF and WT31) unless there is already a board existing with the different modules.
Do you know that you can store your environment in a prod_env.ini ?

PR #1725, opened 2023-08-03
[BT] Restarting the tasks is not enough in case of BLE stack stuck

1technophile:
## Description:
So restart the ESP instead

## Checklist:

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -68,8 +68,6 @@ struct decompose {
   bool reverse;
 };
 
-unsigned long BLEstarts = 0;
-
 vector<BLEAction> BLEactions;
 
 vector<BLEdevice*> devices;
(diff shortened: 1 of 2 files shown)


eljojo:
@1technophile Hi Florian, would you mind cutting a new release? been observing times without data capture since the last update that I think could be fixed by this! (PS: great job on the web ui and gateway in general!)

1technophile:
You can try the development version from here:
https://docs.openmqttgateway.com/dev/ and report if it solves your issue.

eljojo:
I think it does! 
image

before seeing your message it came to mind I could [rebuild by forking the repo and pushing an empty commit](https://github.com/eljojo/OpenMQTTGateway/actions/runs/6117082838), so I did that and extracted the upgrade image from the artifacts. Been running it for a few days and it seems to restart appropriately. 

it's interesting to be between 60k-65k seconds

eljojo:
I wonder why the restarts became necessary, I couldn't observe any other pattern that would indicate. It seemed to not happen before the web ui.

1technophile:
The restart is necessary due to this issue 
https://github.com/espressif/arduino-esp32/issues/5860

PR #1759, opened 2023-09-23
console log formatting

lulol:
## Description:
Fix some console log formatting that is actually mixed with LF and CRLF.

To work properly still is need to change in ArduinoLog/ArduinoLog.h
`#define CR "\n"` <<< actually LF
to
`#define CR "\r\n"`
or to use NL instead of CR in all the relevant files.
NL is available in the latest version of ArduinoLog.

## Checklist:


1technophile:
Thanks for pointing this, I have fixed it in the Arduino log fork

PR #1761, opened 2023-09-27
[SYS] Configuration portal for Ethernet board and HA link to config

1technophile:
## Description:
Enable Wifi Manager as a configuration portal for Ethernet boards, also leverage the Wifi Access point given in the configuration portal as a fallback connectivity method.

With this PR users can now use pre-built binaries with ethernet, wifi manager is used to configure the parameters, it is accessible through ethernet and the wifi access point.
The user can still add a Wifi Access point to connect, it will be used as a fallback connectivity method. The SSID and password can also be left blank to use only ethernet connection.

This PR is interesting for boards like the ThingPulse and Olimex gateways.

Add "visit" link to the WebUI from HA to make the configuration access easier and reduce key length
image

## Checklist:

### DIFF
--- a/docs/upload/portal.md
+++ b/docs/upload/portal.md
@@ -1,14 +1,18 @@
-# WiFi and MQTT configuration
+# Wifi and MQTT configuration
 
-## WiFi and MQTT configuration portal for WiFi boards
+## Configuration portal
 
 Once loaded into your ESP, and if you don't use the manual configuration, you have to set your network parameters with WiFi Manager portal.
 
-From your smartphone, search for  OpenMQTTGateway or OMG WiFi network and connect to it with the `WifiManager_password` you have defined (the default password is **"your_password"**), or if you are using a device sold by Theengs or the macro `-DWM_PWD_FROM_MAC=true`, the password will be the last eight digits of the device MAC Address with upper case.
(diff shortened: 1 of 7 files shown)


1technophile:
@marcelstoer @hallard for information

marcelstoer:
Nice! I need to test this sometime.

docs/upload/portal.md
```diff
@@ -1,23 +1,32 @@
-# WiFi and MQTT configuration
+# Wifi and MQTT configuration
```
marcelstoer, line 1: I value consistency a lot. You sometimes use "Wifi" and sometimes "WiFi". Wikipedia suggests you should use "Wi-Fi" but I personally don't care as long as it's spelled the same consistently. However, as this term is actually an abbreviation I think "Wifi" is the least correct of the three.

1technophile: Several occurrences of this, we will fix it in another PR

DigiH: I change all inconsistent ones as well whenever I spot them in a document I'm making changes in anyway - generally to **WiFi**, e.g.

https://github.com/1technophile/OpenMQTTGateway/commit/ba1877fc3010c8cc030731770af03d5f2f7ff269

as I value consistency a lot as well ;)

hallard:
Awesome, thanks, will test also sometime.

PR #1767, opened 2023-10-03
[SYS] Implement a central queue

1technophile:
## Description:
To avoid concurrency issues when modules or/and gateways are publishing data. A queue system was previously in place for the BT gateway. With this PR it is now used by all the modules.
Previously having different modules potentially assigned to different cores was generating MQTT disconnections.

- Simplify signature
- BuildBTtopic outside of main
- Cherry pick 1736
- Improve queue size
- Simplify Hapresence
- Allocate json buffer in the heap to avoid overloading the callback

Why not use the FreeRTOS queue management function? I gave it a try but was not satisfied with the stability and got better stability with the stl

## Checklist:

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -872,7 +872,7 @@ void setupBTTasksAndBLE() {
   xTaskCreatePinnedToCore(
       procBLETask, /* Function to implement the task */
       "procBLETask", /* Name of the task */
-      7000, /* Stack size in bytes */
+      8000, /* Stack size in bytes */
       NULL, /* Task input parameter */
       2, /* Priority of the task (set higher than core task) */
       &xProcBLETaskHandle, /* Task handle. */
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -418,7 +418,7 @@ void updateDevicesStatus() {
         JsonObject BLEdata = BLEdataBuffer.to<JsonObject>();
         BLEdata["id"] = p->macAdr;
         BLEdata["state"] = "offline";
-        buildBTtopic(BLEdata);
+        buildTopicFromId(BLEdata, subjectBTtoMQTT);
         enqueueJsonObject(BLEdata);
         // We set the lastUpdate to 0 to avoid replublishing the offline state
         p->lastUpdate = 0;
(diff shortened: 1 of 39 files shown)


main/ZgatewayBT.ino
```diff
@@ -1178,21 +1081,38 @@ void PublishDeviceData(JsonObject& BLEdata) {
...
+    enqueueJsonObject(BLEdata);
   } else {
-    Log.trace(F("Low rssi, device filtered" CR));
+    Log.notice(F("Low rssi, device filtered" CR));
+    return;
   }
 }
 
 void process_bledata(JsonObject& BLEdata) {
+  yield();
```
h2zero, line 1141: Why do we need to yield here?

main/ZsensorRN8209.ino
```diff
@@ -40,7 +40,6 @@ float power = 0;
 TaskHandle_t rn8209TaskHandle = nullptr;
 
 unsigned long PublishingTimerRN8209 = 0;
-StaticJsonDocument<JSON_MSG_BUFFER> doc;
```
h2zero, line 43: Not sure about this. Moving it into the loop will use stack space instead of heap, which could lead to an overflow.

main/ZgatewayBT.ino
```diff
@@ -681,16 +590,22 @@ void procBLETask(void* pvParameters) {
...
-
+    // Feed the watchdog
+    //esp_task_wdt_reset();
     if (!ProcessLock) {
       Log.trace(F("Creating BLE buffer" CR));
-      JsonObject& BLEdata = getBTJsonObject();
-      String mac_adress = advertisedDevice->getAddress().toString().c_str();
-      mac_adress.toUpperCase();
-      BLEdata["id"] = (char*)mac_adress.c_str();
+      StaticJsonDocument<JSON_MSG_BUFFER> BLEdataBuffer;
```
h2zero, line 598: May want to consider declaring this outside of a function so it does not use stack space, could be safely declared globally.

h2zero (commented):
Looks good, just a couple comments.

main/ZgatewayBT.ino
1technophile: Thanks updated

main/ZsensorRN8209.ino
1technophile: Moved

PR #1774, opened 2023-10-14
[CI] Keep existing gh-pages file when doing dev deployments

1technophile:
## Description:
This should avoid removing the production portal available in the root directory

## Checklist:


mikepparks:
The docs portal is currently returning a 404.

1technophile:
Thanks, should be fixed in a few minutes

PR #1777, opened 2023-10-20
[BT] BM2 adjustment to only require model_id

DigiH:
BM2 adjustment to only require model_id for connection update and not broadcast names any longer with adjusted BM2 decoder

## Checklist:


1technophile:
good to go, thanks

PR #1814, opened 2023-11-21
[DISC] Do not retain autodiscovery on/off commands.

ianmtaylor1:
Set "retain":false within autodiscovery messages for Home Assistant and OpenHAB discovery switch.

## Description:

[Community thread](https://community.openmqttgateway.com/t/interaction-between-discovery-auto-off-timer-and-retain-flag/2860)

As discovery state is stored in flash memory and there is an auto off timer, we don't want discovery turned back on by a retained MQTT message when device reboots.

## Checklist:


1technophile:
Thanks

PR #1812, opened 2023-11-16
[RTL_433][BREAKING] Beta support for FSK modulation, webUI frequency and active receiver change

1technophile:
## Description:
This PR adds environments supporting FSK with RTL_433 and the capability to change the receiver frequency and switch between receivers (CC1101 only) through the WebUI

BREAKING: this PR changes the MQTT API for frequency and active receiver changes:
Before:
`home/OpenMQTTGateway/commands/MQTTtoRTL_433 {"mhz":315.026}`
`home/OpenMQTTGateway/commands/MQTTto433 {"active":true}`

`Frequency macro: CC1101_FREQUENCY`

Acknowledgement topic for Pilight was 
`{"message":{"id":9,"unit":0,"state":"down"},"protocol":"arctech_screen_old","value":"","repeats":2,"status":2}`

After:
`home/OpenMQTTGateway/commands/MQTTtoRF/config {"frequency":315.026}`
`home/OpenMQTTGateway/commands/MQTTtoRF/config {"active":2}`

`Frequency macro: RF_FREQUENCY`

Acknowledgement topic for Pilight:
`{"id":9,"unit":0,"state":"down"}`

## Checklist:
  - [X] The pull request is done against the latest development branch
  - [X] Only one feature/fix was added per PR and the code change compiles without warnings
  - [X] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).

PR #1874, opened 2024-01-20
[RF] Use exponential backoff delay in initCC1101()

ianmtaylor1:
## Description:

In #1812, the function initCC1101() was created to poll getCC1101() until it returns true before using the CC1101 chip. A 500ms delay was put between polls.

In my experience, this delay is too long as the the chip is often ready very shortly after the first check. This change makes the function use exponential backoff to initially check quickly (16ms), then double the wait time until waiting the original 500ms.

A simpler change would be to just use a smaller constant delay, but exponential backoff is a good compromise in case there are instances where a longer delay is needed.

## Checklist:


1technophile:
Thanks!

PR #1907, opened 2024-03-04
[SYS] Log Theengs props deserilization failures

ssalonen:
Now we log failing theengs json props deserialization failures. 

**Note:** the rest of the function implementation is kept as is, with no null/data-validity checks, and NoMemory-erroring deserialization is likely to follow with a crash later in the strcmp conditionals like in #1905.

The code could be implemented more defensivly, in such a way that approriate null checks are in place before strcmp...then the code would work with "best-effort" basis in limited-memory (non-ESP32) environments. However, as pointed by the main author 1technophile in #1905, support for other boards is likely dropped in the future.

Related to #1905

## Checklist:


1technophile:
Thanks

PR #1917, opened 2024-03-23
Use configured "packages_dir"

ivankravets:
* https://docs.platformio.org/en/latest/scripting/examples/platformio_ini_custom_options.html
* https://docs.platformio.org/en/latest/projectconf/sections/platformio/options/directory/packages_dir.html

A user can override `packages_dir` or `core_dir`. So, it is better to use real value instead of static pointed to the HOME.

## Checklist:


1technophile:
Thanks

PR #1290, opened 2022-10-18
Re-publish BTtoMQTT Theengs discovery topics on MQTT re-connect

ssalonen:
- Improved version of #1274, this should re-publish more according to my local tests, meeting my original intent with https://github.com/1technophile/OpenMQTTGateway/issues/1273
- Handle return value of (MQTT) client.loop() to detect connection loss faster
- remove accidentally committed main.ino.cpp in previous PR, similar to #1289

Signed-off-by: Sami Salonen <ssalonen@gmail.com>

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).

PR #1274, opened 2022-09-25
Re-publish discovery topics on re-connect (via cfg, disabled by default)

ssalonen:
Resolves https://github.com/1technophile/OpenMQTTGateway/issues/1273

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).

PR #1933, opened 2024-04-14
[HA Discovery] Do not deactive autodiscovery when discovery_republish_on_reconnect is enabled

ssalonen:
Do not deactive autodiscovery in certain cases when `discovery_republish_on_reconnect` (see https://github.com/1technophile/OpenMQTTGateway/pull/1274 and https://github.com/1technophile/OpenMQTTGateway/pull/1290 ) is enabled

The introduction of `DiscoveryAutoOffTimer` in https://github.com/1technophile/OpenMQTTGateway/commit/3f92e14ba6d22173bba0b5b7d43743663a5eac1a / https://github.com/1technophile/OpenMQTTGateway/pull/1541 introduced a regression for the re-publishing of discovery topics (`discovery_republish_on_reconnect`) as first introduced in https://github.com/1technophile/OpenMQTTGateway/pull/1290 - after `DiscoveryAutoOffTimer` has elapsed, the system does not republish the topics even when MQTT connection is re-established.

This PR disables the effect of `DiscoveryAutoOffTimer` when `discovery_republish_on_reconnect` is enabled

## Checklist:

### DIFF
--- a/main/main.ino
+++ b/main/main.ino
@@ -2003,8 +2003,9 @@ void loop() {
       // We have just re-connected if connected was previously false
       bool justReconnected = !connected;
 #ifdef ZmqttDiscovery
-      // Deactivate autodiscovery after DiscoveryAutoOffTimer
-      if (SYSConfig.discovery && (now > lastDiscovery + DiscoveryAutoOffTimer))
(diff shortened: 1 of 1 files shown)


1technophile:
Good catch,, thanks

PR #1954, opened 2024-05-28
[SYS] Use PicoMQTT for MQTT communication

mlesniew:
## Description:

This replaces the MQTT library from [PubSubClient](https://github.com/knolleary/pubsubclient) to [PicoMQTT](https://github.com/mlesniew/PicoMQTT).  

This PR doesn't change functionality, the changes shouldn't affect the end user, because the project should work just as before.  

Switching to PicoMQTT opens the path to implementing a MQTT broker withing OMG.   

## TODO

A few things still need to be updated before this is merged:
- The project does not build for the `rfbridge-direct`, `shelly-plus1` and `nodemcuv2-pilight` environments.  These configurations use outdated board core versions, which are not supported by PicoMQTT.  Simply updating the configs to use newer core versions results in linker errors.
- Just like in the original code, MQTT settings are reverted to old values when a new configuration doesn't work.  However, the original logic around this was complex and recently updated, so it needs a careful review.      

## Checklist:

### DIFF
--- a/main/main.ino
+++ b/main/main.ino
@@ -77,7 +77,9 @@ bool ready_to_sleep = false;
 
 #include <ArduinoJson.h>
 #include <ArduinoLog.h>
-#include <PubSubClient.h>
+#include <PicoMQTT.h>
+
+#include <memory>
 
 struct JsonBundle {
   StaticJsonDocument<JSON_MSG_BUFFER> doc;
@@ -221,9 +223,6 @@ struct GfSun2000Data {};
 
 void setupTLS(int index = CNT_DEFAULT_INDEX);
 
-//adding this to bypass the problem of the arduino builder issue 50
-void callback(char* topic, byte* payload, unsigned int length);
-
 char ota_pass[parameters_size] = gw_password;
 #ifdef USE_MAC_AS_GATEWAY_NAME
 #  undef WifiManager_ssid
@@ -238,12 +237,12 @@ int failure_number_ntwk = 0; // number of failure connecting to network
 int failure_number_mqtt = 0; // number of failure connecting to MQTT
 
 unsigned long timer_led_measures = 0;
-static void* eClient = nullptr;
 static unsigned long last_ota_activity_millis = 0;
 // Global struct to store live SYS configuration data
 SYSConfig_s SYSConfig;
 
 bool failSafeMode = false;
+static bool mqttSetupPending = true;
 static int cnt_index = CNT_DEFAULT_INDEX;
 
 #ifdef ESP32
@@ -300,8 +299,8 @@ ESP8266WiFiMulti wifiMulti;
 #  include <Ethernet.h>
 #endif
 
-// client link to pubsub MQTT
-PubSubClient client;
+std::unique_ptr<::Client> eClient;
+std::unique_ptr<PicoMQTT::Client> mqtt;
 
 template <typename T> // Declared here to avoid pre-compilation issue (missing "template" in auto declaration by pio)
 void Config_update(JsonObject& data, const char* key, T& var);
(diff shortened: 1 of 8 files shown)


main/main.ino
```diff
@@ -3183,77 +3251,22 @@ void MQTTtoSYS(char* topicOri, JsonObject& SYSdata) { // json object decoding
 
     if (save_cnt || test_cnt) {
       // Stop the processing/disconnect
+#if 0
+// TODO: Is this needed?
```
mlesniew: Is this section still needed?  

Should `stopProcessing()` be called or `ProcessLock` set inside the `setupMQTT()` function now?

main/main.ino
```diff
@@ -3112,23 +3182,20 @@ void MQTTtoSYS(char* topicOri, JsonObject& SYSdata) { // json object decoding
...
-    char prev_mqtt_user[parameters_size];
-    strcpy(prev_mqtt_user, cnt_parameters_array[prev_cnt_index].mqtt_user);
-    char prev_mqtt_pass[parameters_size];
-    strcpy(prev_mqtt_pass, cnt_parameters_array[prev_cnt_index].mqtt_pass);
-    char prev_mqtt_server[parameters_size];
-    strcpy(prev_mqtt_server, cnt_parameters_array[prev_cnt_index].mqtt_server);
-    char prev_mqtt_port[6];
-    strcpy(prev_mqtt_port, cnt_parameters_array[prev_cnt_index].mqtt_port);
-    bool prev_mqtt_secure = cnt_parameters_array[prev_cnt_index].isConnectionSecure;
-
```
mlesniew, line 3196: The original code first stores the old value and `cnt_index` and `cnt_parameters_array[cnt_index]` to revert the values of both if the new config doesn't work.

However, a few lines below, the `cnt_index` value is reassigned based on what is in the user provided json.  Next the item pointed to by the updated `cnt_index`  in `cnt_parameters_array` gets overwritten.

This means that reverting the settings will not work, because we're storing the wrong `cnt_parameters_array` element.

The PR fixes that (I believe), but this needs careful review, because the logic is complex.

main/main.ino
1technophile: Yes, still needed to free memory before reinitiating a new connection. Especially useful with secure connection. If there is not enough memory available the connection will fail

mlesniew: In that case, does it make sense to always keep ProcessLock set to true and processing stopped when mqtt is not connected? 

We can stop and restart processing easily in the PicoMQTT callbacks which are fired when a connection is established or lost.  Does that make sense?

1technophile: Yes, if we don't do this, the task will start and consume memory, which we need to do the HTTPS update check at the start.

mlesniew: OK, thanks I see.  

There's still one thing unclear to me.  In the original code (on the `development` branch) I see many places where `stopProcessing()` is called or `ProcessLock` is set to `true`.  But I don't see where or how processing is restarted or how `ProcessLock` is set back to `false`.

1technophile: Note that there are two process locks, one general `ProcessLock` and one for BT `BTProcessLock`

 `BTProcessLock` is `true` at the start to enable resource-intensive operations, it is released by `BTProcessLock = !BTConfig.enabled`

`ProcessLock` main goal is to stop the publication of the RN8209

The operations requiring a `processLock` should trigger a restart of the ESP

mlesniew:
The configurations, which failed to build should be fixed now.  Their board cores are upgraded: 
* In contrast to other configs, `shelly-plus1` uses the [Tasmota core](https://github.com/tasmota/platform-espressif32), it's now updated to use [2.0.7](https://github.com/tasmota/platform-espressif32/releases/tag/2023.02.00).  This is the minimum version supported by PicoMQTT.  Unfortunately, upgrading to a newer version of the core is not possible, because it doesn't include support for the legacy SPIFFS file system, which OMG uses.
* `rfbridge-direct` and `nodemcuv2-pilight` are upgraded to use the same core version as all other espressif8266 configs.  Unfortunately, simply updating the platform version number was not sufficient.  There were linking errors in the [ESPiLight](https://github.com/puuu/ESPiLight) library.  I found that the root cause for that was the updated GCC version in the newer core, which has different default compilation flags -- more details can be found [here](https://github.com/pilight/pilight/pull/469).  The problem is now fixed with an extra `-fcommon` build flag.

Let's see if all configs now succeed in the CI.

main/main.ino
mlesniew: I guess that means that after calling `stopProcessing()` the ESP should also be eventually restarted, right?

1technophile: yes it should be, let me know if you see cases where it is not

mlesniew: I see now how the board is reset after applying new MQTT settings and reconnecting, thanks for clarifying.

I've made changes to my code too to keep that behavior. 

I haven't looked into other parts of the code very carefully, but if I spot a corner case leading to a process lock or processing stop without a reboot, I'll let you know.

mlesniew:
@1technophile, I think this PR is now ready to review, please let me know if there's anything that needs updated.

1technophile:
I need to add #1979 , it has been removed by the merge

mlesniew:
@1technophile no problem, I have now additionally squashed the fixup commit too, so this should be ready for merging as is.

1technophile:
Thanks

PR #1958, opened 2024-05-30
[DISC] battery_ok discovery differentiation

DigiH:
Differentiation between 0%-100% float battery_ok devices' discovery and binary_sensor devices battery_ok

**battery_ok binary_sensor**
image
image

**battery_ok float**
image

## Checklist:


DigiH:
The one issue remaining, not quite breaking, but slightly confusing, is that previous wrong battery_ok sensor % discoveries do not get overwritten by the new binary_sensor battery_ok discovery, but they exist alongside each other until the sensor % discovery entry is manually deleted, or the whole MQTT integration device is being deleted and discovered afresh.

Still very much worth it for proper battery_ok discovery in my opinion.

1technophile:
Thanks

PR #1959, opened 2024-06-08
[DISC] always auto-discover Presence/Tracker timeout

DigiH:
auto-discover Presence/Tracker timeout even when HASS presence is not active

Also inject an additional placeholder rssi key and value into the BM2 voltage connection message to avoid the message being registered as a device tracker AWAY

https://community.openmqttgateway.com/t/bm2-home-and-away/3315

## Checklist:


1technophile:
Thanks

Issue #1909, opened 2024-03-09
Wrong state topic (stat_t) in homeassistant config payload for TPMS-Jansite-Solar


riegelbrau:
The  TPMS-Jansite-Solar is supported by OMG since FSK can be enabled. With OMG 1.60 and 1.70 I face the following behaviour running OMG on a LilyGo. The sensor data is sent with the mqtt topic `OMG/OMG_lilygo_rtl_433_ESP/RTL_433toMQTT/TPMS/Jansite-Solar/5e512a` and this payload
```
{
	"model": "Jansite-Solar",
	"type": "TPMS",
	"id": "5e512a",
	"flags": 197,
	"pressure_kPa": 68.8,
	"temperature_C": 18,
	"code": "5e512ac5492b009413",
	"mic": "CRC",
	"protocol": "Jansite TPMS Model Solar",
	"rssi": -75,
	"duration": 21000
}
```
When ohdisc=true is set, the HASS config message for openHAB is created under this topic: `homeassistant/sensor/TPMS-Jansite-Solar-5e512a-pressure_kPa/config` (example for pressure).
The payload is
```
{
	"stat_t": "+/+/RTL_433toMQTT/Jansite-SolarSolar/5e512a",
	"dev_cla": "pressure",
	"name": "pressure",
	"uniq_id": "TPMS-Jansite-Solar-5e512a-pressure_kPa",
	"val_tpl": "{{ value_json.pressure_kPa }}",
	"state_class": "measurement",
	"device": {
		"ids": [
			"TPMS-Jansite-Solar-5e512a"
		],
		"cns": [
			[
				"mac",
				"TPMS-Jansite-Solar-5e512a"
			]
		],
		"mdl": "Jansite-Solar",
		"name": "TPMS-Jansite-Solar-5e512a",
		"via_device": "OMG_lilygo_rtl_433_ESP"
	}
}
```
As you can see, the json value for stat_t contains the wrong topic `+/+/RTL_433toMQTT/Jansite-SolarSolar/5e512a`. The type part `TPMS` is missing and the two consecutive strings `SolarSolar` look like a typo, where only one should be.

**Steps to reproduce the behavior:**
1. Set the config parameter ohdisc to true.
2. Let the sensor send some data by changing the environment pressure or temperature.
3. See good looking mqtt message, but wrong HASS config message.

**Expected behavior:**
[...]

PR #1960, opened 2024-06-09
[RTL] Correct string handling for type key in NAMING_KEYS

riegelbrau:
Extend class structure RTL_433device and modify code to create correct state topic in discovery message to support sensors, that send type key as well as sensors, that don't. Actually only TPMS sensors use type=TPMS. See issue #1909.

## Description:
The key type of the naming_keys array in ZgatewayRTL_433.ino was not used to create the state topic in discovery messages. Thcaused, that for TPMS sensors the state topic was incorrectly created. With this change, both cases are respected, when key is used and when key is not used.

I tested with my sensors TPMS Jansite-Solar (FSK) and AmbientWeather-TX8300-1-65 (OOK). The sensors were correctly discovered and configured in my openHAB instance.

## Checklist:

### DIFF
--- a/main/ZgatewayRTL_433.ino
+++ b/main/ZgatewayRTL_433.ino
@@ -66,11 +66,12 @@ void dumpRTL_433Devices() {
     RTL_433device* p = *it;
     DISCOVERY_TRACE_LOG(F("uniqueId %s" CR), p->uniqueId);
     DISCOVERY_TRACE_LOG(F("modelName %s" CR), p->modelName);
+    DISCOVERY_TRACE_LOG(F("type %s" CR), p->type);
     DISCOVERY_TRACE_LOG(F("isDisc %d" CR), p->isDisc);
   }
 }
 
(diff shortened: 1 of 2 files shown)


riegelbrau:
Who can help me with the Code Format error? I cannot see, what's wrong.

DigiH:
@riegelbrau you have extra spaces after several lines in both files.

Don't you get **Format Document** when right clicking into files in VSC? This should automatically format code correctly as defined by the included .clang-format.

riegelbrau:
Thanks @DigiH, this pointed me to the reason. I do not yet edit with VSCode, but use Notepad++, because I am familiar with it. There I can also see trailing spaces and remove them. I did so and now the PR compiled successfully.

DigiH:
You even caught some extra spaces in the commented starting section of config_RF.h which the automatic clang formatting missed, it only looking at actual code 😄

riegelbrau:
I recognized that. Is it good or bad?

DigiH:
It's good 👍

1technophile:
Thanks

PR #1988, opened 2024-07-07
Allow running an MQTT broker within OMG

mlesniew:
## Description:
This change adds the possibility to run a MQTT broker instead of an MQTT client on the microcontroller. 

To switch to broker mode, the `MQTT_BROKER_MODE` must be defined to `true`.  A new environment configuration `esp32dev-ble-broker` has been added and it has this preprocessor macro defined.  It is otherwise identical to `esp32dev-ble`. 

Currently the broker is listening on the WiFi interface, on port 1883.  To keep the PR small and simple, there currently are little configuration possibilities in broker mode.  However, the code is extendible and support for listening on different ports, allowing websocket connections, or listening on Ethernet can be easily added.  

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,6 +79,7 @@ jobs:
           - "theengs-bridge"
           - "esp32dev-ble-idf"
           - "theengs-bridge-v11"
+          - "esp32dev-ble-broker"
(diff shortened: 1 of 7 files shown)


1technophile:
Thanks!

Issue #1996, opened 2024-08-05
Add ability to change discovery prefix on the fly either via WebGUI or mosquitto_sub


puterboy:
**Is your feature request related to a problem? Please describe.**
I am using OMG_lilygo_rtl_433_esp to capture 433 messages from my various sensors around my house and then post them to my Mosquitto MQTT broker for use in HomeAssistant.

- On the one hand, I need to **shut off** OpenMQTT Gateway "Auto Discovery" because otherwise HA dutifully recognizes and inserts new devices and sensors for the dozens of devices that my neighbors have (which then get further duplicated whenever they or even I change batteries in various devices).

- On the other hand, when I add a new device myself, I want to generate a discovery message(s) for that device alone so that HA will recognize that sensor -- but I don't want to suck in my neighbors devices at the same time

**Describe the solution you'd like**
I would like to have the discovery prefix be changeable on-the-fly without having to recompile and re-flash firmware.
This could done either with the Web GUI (similar to how one can set the `MQTT Base Topic` from the `Configure MQTT` menu or by sending a "System Command" via mosquitto_pub as in https://docs.openmqttgateway.com/use/gateway.html

So one could have:
```
mosquitto_pub -t "home/OpenMQTTGateway/commands/MQTTtoSYS/config" -m '{  "discovery_topic": "homeassistant-staging" }
```
which would be analogous to how one sets the MQTT_topic:
```
mosquitto_pub -t "home/OpenMQTTGateway/commands/MQTTtoSYS/config" -m '{  "mqtt_topic": "topic/" }
```

NOTE: I used `discovery_topic` rather than say `discovery_prefix` to be consistent with the variable naming in the source code.

**Describe alternatives you've considered**
I have considered the following 3 workarounds:

1. Recompile and reflash the firmware for my Lilygo device with the Discovery Prefix changed to something other than the default discovery prefix used by HA (which is "homeassistant") and then manually cut and paste my newly discovered device from the new discovery prefix to the HA assistant one (or use mosquitto_sub/mosquitto_pub to do the same)

2. Change either temporarily or permanently the default discovery prefix for HA away from the default discovery prefix for OpenMQTTGateway (which is "homeassistant"). Then delete any unwanted neighbor discovery messages and turn off OpenMQTT Gateway auto-discovery before changing the HA discovery prefix back to its default

[...]

PR #2007, opened 2024-08-16
Add ability to change Discovery Prefix via WebUI or via mosquitto_pub

puterboy:
## Description:
Changes made as discussed in https://github.com/1technophile/OpenMQTTGateway/issues/1996
Specifically:
1. Declaration `discovery_Topic` deprecated (but still working) and changed to `discovery_Prefix` to avoid confusion with `discovery_topic`
2. New *variable* `discovery_prefix` introduced that defaults on new installs to discovery_Prefix (or discovery_Topic if present), is stored in NVS, and can be changed via WebUI  (under `Configure MQTT`) or via mosquitto_pub, using something like:
```
mosquitto_pub -t "home/<GATEWAY_NAME>/commands/MQTTtoSYS/config" -m '{  "discovery_prefix": "<NEW_DISCOVERY_PREFIX" }'
```

I tested it using an esp32 with `esp32dev-ble` and the functionality works as described.
Let me know if you are ok with the PR.

## Checklist:
  - [X ] The pull request is done against the latest development branch
  - [X ] Only one feature/fix was added per PR and the code change compiles without warnings


main/main.ino
```diff
@@ -1332,6 +1332,7 @@ void setup() {
   Log.trace(F("OpenMQTTGateway ip: %s" CR), WiFi.localIP().toString().c_str());
   Log.trace(F("OpenMQTTGateway index %d" CR), cnt_index);
   Log.trace(F("OpenMQTTGateway mqtt topic: %s" CR), mqtt_topic);
+  Log.trace(F("OpenMQTTGateway mqtt topic: %s" CR), discovery_prefix);
```
1technophile: OpenMQTTGateway mqtt discovery prefix

1technophile:
Thanks, could you update the documentation also.

The build ko for the AWS environment is because some environments do not use discovery. Your change needs to be included when the macro ZmqttDiscovery is defined.

Finally check the box for the DCO in the PR and we will be good

puterboy:
- Corrected the typo where mqtt_topic was used instead of discovery_prefix
- Added #ifdef ZmqttDiscovery in all *new* places that reference discovery_prefix (note if a line of code previously referenced discovery_Topic and was not conditioned before then I left it as is now)
- Added discovery_prefix key to the mosquito_pub documentation (Note the documentation doesn't address setting "MQTT Base Topic" except in initial startup where "Discovery Prefix" is not relevant -- so I didn't know where or if I should document "Discovery Prefix" for the regular WebUI

puterboy:
Not sure what the lint formatting issues are - I tried to copy how you spaced and indented macro if/else/end statements.

1technophile:
So as to format automatically your document you have to add the "clang-Format" extension to VSCode, once done, you can format the code by doing a right click into the code file window and clicking "Format document".

puterboy:
I was actually doing the editing in the web version of github – not VS.

Do you have the ability to fix the formatting on my patch?

1technophile:
Yes sure

puterboy:
Thanks!

Issue #2012, opened 2024-08-16
CORRUPTION in display of SSID and construction of discovery config strings


puterboy:
**Describe the bug**
I just upgraded to the latest dev branch and am finding two potentially related errors:
Note: The problem is not due to my discovery_prefix patches since I reverted and still see the problems
Note: I also noted it on 2 different devices (lilygo-rtl-433 and esp32dev-ble)

1. In the WebUI, under "Configure WiFi" the WiFi network name appears as gibberish characters -- this persist even if I re-name the network. Note that the stored name must be working since it attaches to the correct network.
![Screenshot 2024-08-16 165215](https://github.com/user-attachments/assets/84a665cb-d53f-49a1-8dc6-0b6ae7af4300)

 
2. Separately, but perhaps related, looking at the discovery topic config lines (via mosquitto_sub), there is gibberish and over-writing in the value of the 'stat_t' key.
For example:
```
{"stat_t":"+/+/RTL_433toMQTT/LaCrosse-TX141THBv2/1/212","0�Chome/OMG_lilygo_rtl_433_ESP/RTL_433toMQTT/LaCrosse-TX141THBv2/1/212{"model":"LaCrosse-TX141THBv2","id":212,"channel":1,"battery_ok":1,"temperature_C":-18.3,"humidity":78,"test":"No","mic":"CRC","protocol":"LaCrosse TX141-Bv2, TX141TH-Bv2, TX141-Bv3, TX141W, TX145wsdth, (TFA, ORIA) sensor","rssi":-67,"duration":143000}dev_cla":"humidity","unit_of_meas":"%","name":"H
```
It as if it the components constructing the json string are pointing to the wrong elements of memory since the resulting json string looks like a combination of a valid discovery topic config keys & values PLUSsome elements of a published data topic PLUS some elements of the config string of another entity for the device -- all mashed together.

Note that it should look something like:
```
[...]

PR #2021, opened 2024-08-25
[WEBUI] Update ZwebUI.ino - Convert WiFi.SSID() to c_string (JJK)

puterboy:
## Description: This fixes ONE of the TWO bugs mentioned in https://github.com/1technophile/OpenMQTTGateway/issues/2012
Specifically, the WiFi Network **always** displayed as garbage in the WebUI WiFi Configuration screen.
Problem is that WiFi.SSID() needs to be converted to a c_string -- so the fix is trivial.

I am surprised that nobody else has noted this bug as it seemingly applies to all installations and goes back to at least v1.60.

Please add this fix asap as it restores basic WebUI functionality for pretty much all users.

## Checklist:
  - [X ] The pull request is done against the latest development branch
  - [X ] Only one feature/fix was added per PR and the code change compiles without warnings
  - [X ] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).

### DIFF
--- a/main/ZwebUI.ino
+++ b/main/ZwebUI.ino
@@ -616,7 +616,7 @@ void handleWI() {
       response += String(wifi_script);
       response += String(script);
       response += String(style);
-      snprintf(buffer, WEB_TEMPLATE_BUFFER_MAX_SIZE, config_wifi_body, jsonChar, gateway_name, WiFiScan.c_str(), WiFi.SSID());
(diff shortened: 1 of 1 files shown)


1technophile:
Thanks!

PR #2033, opened 2024-08-28
Update main.ino - Remove restartesp from discovery_prefix stanza

puterboy:
## Description:
Revert from original patch as not necessary and also prevents saving of new MQTT variable values
See https://github.com/1technophile/OpenMQTTGateway/issues/1996#issuecomment-2313638704

## Checklist:

### DIFF
--- a/main/main.ino
+++ b/main/main.ino
@@ -3339,7 +3339,6 @@ void MQTTtoSYS(char* topicOri, JsonObject& SYSdata) { // json object decoding
 #ifdef ZmqttDiscovery
       if (SYSdata.containsKey("discovery_prefix")) {
         strncpy(discovery_prefix, SYSdata["discovery_prefix"], parameters_size);
-        restartESP = true; //Need to reset so all devices get re-discovered & published to new discovery_prefix
       }
 #endif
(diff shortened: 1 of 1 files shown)


1technophile:
Should it be added to the documentation that the gateway needs to be restarted after to republish to the new topic ?

puterboy:
No - as I detailed at length (maybe too much :)) in several new comments in https://github.com/1technophile/OpenMQTTGateway/issues/1996, a restart is triggered anyway since `discovery_prefix` is on the same WebUI page as the various MQTT login variables that in turn show (always) show as changed and trigger a configuration backup when `setupMQTT()` is called which in turn triggers a restart from within `setupMQTT()` (assuming changes were successful)

Indeed one way or other, most variables end up triggering a restart that way (whether intended or not).. Indeed, as I detail in my comments a restart really shouldn't be necessary for most if not all variables.

Also, `saveConfig()` is often called twice.

In all the code could deserve some straightening out as its effects are non-obvious and may not be as intended but I understand this may not be a priority.

1technophile:
We would need some testing to see if the restart from setupMQTT could be avoided, we will keep it for now as I have other PR in stock.
Thanks

PR #2025, opened 2024-08-26
Add protection for jsonQueue when emptying queue via emptyQueue()

puterboy:
## Description:
The jsonQueue is already protected during queue additions (via push) via the routine `handleJsonEnqueue`.
This adds protection for the complementary queue deletions via `emptyQueue` (via pop) since you probably don't want to be popping while another thread is pushing.

## Checklist:
  - [X ] The pull request is done against the latest development branch
  - [X] Only one feature/fix was added per PR and the code change compiles without warnings
  - [X] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).

Issue #2014, opened 2024-08-19
Several rtl_433 sensors are sporadically broadcasting a spike of nonsensical temperature or battery data after upgrade to v175


puterboy:
I have a bunch of cheap LaCrosse, AmbientWeather, and Nexus temperature & humidity sensors that I read using a LILYGO LORA 433 esp32 running OpenMQTT Gateway compiled under the lilygo-rtl_433 environment.

Everything was working properly until I upgraded to v175.
I started noticing that multiple of the sensors would sporadically (without any seeming pattern) transmit erroneous MQTT data sensor data.

For example, my 4 older LaCrosses sensors randomly broadcast a temperature of 2 degC (33.8 degF) regardless of the actual room temperature (note the erroneous value is always exactly 2 deg C). The newer ones are a different model and seem to be OK

Similarly, the battery level would go from a normal 100.0 (corresponding to 100%) to a non-sensical but value that seems different for each sensor (vs. temperature where 33.8 was always the error value). Examples include: 21385.0, 20989.0, 12079.0, 21682.0, 8218.0, 2575.0

In both cases, the erroneous spikes last for only a single reading before returning to normal and then potentially spiking again in an hour or two.

I can't see any pattern in the timing of the spikes or even which sensors are spiking (some of them seem to not be spiking or it could just be that I haven't observed them long enough since the update)

Reverting to the prior version (development branch with last commit on 3/5/23) removes this issue -- so it very much seems to be a problem with the latest code and not with my setup or sensors.

**To Reproduce**
1. Install v175 on a Lilygo LORA 433mHz board with the lilygo-rtl_433 build environment
2. Observe the temperature or battery values of a LaCrosse tx141thbv2 sensor for sudden nonsensical spikes

**Environment (please complete the following information):**
As above

puterboy:
The version that works (dev branch from March 2024) interestingly uses a newer version of the library rtl_433_ESP  (rtl_433_ESP.git#v0.3.2 than v175 (rtl_433_ESP.git#v0.3.1).

The only seemingly significant differences (beyond spelling corrections in comments) are as follows:
```
--- OpenMQTTGateway-old/.pio/libdeps/lilygo-rtl_433-jjk/rtl_433_ESP/src/rtl_433/r_api.c 2024-03-13 23:51:45.000000000 -0400
[...]

PR #2035, opened 2024-08-28
[SYS] Fixed data corruption associated with jsonQueue enqueueing and dequeuing

puterboy:
- Changed `jsonQueue` structure and associated routines to use json String rather than json Object
- Added Mutex protection to `emptyQueue()`
- Moved Mutex protection to within `enqueueJsonObject()` to significantly shorten lock time

## Description:
This fixes a **severe** bug whereby any MQTT messages enqueued on the jsonQueue were subject to unpredictable data corruption if the encoded json documents were more than one level deep since pushing/popping onto & from the queue do only shallow copies. This corruption could happen even with a queue size of one if the memory is overwritten.
In practice, this caused sporadic data MQTT message corruption for me as discussed in https://github.com/1technophile/OpenMQTTGateway/issues/2014

1. _NOTE YOU SHOULD ALSO BE ABLE TO REVERT THE REVERT THE CHANGE IN `ZgatewayRTL_433.ino` that called `pub` directly instead of using the JSON queue_ It's pretty likely that this PR fixes the cause of the instability that others experienced.
2. _NOTE THAT THIS PR SUPERSEDES https://github.com/1technophile/OpenMQTTGateway/pull/2025_

Specifically, the PR does the following:
- Changes the format of the queue (and the associated actions `enqueueJsonObject()` and `emptyQueue()`) to push/pull JSON strings rather than Documents onto the queue
- Extends the Mutex protection (xQueueMutex) that was previously applied only to enqueuing als to `emptyQueue()`
- Shortens the lock time by moving the Mutex for enqueueing into the  `enqueueJsonObject()` routine rather than `handleJsonObject()` since only the actual push needs to be protected while the deserializing and other actions don't

Note this leaves `handleJsonObject()` as a stub that just calls `eqnueueJsonObject()`.

## Checklist:

### DIFF
--- a/main/main.ino
+++ b/main/main.ino
@@ -607,7 +607,7 @@ void emptyQueue() {
   if (error) {
     Log.error(F("deserialize jsonQueue.front() failed: %s, buffer capacity: %u" CR), error.c_str(), jsonBuffer.capacity());
   } else {
-    pubMainCore(obj);
+    jsonDispatch(obj);
   }
   queueLengthSum++;
 }
--- a/main/main.ino
+++ b/main/main.ino
@@ -89,7 +89,7 @@ struct JsonBundle {
   StaticJsonDocument<JSON_MSG_BUFFER> doc;
 };
 
-std::queue<JsonBundle> jsonQueue;
+std::queue<String> jsonQueue;
 
 #ifdef ESP32
 #  include <driver/adc.h>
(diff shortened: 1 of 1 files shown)


puterboy:
Note now that you are emptying the queue with jsonDispatch that assumes an origin key, you could save some processes if you stored the origin and the rest of the json string on the queue and then called a new version of jsonDispatch that takes a topic and a json string. Perhaps push origin, then push string. Then pop string, then origin.
If you do this then the check for 'origin' should be when you enqueue and doesn't need to be done when you dequeue.

Currently `emptyQueue` serialize the json string and then jsonDispatch de-serializes it - so you could save this round-trip.

1technophile:
Great, thanks for this!

What do you think about using a char array instead of a String ?

puterboy:
That would of course be fine.

puterboy:
Please also look at my note above regarding creating a new version of `jsonDispatch` to avoid having to serialize and deserialize.

In that vein, it may make sense to create a lower level enqueue that took a pair of strings: origin, jsonString and push them successively on the queue. Perhaps call it `enqueueJsonString`

Then a higher level enqueue routine could take just a jsonDocument, extract & remove the origin key & value, serialize the remaining, and call the lower-level enqueue routine. That higher level routine would be the only that would need the error checking to see if origin key exist.

For emptyQueue, after extracting the origin and content strings, they would be sent to a version of `jsonDispatch` that accepts such a pair. Note: I think `emptyQueue` is a bit of a misnomer since it really only removes one element from the queue. Maybe it should be called `dequeueJsonString`

Also you might want to remove the now unnecessary `handleJsonEnqueue` wrapper - you may though want to create an overloaded version of `enqueueJsonObject` (or whatever it will be called) that doesn't require the timeout.

I am really getting beyond my skill level here - so now that I have a POC working, would you mind taking it from here to polish?

1technophile:
> would you mind taking it from here to polish?

Yes sure, I would need your help to retest after by the way

puterboy:
Of course!
Great platform! And kudos for all the work you have put into it...

1technophile:
I will merge this one and start from there.
Thanks!

PR #2037, opened 2024-08-29
Changed from GET to POST method for config forms with passwords

puterboy:
## Description:
Passwords on MTT, WiFi and Gateway config pages were being sent in plaintext in broswer address bar.
Fixed my changing to (and requiring) POST method for those pages.

I let the other pages using GET since they have no sensitive info but feel free to modify those too if you want

## Checklist:
  - [x ] The pull request is done against the latest development branch
  - [x ] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x ] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).


1technophile:
Thanks

PR #2053, opened 2024-09-11
Modify gatewayRF auto discoverability

JamiePhonic:
## Description:
Pass `switchRF[0]` as the subtype argument to `announceDeviceTrigger()` which will cause recieved codes to be picked up by home assistant as triggers, i.e. When setting up automations in Home Assistant, selecting a `device trigger` and an OMG device will allow the user to select a code picked up during the autoDiscover window as the trigger for the automation, for example `"1394004" pressed`

Also change `getUniqueId()` call to remove leading '-' which will prevent MQTT topics having '--' in them

## Checklist:


JamiePhonic:
Closing becasue this needs some more work. Triggers seem to be inconsistent.

Issue #2064, opened 2024-09-24
BME280 not working (Bosch BME280/BMP280 failed: 0x0) - Heltec v2 running OMG heltec-rtl_433-fsk


mcmlxxvii:
I'm facing an issue with a BME280 and Heltec v2 (heltec-rtl_433-fsk development). I'm using this one: https://www.amazon.ca/dp/B07KYJNFMD 

That sensor is an I2C only version that works both in 3.3v and 5v, as it has a built-in voltage regulator. Mine is connected to 3.3v though.

I tested 3 different sensors (using both 0x76 and 0x77 addresses) with my Heltec v2 board, using the recommended pinout in OMG's documentation: 2 of the [aforementioned model](https://www.amazon.ca/dp/B07KYJNFMD) and 1 of a [similar model](https://www.amazon.ca/dp/B0CXF89ZYQ) which has 6 pins and is a hybrid I2C/SPI version. All of them show the same behavior: _Bosch BME280/BMP280 failed: 0x0_ and weird readings, preceded by _[E][Wire.cpp:513] requestFrom(): i2cRead returned Error 263_ error message in the serial console.

I tested all those sensors with an ESP-WROOM-32 board using the exact same wiring, voltage and pinout. For that, I leveraged an I2C Scanner Sketch to make sure of the sensor's address and a BME280 Testing Sketch using both SparkFun (2.0.9) and Adafruit (2.2.4) libraries in Arduino IDE.

Everything worked as expected using the ESP-WROOM-32. However, when trying the exact same sensors with a Heltec v2 running OMG heltec-rtl_433-fsk compiled via Gitpod.io, they don't work.

You can see some console logs illustrating the issue below:

```
rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:1
load:0x3fff0030,len:1184
load:0x40078000,len:13160
load:0x40080400,len:3036
entry 0x400805e4
N:
************* WELCOME TO OpenMQTTGateway **************
N: SYS config not found
N: SSD1306 config initialised
N: Saved SSD1306 config loaded
N: Setup SSD1306 Display end
N: OpenMQTTGateway Version: 20240918-01
[...]

PR #2066, opened 2024-09-24
[DISC] Warning message for discovery prefix change

NorthernMan54:
## Description:

I did a refresh this AM, as part of investigating #2064, and found that my Homebridge environment was filled with rtl_433 devices.  Went back and looked at the recent changes, and saw the compiler directive change from `discovery_Topic` to `discovery_Prefix` ( which is a better directive name ), and saw that the old directive was being overwritten with 'homeassistant', rather than being backwardly compatible with a warning.

## Proposed fix

1 - Add backward compatibility on the directive `discovery_Topic`
2 - Add a compiler warning when the backward compatibility is invoked. So that others are not impacted as I was.

PS I see that we are using the complier option '-w' which disables warnings, which prevents usage of `#warning`, hence my usage of `#  pragma message("compiler directive discovery_Topic is deprecated, use discovery_Prefix instead")`.  Maybe we should enable warnings, and fix the warnings.  I did see a lot of warnings when compiling with -w removed

PSS -  ZgatewayRTL_433 -> Log.notice(F("type: %s" CR), type.c_str());  I commented this out as it just fills the log with type: null messages

## Checklist:

### DIFF
--- a/main/ZgatewayRTL_433.ino
+++ b/main/ZgatewayRTL_433.ino
@@ -281,7 +281,7 @@ void rtl_433_Callback(char* message) {
   String topic = subjectRTL_433toMQTT;
   String model = RFrtl_433_ESPdata["model"];
   String type = RFrtl_433_ESPdata["type"];
-  Log.notice(F("type: %s" CR), type.c_str());
+  // Log.notice(F("type: %s" CR), type.c_str());
   String uniqueid;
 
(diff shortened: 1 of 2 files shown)


1technophile:
Thanks for the PR

>  ZgatewayRTL_433 -> Log.notice(F("type: %s" CR), type.c_str());

We can remove it I think

NorthernMan54:
@1technophile Removed -  Log.notice(F("type: %s" CR), type.c_str());

PR #2083, opened 2024-10-08
Implement changes to build against ArduinoJson@7.2.0. \n

peshovec:
In ArduinoJson@6.20 a deprecation (naked char) is enforced. The deprecation origins from ArduionoJson@6.18.0

## Description:
The proposed changes implement the described in 6.18.0 way to use - namely using const char, instead of char.

## Checklist:


1technophile:
Thanks

PR #2084, opened 2024-10-08
[RF] Disable the active RF receive when transmitting with RC-Switch and "Kaku"

peshovec:
Pilight RF transmit is behaving this way already.

## Checklist:

### DIFF
--- a/main/ZgatewayRF.ino
+++ b/main/ZgatewayRF.ino
@@ -236,8 +236,8 @@ void XtoRF(const char* topicOri, JsonObject& RFdata) { // json object decoding
       Log.notice(F("RF Protocol:%d" CR), valuePRT);
       Log.notice(F("RF Pulse Lgth: %d" CR), valuePLSL);
       Log.notice(F("Bits nb: %d" CR), valueBITS);
-#    ifdef ZradioCC1101
       disableCurrentReceiver();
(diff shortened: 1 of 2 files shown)


1technophile:
Thanks, was this tested with emitting and receiving for the different modules ?

peshovec:
Right now, I am using with this change applied.
Did not run through tests, however this is what I can observe:
The gateway is listening on pilight.
 Can send mqttto433 (trigger my aqua light)
 Can send mqtttorf2 (trigger power outlets)
 The sended from the gateway codes are not catched by the pilight receiver.
 Radio emmited codes (with remote controls) are catched by the gateway
(pilight receiver)

Should I replicate that with the other two receiver modes? If i have to,
will take some time, as I run some longer test right now (and switching to
rf2 will require restart, which i'd like to avoid now)

1technophile:
> Should I replicate that with the other two receiver modes? 

Yes if possible, As the decoder change can be sensible it is necessary to do a regression testing

peshovec:
Run the tests, results are ok. Details below.

Puted the gateway in RF receive mode. Checked that external signals are
received. Issued few commands to send codes from the gateway. Rechecked
that external signals are still received.

Did the same with RF2 receiver mode

By external signals I mean, pressing RF remote control buttons.

PR #2088, opened 2024-10-13
[BOARD] add esp32s3-m5stack-stamps3

mhaberler:
https://docs.m5stack.com/en/core/StampS3

## Checklist:

### DIFF
--- a/environments.ini
+++ b/environments.ini
@@ -1805,6 +1805,7 @@ build_flags =
   '-DCONFIG_ESP_CONSOLE_UART=ESP_CONSOLE_USB_SERIAL_JTAG'
 
 [env:esp32c3-m5stamp]
+; https://docs.m5stack.com/en/core/stamp_c3 
 platform = ${com.esp32_platform}
 board = esp32-c3-devkitm-1
 board_build.partitions = min_spiffs.csv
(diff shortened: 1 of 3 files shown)


environments.ini
```diff
@@ -1654,6 +1654,24 @@ build_flags =
...
+[env:esp32s3-m5stack-stamps3]
+platform = ${com.esp32_platform}
+board = esp32-s3-devkitc-1
+lib_deps =
+  ${com-esp32.lib_deps}
+  ${libraries.ble}
+  ${libraries.decoder}
+  ${libraries.adafruit_neopixel}
+build_flags =
+  ${com-esp32.build_flags}
```
1technophile, line 1666: Could you also add TRIGGER_GPIO so that the board user settings could be erased with a long push, here is an example:
https://github.com/1technophile/OpenMQTTGateway/blob/3b3fe3eabed675b51e96d75e5edabe897c35c7f4/environments.ini#L1802

Not sure if it is the same pin for the S3 though

mhaberler: let me check

mhaberler: the M5Stamp C3 and C3U have two buttons:
the C3U has a usable button on G9 : https://docs.m5stack.com/en/core/stamp_c3u
the C3 has a usable button on G3 : https://docs.m5stack.com/en/core/stamp_c3
the other button on C3 and C3U is hard reset 

the M5Stamp S3 **only** has a hard reset button on G0 : https://docs.m5stack.com/en/core/StampS3 

so - no, as no free button available and G0 didnt work for me as it immediately resets

1technophile:
Could you add also the environments to https://github.com/1technophile/OpenMQTTGateway/blob/development/.github/workflows/build.yml

This way they will be part of the CI per PR

mhaberler:
sorry, overlooked. will do

1technophile:
Thanks

PR #2117, opened 2024-11-18
HA: RF modules discovery as sensors. Subscribe topics, accordingly to (not) defined valueAsATopic.

peshovec:
The RF sensors discovery are created with topics like "/433toMQTT" 
If valueAsATopic is defined, then published messages are not in the main topic, but in subtree.
Change the topic to be like "/433toMQTT/#" , when valueAsATopic is defined.

## Checklist:

### DIFF
--- a/main/ZmqttDiscovery.ino
+++ b/main/ZmqttDiscovery.ino
@@ -1075,7 +1075,12 @@ void pubMqttDiscovery() {
 
   //trc(gatewayRF[1]);
   createDiscovery(gatewayRF[0],
-                  subjectRFtoMQTT, gatewayRF[1], (char*)getUniqueId(gatewayRF[1], gatewayRF[2]).c_str(),
+                #if valueAsATopic
+                  subjectRFtoMQTTvalueAsATopic,
+                #else
+                  subjectRFtoMQTT,
+                #endif
+                  gatewayRF[1], (char*)getUniqueId(gatewayRF[1], gatewayRF[2]).c_str(),
                   will_Topic, gatewayRF[3], gatewayRF[4],
                   gatewayRF[5], gatewayRF[6], gatewayRF[7],
                   0, Gateway_AnnouncementMsg, will_Message, true, "",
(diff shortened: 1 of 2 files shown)


DigiH:
Hi @peshovec 

Right clicking into the code files and choosing **Format Document** should get rid of your Code Format error.

Odyno:
Hi @DigiH, why not use the real value on the topic declaration instead of using the "#" on the  https://github.com/1technophile/OpenMQTTGateway/pull/2117/files#diff-8394cc3dba8dcaebfceaadd7e670e46424b3873ce5db0dfcfd74efd9741f89d5R149 and successive?

peshovec:
@DigiH 10x - I am using sometimes vscodium , so nice hint. In this particular case it will reformat the whole file, isn't it? https://github.com/1technophile/OpenMQTTGateway/actions/runs/11890004537/job/33127729802?pr=2117 . Probably it will be better the reformat to be separate "cosmetic" commit.

@Odyno If you mean this:
#define subjectRFtoMQTT       "/433toMQTT"
#define subjectRFtoMQTTvalueAsATopic       "/433toMQTT/#"
The second one is for the case, when valueAsaTopic is declared - In this case publishing will be like that /433toMQTT/DifferentCodeHere/values, or even /433toMQTT/DeviceCode/UnitCode/values . In any case the topic will vary according the source device. Technically the "#" can be replaced with "+" (or series "+/+") However as i understood these topics are for diagnostic information. I use them just to be sure, that something is received.

Just to add for clarify: when valueAsATopic is not defined, publishing is like /433toMQTT/value - and in the value there is information for the device, unit and code.

DigiH:
I only really wanted to mention the formatting option above 😉 

Someone else will look at and review the functional size of the code, but only really once all checks have passed, including the formatting.

> In this particular case it will reformat the whole file, isn't it?

As the whole document has already been correctly formatted, as defined in **clang-format**, formatting it again with these rules would only really affect and correct your newly added code.

peshovec:
Understood. Seems like I misunderstood the failed check log. The detailed
one.
Will edit and PR new one. And delete this one aftereards

peshovec:
reformated and submited new PR

peshovec:
reformated and submited new PR

PR #2129, opened 2024-12-10
[Pilight] Handle when 'value' is not a string

melyux:
## Description:
I noticed when I was using the `iwds07` protocol with Pilight on OMG that `valueAsATopic` wasn't working properly. Instead of creating a topic for the device ID, it would send it on the `PilighttoMQTT` topic directly. While the Pilight module in OMG has code to handle this (`iwds07` uses "unit" instead of "id" for the ID), it didn't handle when "unit" was an integer instead of a string. This PR handles that.

## Checklist:


1technophile:
Thanks!

PR #2130, opened 2024-12-11
Allow multiple RF modules simulatenously

melyux:
## Description:
So this is how I've been successfully using the Pilight and RF modules simultaneously on the same OMG installation. I have a normal RXB6 receiver, and its same data pin goes to 2 different pins on the ESP8266 (presumably to take advantage of the hardware interrupts thereof).

The code change is **dead simple**: just enable all the RF modules you want instead of only enabling one.

I use a bitfield for setting the active receiver (0000), so you can set multiple receivers at the same time with e.g. 0011, meaning {"active": 3} for Pilight + RF. Do this using the usual active receiver setting thing (e.g. publish `{"active": 3}` to `rf/OpenMQTTGateway/commands/MQTTtoRF/config`).

There are obviously some things that might not work that I haven't tried, like CC1101 using Pilight + RF or HA discovery, but I'm just a simple man, I use the old fashioned RXB6 variants for most OOK things like this. The code will need adjustment and sanity checks like this. Since I'm not sure how you'd want to implement configurations and etc, this is a draft, but it works perfectly for my purposes, so I wanted to share it in hopes of getting a variation of it merged for others' benefit. (see #450)

Thanks @1technophile for your efforts and openminded kindness through the years.

## Checklist:

### DIFF
--- a/docs/use/rf.md
+++ b/docs/use/rf.md
@@ -39,12 +39,12 @@ Note that with CC1101 this frequency will be used as the default sending frequen
 
 Switching of the active transceiver (RTL_433 receiver only) module is available between the RF, RF2, and (RTL_433 or Pilight) gateway modules, allowing for changing of signal decoders without redeploying the OpenMQTTGateway package.  Sending a JSON message to the command topic will change the active transceiver module.
 
-To change the RF gateway module, which will receive, send a json message to the RF gateway module command subject (home/OpenMQTTGateway/commands/MQTTtoRF/config) with the corresponding value of the key "active" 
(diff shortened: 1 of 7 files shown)


melyux:
Someone in #450 mentioned, why can't this be done in software instead of in hardware like we're doing here? And the more I think about it, it seems like it should be possible. It's not like the receiver creates a new interrupt for each module (it has no idea if a signal is Pilight or RF2 after all), so it should theoretically be possible to do handle a single hardware interrupt and decode it with multiple modules.

melyux:
Hmm @1technophile?

1technophile:
Hey @melyux, let's keep it open. We can use it as a base once @h2zero gets the refactoring merged

PR #2131, opened 2024-12-11
Update ZgatewaySERIAL.ino Fix tx issues

mycprotein:
The serial gateway can only receive serial data and publish via mqtt, but can not convert mqtt into serial data. It is because the receiverReady is not handled correctly for the tx part. A quick work around is to set receiverReady to true. (it seems not used)

## Checklist:
  - [ ] The pull request is done against the latest development branch
  - [ ] Only one feature/fix was added per PR and the code change compiles without warnings
  - [ ] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).

### DIFF
--- a/main/ZgatewaySERIAL.ino
+++ b/main/ZgatewaySERIAL.ino
@@ -48,12 +48,17 @@ const TickType_t semaphoreTimeout = pdMS_TO_TICKS(1000); // 1 second timeout
 #    define SEMAPHORE_SERIAL_GIVE
 #  endif
 
+#  ifdef SENDER_SERIAL_HEARTBEAT
+bool receiverReady = false;
+#  else
+bool receiverReady = true;
+#  endif
+
 // use pointer to stream class for serial communication to make code
 // compatible with both softwareSerial as hardwareSerial.
 Stream* SERIALStream = NULL;
 //unsigned long msgCount = 0;
 
-bool receiverReady = false;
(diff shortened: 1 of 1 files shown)


1technophile:
Do you have `SENDER_SERIAL_HEARTBEAT` defined ?

mycprotein:
It is not mentioned in the header file, so I initially thought it was still under construction. I recommend updating and commenting on it in the header file. Otherwise, many users may struggle to understand and utilize the tx function effectively.

1technophile:
We could condition `bool receiverReady = true;` to the non definition of macro `SENDER_SERIAL_HEARTBEAT`

mycprotein:
Sounds great

1technophile:
> We could condition bool receiverReady = true; to the non definition of macro SENDER_SERIAL_HEARTBEAT

@mycprotein would you like to update the PR to do this ?

mycprotein:
not sure if it will work

1technophile:
> not sure if it will work

Why do you think so ?

mycprotein:
I'm not a C++ programmer, so I'm not entirely sure if my modification is correct. However, it seems right to me. If it makes sense to you, then it should be fine.

1technophile:
Please validate the PR checkboxes and we are good to merge

PR #2133, opened 2024-12-14
[DISC] Whitelist immediate change

benzok:
Today when adding a BLE device that is auto discovered you need to temporarily disable the white-list first and then add the new device to the list, as described in the [docs.](https://docs.openmqttgateway.com/use/ble.html#setting-a-white-or-black-list) 

With this change you just update the list and it gets discovered. (You need to restart the gateway if you have not enabled discovery)

See discussion in [Theengs community](https://community.openmqttgateway.com/t/why-do-we-need-to-temporarily-disable-the-white-list-when-adding-a-new-entry/3584)

The PR has been built and tested on environment `esp32-lolin32lite-ble`

## Checklist:
  - [ x] The pull request is done against the latest development branch
  - [ x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [ x] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).

### DIFF
--- a/docs/use/ble.md
+++ b/docs/use/ble.md
@@ -124,6 +124,12 @@ A white list is a list of MAC addresses permitted to be published by OMG
 to set white list
 `mosquitto_pub -t home/OpenMQTTGateway/commands/MQTTtoBT/config -m '{"white-list":["01:23:14:55:16:15","4C:65:77:88:9C:79","4C:65:A6:66:3C:79"]}'`
 
+to temporarily disable white/black list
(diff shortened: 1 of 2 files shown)


DigiH:
Thanks @benzok , also for amending the documentation, but the optional commands how to temporarily disable or enable white/black lists need to remain, as these are still useful working commands.

Just the top paragraph about this being required for new white-list discoveries can go.

1technophile:
Thanks

PR #2134, opened 2024-12-17
HA: RF, RF2, Pilight modules as sensors. Subscribe topic, according valueAsATopic.

peshovec:
## Description:
The RF sensors discovery are created with topics like "/433toMQTT"
If valueAsATopic is defined, then published messages are not in the main topic, but in subtree.
Change the topic to be like "/433toMQTT/#" , when valueAsATopic is defined.

## Checklist:

### DIFF
--- a/main/ZmqttDiscovery.ino
+++ b/main/ZmqttDiscovery.ino
@@ -1083,7 +1083,12 @@ void pubMqttDiscovery() {
 
   //trc(gatewayRF[1]);
   createDiscovery(gatewayRF[0],
-                  subjectRFtoMQTT, gatewayRF[1], (char*)getUniqueId(gatewayRF[1], gatewayRF[2]).c_str(),
+#    if valueAsATopic
+                  subjectRFtoMQTTvalueAsATopic,
+#    else
+                  subjectRFtoMQTT,
+#    endif
+                  gatewayRF[1], (char*)getUniqueId(gatewayRF[1], gatewayRF[2]).c_str(),
                   will_Topic, gatewayRF[3], gatewayRF[4],
                   gatewayRF[5], gatewayRF[6], gatewayRF[7],
                   0, Gateway_AnnouncementMsg, will_Message, true, "",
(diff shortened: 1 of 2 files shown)


Odyno:
Hi @peshovec 
I’m sorry, but it’s not clear to me why we need to provide "/433toMQTT/#" instead of the actual value like "/433toMQTT/123123". I agree that the PR is an improvement, but why do you prefer to keep it generic?

P.S. One small note: it seems that you’ve updated RF, RF2, and Pilight, not just RF. It might be helpful to include all of them in the PR title to simplify future searches.

peshovec:
No worries:) title is edited now.

The HA uses this topics to subscribe to them and to collect "sensor" information. At least i understood it this way. I agree, this information is mainly for diagnostic purposes.

The point is that, when valueAsATopic is defined, then the OMG publish like this:

## Add the received "value" at the end of the topic
For the gateways that publish a "value" parameter on the json (RF, IR...), it is possible to concatenate this parameter at the end of the topic.
So as to activate this function you need to set `valueAsATopic` to true in User_config.h

``` cpp
#ifndef valueAsATopic
#  define valueAsATopic false // define true to integrate msg value into the subject when receiving
#endif
```

Per default you see the following data:
```
home/OpenMQTTGateway/SRFBtoMQTT {"raw":"2B660186042E00E7E5","value":"59365","delay":"1111","val_Thigh":"390","val_Tlow":"1070"}
```

With the line uncommented you will see:
```
home/OpenMQTTGateway/SRFBtoMQTT/59365 {"raw":"2B660186042E00E7E5","value":"59365","delay":"1111","val_Thigh":"390","val_Tlow":"1070"}
```
As i understood - the value (in this example 59365) is not fixed, and changes for the various devices (transmitters)
Took the example from the docs, as i am not able now to get ones from my setup. If i recall correctly, in my setup i even observed subtopics, two leve deep than .../MQTT. Hence i decide to "instruct" HomeAssistant to subscribe to the whole subtree (#, instead of +, +/+)

Odyno:
Ok, all of this was clear to me, but I finally understood where things don't add up for me. In short, I'm not sure about rf2 or Pilight, but for RF this PR is only useful if two discovery messages continue to be sent for the same RF signal.
I believe there is a hypothetical issue with the RF module (RCSwitch), tracked under #2139

1technophile:
Thanks

PR #2136, opened 2024-12-17
fix stacksize in procBLETask

Lukaswnd:
add OVERRIDE_PROC_BLE_TASK_STACK_SIZE as value for custom stacksize in the procBLETask

## Description:
When using Arduino and ESP32s3-devkit-c1 as a BLE gateway the stackwatchpoint gets triggered, when it scans for devies and gets triggered at the same time. I just tested it with 13500 and it works fine. I think a define, which the user can set in the build_flags is the best option to keep it flexible

## Checklist:
  - [x ] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).


mhaberler:
I needed a similar patch

PR #2137, opened 2024-12-17
add dynamic stacksize for procBLETask

Lukaswnd:
add OVERRIDE_PROC_BLE_TASK_STACK_SIZE to set the stacksize for the procBLETask by a define in the build_flags

## Description:
When using the ESP32s3-devkit-c1 with Arduino and using it as a ble gateway, the task watchdog for the stacksize gets triggerd, if you receive a request the same time it scans for new devices. This results in a crash. I testes it with 13500 and it works fine. I think a build flag which each user can set seperatly but doesn't have to is the best option to solve it.

## Checklist:

### DIFF
--- a/main/ZgatewayBT.ino
+++ b/main/ZgatewayBT.ino
@@ -852,6 +852,8 @@ void setupBTTasksAndBLE() {
       "procBLETask", /* Name of the task */
 #  if defined(USE_ESP_IDF) || defined(USE_BLUFI)
       13500,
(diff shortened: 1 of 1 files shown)


1technophile:
Would be interesting to adjust the stack size from the 8500 low limit, I would prefer to increase this one rather than going directly to the 13500.

If you put the log level to trace you will have a view of how much stack is remaining at the end of the scan. You may want to add other trace but this first log should give you an idea.

1technophile:
Closing, task size increased

PR #2140, opened 2024-12-19
[BT] Update NimBLE-Arduino to 2.2.1

h2zero:
## Checklist:


h2zero:
Ready for testing if anyone is interested.

peshovec:
Hi, i tried to run some test with this PR applied.
Should it be used with framework = arduino, espidf ? I ask, as in my setup i did not succeed to test with idf (i use manual wifi configuration and hit some duplicate names of defines and struct ...)

This question is raised, as my normal tests (without specifying framework) are not stable. That's why i decide to try predefined configured environments and saw env:esp32dev-ble-idf..

h2zero:
If using the ` arduino, espidf` framework then the `esp-nimble-cpp` library should be used instead of `NimBLE-Arduino`

peshovec:
10x, had to modify the define wifi_ssid (as i use manual wifi config and the idf version have wifi_ssid named structure) and was able to test it too.

So my observation (both with Nimble-Arduino and esp-nimble-cpp ) with the PR applied are not good. My esp keeps rebooting.
From the mqtt messages, it seems (in both cases), that somehow is related to the id field. :
```
{"id":",��?","mac_type":1,"adv_type":0,"rssi":-95,"distance":35.51099766,"servicedata":"04466072a6ae2125512f23942471cc8d628b93d5b0","servicedatauuid":"0xfcf1"}
````

```
{"id":"␦��\n�6XJ�a��?T","mac_type":1,"adv_type":0,"rssi":-93,"distance":30.15532078,"servicedata":"04accc85a0b4e4d7523ef2ffd6b84b39bae8825335","servicedatauuid":"0xfcf1"}
```

Happy to run more detailed test (in the weeknd if physical access to the esp will be needed)

peshovec:
Turns out, it was my setup (probably some compile caches), now after full clean and build it is working ok (no reboots so far).
After some time (right now it is the idf version), will be able to confirm for the nimble too. (i am pretty sure it will work, as in my setup i use arduinojson 7.3.1 and some refactoring is needed when extracting json fields)

edit: tested with Nimble-Arduino too - it is ok too, working as expected.

h2zero:
Cool, thanks for the feedback!

1technophile:
Thanks!

PR #2153, opened 2025-01-18
[DISC] Removal of openHAB specific MQTT discovery patch

DigiH:
Removal of openHAB specific MQTT discovery patch "ohdisc", as it is no longer required with openHAB 4.3 and later, and with openHAB before 4.3 MQTT discovery never was fully implemented anyway.

## Checklist:

### DIFF
--- a/docs/integrate/openhab3.md
+++ b/docs/integrate/openhab3.md
@@ -76,13 +76,6 @@ Or use another upload method from the Upload section that requires credentials b
 
 You should see the gateway connected to your broker by checking with an MQTT client like MQTT Explorer.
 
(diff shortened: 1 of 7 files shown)


1technophile (commented):
You could also remove:
```
// Set the line below to true so as to have autodiscovery working with OpenHAB
#ifndef OpenHABDiscovery
#  define OpenHABDiscovery false
#endif
```
From config_mqttDiscovery.h

DigiH:
> You could also remove:

True, I missed that one. thanks. Done

1technophile:
Thanks

PR #2158, opened 2025-02-03
[RF] GatewayRF2: fix json formatted sending, when the active receiver is not RF2

peshovec:
## Description:
Catch that recently. 
I was using CC1101 module and active receiver RTL. When sending (via mqtt publisher) RF2 code the board "goes in guru meditation".
The proposed commit fix the issue.

## Checklist:

### DIFF
--- a/main/ZgatewayRF2.ino
+++ b/main/ZgatewayRF2.ino
@@ -248,7 +248,7 @@ void XtoRF2(const char* topicOri, JsonObject& RF2data) { // json object decoding
     int boolSWITCHTYPE = RF2data["switchType"] | 99;
     bool success = false;
     if (boolSWITCHTYPE != 99) {
-      NewRemoteReceiver::disable();
+      disableCurrentReceiver();
       pinMode(RF_EMITTER_GPIO, OUTPUT);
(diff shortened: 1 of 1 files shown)


main/ZgatewayRF2.ino
```diff
@@ -283,7 +282,6 @@ void XtoRF2(const char* topicOri, JsonObject& RF2data) { // json object decoding
           }
         }
         Log.notice(F("MQTTtoRF2 OK" CR));
-        enableActiveReceiver();
```
1technophile, line 286: I'm not sure removing this is not going to affect other combinations, like receiving with RF sending with RF2. Did you tested it with the others combinations of RF receivers and senders?

peshovec: Yes, tests were done.
Just below the code fragment, there is if (success) branch, and right after it there is enableActiveReceiver(). Should be at line 297 / 299.

I am not exactly sure which one of the placements is better. The proposed to delete at line 286, or the the one at line 297 / 299.

1technophile: Make sense, thanks

PR #2159, opened 2025-02-07
[RF] ZgatewatPilight: toggle RawEnabled only if Pilight is active

peshovec:
## Description:
When enabling or disabling RawEnabled for Pilight gateway the PilightGateway is deactivated and activated.
The proposed change the rawenabled, only if current active receiver is Pilight. Trying to avoid situation such as: Active receiver is RTL , toggling RawEnabled will finally enable Pilight receiver and potentially there will be "race" between two active receivers, or potentially not "synchronized" changes to the radio module (RTL using RadioLib, Pilight using ELECHOUSE).

If accepted, the documentation regarding Pilight gateway, particular using rawenabled should clarify, that toggling is possible only if Pilight is the active receiver. And that if Pilight is not the active receiver, the mqtt error message when trying to toggle is normal. 

## Checklist:

### DIFF
--- a/docs/use/rf.md
+++ b/docs/use/rf.md
@@ -198,7 +198,10 @@ Generate your RF signals by pressing a remote button or other and you will see :
 #### Enabling RAW packet return support
 First, you need to compile a binary with `Pilight_rawEnabled true` uncommented in config_RF.h.
 
-Once the device is online, you can turn on the RAW packet return support with the following MQTT command:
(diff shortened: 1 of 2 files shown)


1technophile:
Do you mind updating the documentation as part of this PR ?
It will make it easier to track the changes

peshovec:
Ok, pushed.

1technophile:
Thanks

PR #2160, opened 2025-02-10
Support lilygo-t3-s3

TomRoush:
## Description:
Adds support for the `lilygo-t3-s3` with a SX1276: [https://lilygo.cc/products/t3s3-v1-0?variant=42740446298293](https://lilygo.cc/products/t3s3-v1-0?variant=42740446298293). Tested with the 915MHz board with the FSK encoding version, and a WH51 moisture sensor. This also includes a bump to `espressif32@6.10.0` to get the board defintion

## Checklist:


platformio.ini
```diff
@@ -176,7 +178,7 @@ monitor_speed = 115200
 
 [com]
 esp8266_platform = espressif8266@4.2.1
-esp32_platform = espressif32@6.1.0
+esp32_platform = espressif32@6.10.0
```
Odyno: 🙆 I never realized that we have a version from March 8, 2023, for Espressif32... maybe there's a reason for that.

TomRoush: Looks like it breaks going from `6.7.0` to `6.8.0`

1technophile: > maybe there's a reason for that

Yes, we need to do an extensive testing of the BLE controller before going further. I would suggest to differentiate the PR between the platform evolution and the new board.
Does 6.7.0 supports your board ?

peshovec: I use 6.10.0 in my builds. BT seems working (no need for a patched library).
Where is the appropriate place to share and discuss the topic?

TomRoush: > > maybe there's a reason for that
> 
> Yes, we need to do an extensive testing of the BLE controller before going further. I would suggest to differentiate the PR between the platform evolution and the new board. Does 6.7.0 supports your board ?

It was added in 6.10.0 unfortunately. I'm fine with leaving this in draft until the platform version is bumped

1technophile:
@TomRoush Do you want to rebase the PR now that we have bumped version ?

Also I would suggest to add the envs as `;default_envs = ` in platformio.ini

1technophile:
Thanks

PR #2165, opened 2025-02-27
Features/rf refactoring

Odyno:
## Description:
Following the idea exposed on https://community.openmqttgateway.com/t/convert-arduino-file-to-c/3724 this is my proposal to start abandoning the ino files: I performed a refactory of the ZCommonRF.ino and ZGatewayRF.ino so that they are encapsulated classes with minimal external dependencies. 
In order not to distort too much, I left the function names as they were and left all the .h configuration already in place so that the user can configure them as described in the documentation.

Missing: the other drives used for the RF module (only rcswitch is done here)

I look forward to comments here or in the post https://community.openmqttgateway.com/t/convert-arduino-file-to-c/3724

## Checklist:
 

  - [x]  The pull request is done against the latest development branch
  - [ ] Only one feature/fix was added per PR and the code change compiles without warnings


1technophile:
Thanks for the proposals @Odyno, @h2zero is working on extending this to the whole codebase

Odyno:
@1technophile and @h2zero I made several changes on this PR but the real problem is LD... I stopped moving all the other RF modules just for this reason.... do you have solution for this kind of problems? these days I will push a version where I will put all the RF part as an external library... but I do not think this solves

Issue #2121, opened 2024-11-24
Sensors which do not response for a long time are not marked as "unavailable"


pointhi:
**Describe the bug**
When a sensor is no longer available (e.g. due to dead battery) Home Assistant is still reporing the current state as if it is a correct reading.

**To Reproduce**
1. Publish sensor reading
2. Remove e.g. battery from the sensor
3. Home assistant is showing the old reading even days later

**Expected behavior**

Expire sensor readings after some (configurable?) time when no new message arrives. See https://www.home-assistant.io/integrations/sensor.mqtt/#expire_after

**Environment (please complete the following information):**
* OpenMQTTGateway version used: v1.8.0

NorthernMan54:
Openmqttgateway does not maintain state for any sensors sending data thru it.  This is an issue for the application receiving the sensor data and not omg.

pointhi:
The option `expire_after` I shared here is not a state Openmqttgateway needs to keep but an option which tells HA when to expire messages.

Nealtron:
I believe this can be set via MQTT Discovery topics. 

The included [rtl_433_mqtt_hass.py](https://github.com/merbanan/rtl_433/blob/master/examples/rtl_433_mqtt_hass.py) auto discovery script provided by [rtl_433](https://github.com/merbanan/rtl_433/) has a flag (-x) that sets a single value for the expiration of all data. 

```
    parser.add_argument("-x", "--expire-after", type=int,
                        dest="expire_after",
                        help="Number of seconds with no updates after which the sensor becomes unavailable")
```



```
    if args.expire_after:
        config["expire_after"] = args.expire_after

    logging.debug(path + ":" + json.dumps(config))

    mqttc.publish(path, json.dumps(config), retain=args.retain)

    return True

```

PR #2178, opened 2025-04-13
Make measurement sensor expiry configurable

peterg79:
## Description:
Following up on https://github.com/1technophile/OpenMQTTGateway/issues/2121, I'd like to introduce a configurable expiration  for measurement sensors. Default is 0 which means no expiry, to keep it backward compatible.

## Checklist:


1technophile:
Thanks for the PR.

I'm worried that we have sensors that publish sporadically for which this settings may not be relevant.
I still think it is interesting for most of the sensors but for those who are not regular publishers we should propose a bypass rather than alerting HA of something being wrong but is not wrong.
Maybe we could have a setting in the WebUI to configure this expire after function, it could be configured in minutes, if the user set it to 0 we don't send an expire_after during discovery.

peterg79:
Thanks for the feedback! I'll do the change accordingly.

peterg79:
Change complete and verified, please review.

main/config_WebContent.h
```diff
@@ -130,7 +130,7 @@ const char config_logging_body[] = body_header
 #endif
     "</select></p><br><button name='save' type='submit' class='button bgrn'>Save</button></form></fieldset>" body_footer_config_menu;
 
-const char config_webui_body[] = body_header "<fieldset class=\"set1\"><legend><span><b>Configure WebUI</b></span></legend><form method='get' action='wu'><p><b>Display Metric</b><br><input id='dm' type='checkbox' %s></p><p><b>Secure WebUI</b><br><input id='sw' type='checkbox' %s></p><br><button name='save' type='submit' class='button bgrn'>Save</button></form></fieldset>" body_footer_config_menu;
+const char config_webui_body[] = body_header "<fieldset class=\"set1\"><legend><span><b>Configure WebUI</b></span></legend><form method='get' action='wu'><p><b>Display Metric</b><br><input id='dm' type='checkbox' %s></p><p><b>Secure WebUI</b><br><input id='sw' type='checkbox' %s></p><p><b>Expire After</b><br><input id='ea' name='ea' placeholder=" EXPIRE_AFTER " value='%u'></p><br><button name='save' type='submit' class='button bgrn'>Save</button></form></fieldset>" body_footer_config_menu;
```
1technophile, line 133: Could you add the unit used after  "Expire After" like  "Expire After (s)"

1technophile:
Thanks, could you add an explanation in the docs please:
https://docs.openmqttgateway.com/integrate/home_assistant.html#auto-discovery

Issue #2170, opened 2025-03-18
Major MQTT corruption when reading battery service/char


DigiH:
Fresh OMG 1.8.1 installation of `esp32dev-ble` binary on an ESP32-WROOM-32D board. 

**Describe the bug**
trying to read the battery levels of my Playbill Candles again, first trying through my openHAB automation, but then also separate by issuing the READ command(s) through MQTT Explorer. There is predominantly major MQTT message corruption, instead of receiving correct responses on the correct device topic. 

Trying to read the battery level with

**Command topic:**
home/OMG_ESP32_BLE/commands/MQTTtoBT

**JSON playload:**
```
{
	"ble_read_address": "AA:BB:CC:DD:EE:FF",
	"ble_read_service": "180f",
	"ble_read_char": "2a19",
	"value_type": "INT",
	"ttl": 4,
	"immediate": true
}
```

With such strange corrupt MQTT message results, on a topic "d", even though the response JSON does look fine apart from the also "id":"d" ;)

![Image](https://github.com/user-attachments/assets/c020b523-9ec1-43b1-8879-93f9f46702ec)

With the corruption sometimes even going as far as the whole top topic home/internal getting mangled into the general home/OMG_ESP32_BLE/BTtoMQTT list.

This screenshot also shows that this seems to occur mainly with successful responses, whereas failed responses - produced by OMG itself when not receiving any response from the device - look fine, but also only most of the time.

![Image](https://github.com/user-attachments/assets/f2fb38fc-08bc-4c12-a523-704faef0a93e)

This should be reproducible with any device which has the standard Bluetooth SIG Battery level 180f/2a19 service/char.

DigiH:
@h2zero @1technophile 


Can you reproduce this issue on your end? I am constantly getting it, also with WRITE commands now when trying them with a new device, and also still after updating with the latest https://github.com/1technophile/OpenMQTTGateway/pull/2177

DigiH:
Ok, testing it again with a default pre-built binary `esp32dev-ble` things are working fine as expected.

I suspect it is me using all manual definitions for `ESPWifiManualSetup=true` in my custom environment, but likely especially

`NetworkAdvancedSetup=true`

with all its network definitions. Just for complete info, I also have (un)defined

'-UMQTTsetMQTT'
'-UMQTT_HTTPS_FW_UPDATE'
'-DCHECK_OTA_UPDATE=false'

[...]

PR #2183, opened 2025-05-16
[BT] MQTT corruption fix

DigiH:
Storing a copy of the ID instead of a pointer to a local variable to avoid MQTT publishing id topic and id property corruption.

https://github.com/1technophile/OpenMQTTGateway/issues/2170#issuecomment-2885551791

## Checklist:

### DIFF
--- a/environments.ini
+++ b/environments.ini
@@ -188,7 +188,7 @@ custom_hardware = Theengs Plug
 platform = ${com.esp32_platform}
 board = esp32dev
 extra_scripts = ${com-esp32.extra_scripts}
-board_build.partitions = partitions/no_coredump.csv
+board_build.partitions = partitions/alltest_large.csv
 lib_deps =
   ${com-esp32.lib_deps}
   ${libraries.ble}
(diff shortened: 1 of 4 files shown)


DigiH:
@h2zero - could you have a look at all my changes?

@1technophile  - `esp32dev-all-test` failing again as the build is slightly over the, here theoretical only, limit

h2zero:
Looks good to me. I would suggest also adding to this a change of the partition table used by the all test build, single app (no ota) or something similar to pass the test build.

DigiH:
Not quite ready, as the id fix has now broken the white- and black-list functionality. So investigating the changed id comparison to the lists' definitions now to get things back in working order.

Any thoughts as where this might additionally also be affecting other MAC address comparisons?

DigiH:
@h2zero @1technophile - am a bit stuck here. Do you have any insight into the white-/black-list problem now? I think some of the connection MAC address comparisons might also fail, but don't have any of the devices to verify.

Do you think we should include these necessary changes in the PR or create a separate fix once this corruption fix has been merged?

DigiH:
LOL this is fine for merging!

I'm really pushing things it seems, first the MQTT publishing corruption leading to this, and then the white-list failure, which accidentally has nothing to do with this here, but is due to the fact that at the same time I got two new devices, but the white-list breaks if there are more than 28 MAC addresses defined in the white-list.

Not sure if I overlooked something in the docs and if this is a known limit, or if it is something to look into for addressing for possible larger white-lists - to be look into separately ;)

h2zero:
Lol glad you got that figured out, I didn't know the list was limited like that either.

PR #2207, opened 2025-06-25
[RF] fixups: RF2 trasnmit, Pilight rawenabled compile

peshovec:
## Description:
RF2 transmit (XtoRF2 with json payload) was not handling well cc1101 init - fixup in the same style as gatewayRF.
Pilight - when compiling with defined Pilight_rawEnabled - a check for the currentRecevier is made. Define the variable as defined externally.

## Checklist:

### DIFF
--- a/main/gatewayPilight.cpp
+++ b/main/gatewayPilight.cpp
@@ -44,7 +44,7 @@
 
 #  include <ESPiLight.h>
 ESPiLight rf(RF_EMITTER_GPIO); // use -1 to disable transmitter
-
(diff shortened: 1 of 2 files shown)


1technophile:
We could move the changes into `initCC1101` which already has the `#  ifdef ZradioCC1101`.
Also this would enable to factorize some lines of code.

1technophile:
This would require to move it from the other RF modules I suppose

1technophile:
Merging like this, we can do a separate PR to factorize it

PR #2220, opened 2025-08-06
[DISC] Fix cosmetic issue in Home Assistant where devices don't link to gateway

plambrechtsen:
## Description:

The `via_device` is set for each device when it is being created in Home Assistant. Currently this links to  the `gateway_name` which is incorrect so all the devices show up linking to an unknown entity.
In Home Assistant the OMG entity is named by it's MAC address, so this change now means that they are all correctly linked together. 

It will stop this from happening when looking in the MQTT Entities in Home Assistant. As the unnamed device is the missing `via_device` gateway_name.

image
and

| Before| After|
| - | - |
|image|image|

## Checklist:

### DIFF
--- a/main/mqttDiscovery.cpp
+++ b/main/mqttDiscovery.cpp
@@ -43,6 +43,7 @@
 
 extern bool ethConnected;
 extern JsonArray modules;
+String gateway_mac;
 
 char discovery_prefix[parameters_size + 1] = discovery_Prefix;
 // From https://github.com/home-assistant/core/blob/d7ac4bd65379e11461c7ce0893d3533d8d8b8cbf/homeassistant/const.py#L225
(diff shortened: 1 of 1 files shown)


main/mqttDiscovery.cpp
```diff
@@ -593,7 +595,7 @@ void createDiscovery(const char* sensor_type,
       }
     }
 
-    device["via_device"] = String(gateway_name); //device name of the board
+    device["via_device"] = String(gateway_mac); //mac address of the gateway so that the devices link to the gateway
```
1technophile: Could we simplify with : `device["via_device"] = String(getMacAddress());` ?

plambrechtsen: No as then you have the Mac Address of the Bluetooth sensor and not the gateway.

So you need to store it as a global variable that is referenced later.

I thought exactly the same thing before I tested it and found out that wasn't going to work.

plambrechtsen: I have just re-tested this, and `getMacAddress()` is working returning the OMG mac address and not the Bluetooth sensor mac address. It's weird as when I was testing this before I was getting the Bluetooth sensor mac address. Pushing an update.

1technophile:
Thanks

PR #2224, opened 2025-08-12
[DISC] Add RSSI as diagnostic entity_category for Bluetooth devices

plambrechtsen:
## Description:

It would be very useful to track the RSSI of Bluetooth devices as the data is sent by OMG but not displayed against the entity. As per @1technophile suggestion I have added it as a diagnostic entity_category.

image

## Checklist:


1technophile:
Thanks

PR #2230, opened 2025-08-15
[BT] Consistent checking for BLEDecryptor as a Boolean

DigiH:
Consistent checking for BLEDecryptor as a Boolean, not just as a Define, so that
```
…
#    define BLEDecryptor true //true if decrypt encrypted PVVX, BTHome v2 or Victron Energy broadcast data
…
```
also behaves correctly when set to **false**.

## Checklist:

### DIFF
--- a/main/User_config.h
+++ b/main/User_config.h
@@ -654,13 +654,13 @@ void storeSignalValue(uint64_t);
 
 #ifdef ZgatewayBT
 #  ifndef BLEDecryptor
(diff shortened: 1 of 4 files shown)


DigiH:
@1technophile @plambrechtsen 

Any comments on this one?

plambrechtsen:
Looks good to me. Should have thought about if vs ifdef.

PR #2252, opened 2025-11-27
[SYS] Feature/global filter and tests

Odyno:
## Description:
- [x] Add Test structure + Workflow CI
- [x] Abstraction Storage Logics
- [x] Extracted  Logs Tool, e Messages config for native tests
- [x] Introduced Filter logic + Docs + Tests
- [x] Update and manage filters via MQTT commands

*Test Documentation*:  https://github.com/Odyno/OpenMQTTGateway/tree/feature/global-filter-and-tests/test
*Filter Documentation*: https://github.com/Odyno/OpenMQTTGateway/blob/feature/global-filter-and-tests/docs/use/gateway.md#filter-messages 

## Checklist:
  - [x] The pull request is done against dev
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] I accept the [DCO](https://github.com/1technophile/OpenMQTTGateway/blob/development/docs/participate/development.md#developer-certificate-of-origin).

PR #2306, opened 2026-05-02
Add RTL_433 device whitelist support

waiet:
## Summary

This PR adds an optional whitelist for the RTL_433 gateway.

When enabled, only RTL_433 devices whose generated device identifier is present in the whitelist are published to MQTT. Non-whitelisted RTL_433 messages are ignored. The whitelist is persisted in preferences and can be managed from the WebUI RF configuration page or through MQTT SYS commands.

The Home Assistant MQTT discovery payload also exposes a gateway switch to enable or disable the RTL_433 whitelist.

## Details

- Adds persistent RTL_433 whitelist configuration
- Uses the RTL_433 device identifier format: `type/model/subtype/channel/id`
- Tracks seen RTL_433 devices during runtime so they can be selected from the WebUI
- Filters MQTT publishing and RTL_433 discovery when the whitelist is enabled
- Adds `RTL_433: Whitelist` switch to Home Assistant discovery
- Publishes whitelist state in the SYS state topic

## Testing

Built successfully with PlatformIO:

- `esp32dev-rtl_433`
- `lilygo-rtl_433`

Also tested on an ESP32 + CC1101 setup receiving 433 MHz weather sensor data.

### DIFF
--- a/docs/integrate/home_assistant.md
+++ b/docs/integrate/home_assistant.md
@@ -74,6 +74,8 @@ mqtt:
 
 Alternatively the rssi signal could be used also.
 
+The gateway device also exposes an `RTL_433: Whitelist` switch through MQTT discovery. When the whitelist switch is enabled, only matching RTL_433 devices are published to MQTT and discovered as Home Assistant entities. The whitelist entries can be edited from the OpenMQTTGateway WebUI RF configuration page or with MQTT commands.
+
 ### RF (RCSwitch based gateway) Auto discovery specificity
 With OpenMQTTGateway [configured to receive RF signals](../setitup/rf.html) messages are transmitted accordingly.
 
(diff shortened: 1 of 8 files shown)


Odyno:
I hope you're luckier :) I made a Merge Request https://github.com/1technophile/OpenMQTTGateway/pull/2252 to provide this functionality (if I understand correctly) globally as an evolution of another one I made even earlier for the RF gateway only... but unfortunately the bot passed and shut down before @1technophile   did the necessary testing... if you care keep the thread alive otherwise it will happen like my MR

1technophile:
Thanks for this, and sorry, RF filter wanted, so your instinct is
right. But it turns out @Odyno already solved this at a more general level in #2252:
a single message filter wired into the central publish path, so it filters RF (and
BLE, and everything else) by topic/pass/block in one place. It stalled because the
PR bundled too much and the stale bot closed it.

Because that filter already covers the RF case, I don't think a separate RF-only
whitelist should land on top of it. it'd duplicate a slice of #2252 and leave two
contracts to reconcile. I'm planning to take inspiration from #2252 in scoped pieces instead.

I don't want this to be a wasted effort on your side though. If you're interested,
the most useful contribution would be on the revived filter, RF-specific test
cases and docs especially, since you clearly know this path well. Happy to ping you
when the first scoped PR is up.

PR #2318, opened 2026-05-30
Fix esp32dev-multi_receiver build on ESP32-C3 (missing soc/sens_reg.h)

Schism86:
## Description:

`soc/sens_reg.h` is not present in the ESP32-C3 SDK (only available on ESP32, ESP32-S2, and ESP32-S3). The `esp32dev-multi_receiver` environment targets `esp32-c3-devkitm-1`, so the internal temperature reading code that includes this header fails to compile.

Adding `-DNO_INT_TEMP_READING` to the environment's `build_flags` skips that code path and restores a successful build.

## Checklist:

### DIFF
--- a/environments.ini
+++ b/environments.ini
@@ -1236,7 +1236,7 @@ custom_hardware = ESP32 LILYGO LoRa32 V2.1
 
 [env:esp32dev-multi_receiver]
 platform = espressif32@6.1.0
-board = esp32dev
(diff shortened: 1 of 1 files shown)


1technophile:
I suggest to create another environment dedicated for the c3 with multi receiver. If we apply the PR as is, the users with multi receiver on esp32 are going to loose compatibility (at least)

PR #2339, opened 2026-07-22
Add SX1262 FSK gateway for Heltec V4 weather sensors, with Home Assistant discovery

victorserbu2709:
## Description:
- `rtl_433_ESP` cannot drive the SX1262 radio on Heltec WiFi LoRa 32 V3/V4 boards, so Fine Offset/Ecowitt/LaCrosse "Group A" sensors (WH32, WH31E, WH40, WH57, WH51, WH45, WS80, WS90, TX35) were previously unreceivable there. This adds a standalone RadioLib-based gateway (`gatewayRTL_433SX1262.cpp`) that talks to the SX1262 in FSK mode directly and publishes readings on the existing RTL_433 MQTT topic, so downstream automations built around rtl_433 sensors keep working unchanged.
- Adds a new build environment, `heltec_v4-rtl_433sx1262` (`pio run -e heltec_v4-rtl_433sx1262 -t upload`).
- Wires the new gateway into `main.cpp`'s setup/loop and into the shared duplicate-signal-filtering logic alongside RF/IR/SRFB/WeatherStation/RTL_433.
- Adds Home Assistant MQTT discovery support for this gateway. It previously published sensor readings but never called `createDiscovery()`, since the real RTL_433 discovery machinery lives in `gatewayRTL_433.cpp` gated behind `ZgatewayRTL_433` (a flag this standalone gateway intentionally doesn't define). Adds a lightweight per-device discovery path directly in `gatewayRTL_433SX1262.cpp`, firing synchronously from the main loop since this gateway has no separate decode task to synchronize with.
- Renames the published `strike_distance` key to `strike_distance_km` to match its actual unit and the discovery template.

This was adapted by claude from https://github.com/iz0kew/WeatherStation2Meshtastic

### Test plan

- [x] `pio run -e heltec_v4-rtl_433sx1262` builds successfully
- [x] Flash to a Heltec WiFi LoRa 32 V4 and confirm WH32/WH31E/WH40/WH57/WH51/WH45/WS80/WS90/TX35 readings are received and published on the RTL_433 MQTT topic
- [x] Confirm Home Assistant auto-discovers each sensor's entities (temperature, humidity, pressure, rain, wind, UV/lux, lightning, PM2.5/PM10, CO2, soil moisture, battery) with correct units/device classes
- [x] Confirm no regressions on existing `ZgatewayRTL_433`/RF/IR/SRFB/WeatherStation duplicate-filtering behavior
## Checklist:

### DIFF
--- a/environments.ini
+++ b/environments.ini
@@ -1216,6 +1216,24 @@ build_flags =
 custom_description = For ESP32, Gateway using RTL_433_ESP and RadioLib with FSK modulation (beta)
 custom_hardware = ESP32 LILYGO T3-S3 SX1276 v1.0
 
+[env:heltec_v4-rtl_433sx1262]
+platform = espressif32@6.10.0
+board = heltec_wifi_lora_32_V3   ; V4 is pin-compatible; no dedicated board def exists yet
+board_build.partitions = min_spiffs.csv
+lib_deps =
+  ${com-esp32.lib_deps}
+  ${libraries.radiolib}
+build_flags =
+  ${com-esp32.build_flags}
+; *** OpenMQTTGateway Config ***
+  '-DvalueAsATopic=true'    ; MQTT topic includes model and device
+  '-DGateway_Name="OMG_heltec_v4_rtl_433_sx1262"'
+  '-DNO_INT_TEMP_READING=true'
+; *** OpenMQTTGateway Modules ***
+  '-DZgatewayRTL_433SX1262="RTL_433SX1262"'
+custom_description = Receives Fine Offset/Ecowitt/LaCrosse weather sensors via SX1262 FSK (RadioLib), publishes on the RTL_433 MQTT topic
+custom_hardware = ESP32-S3 Heltec WiFi LoRa 32 V4
+
 [env:lilygo-ble]
 platform = ${com.esp32_platform}
 platform_packages = ${com.esp32_platform_packages}
--- a/environments.ini
+++ b/environments.ini
@@ -1235,6 +1235,7 @@ build_flags =
 ; *** ssd1306 Display Options ***
   '-DZdisplaySSD1306="HELTEC_SSD1306"'
   '-DLOG_TO_OLED=true'          ; show log messages (wifi/mqtt/rf) on the OLED instead of JSON pages
+  '-DLOG_LEVEL=LOG_LEVEL_TRACE' ; compile in TRACE-level calls (raw packet/RSSI/hex-dump logging)
   '-DLOG_LEVEL_OLED=LOG_LEVEL_TRACE'
   '-DDISPLAY_IDLE_LOGO=false'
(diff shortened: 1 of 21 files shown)


victorserbu2709:
I tested with ecowitt ws90 (868 mhz)
image

1technophile (commented):
Thanks for this, the SX1262 gap is real and the packet mode looks like the only way to address it on these boards, as the SX126x has no direct mode to feed the pulse decoding of rtl_433_ESP.

A few points before merging:

`platformio.ini` line 65, `default_envs = heltec-rtl_433-fsk` is uncommented, it looks like a leftover of your local testing. Could you comment it back?

Could you add `heltec_v4-rtl_433sx1262` to https://github.com/1technophile/OpenMQTTGateway/blob/development/.github/workflows/environments.json ? Without it the environment is not built by the CI or the nightly, so nothing checks this code.

The models are published as `WS80`, `WH51`, `WH45`, while rtl_433 publishes `Fineoffset-WS80`, `Fineoffset-WH51`. As the environment sets `valueAsATopic=true` the topic differs also. So the existing automations and HA entities will not match, which is the compatibility you mention in the description. Could you align the model names?

About the naming, the module decodes 9 sensors on one frequency and one modulation, `RTL_433SX1262` suggests much more. I'm afraid we will get many issues from users expecting their 433Mhz devices to work with it. I would suggest to rename it to something like `SX1262`, keep the RTL_433 topic for HA, and list the supported sensors in `custom_description`.

Would you consider also adding a few captured frames per decoder as a test? Those parsers are pure functions on byte arrays, so they can be tested without hardware. Otherwise nobody without the physical sensor can review a change here later.

Minor, `lightningCountValid` is set in `wh57.cpp` but never read, so the unreliable startup strike counts are published anyway.

Thanks

1technophile:
One more point, about the license.

`main/rtl_433sx1262/LICENSE` declares the decoders as MIT, but they are a direct port of the rtl_433 decoders, which are GPL-2.0-or-later. In `ws90.cpp`:

```c
int wind_avg = ((b[7] & 0x10) << 4) | b[10];
int wind_dir = ((b[7] & 0x20) << 3) | b[11];
int wind_max = ((b[7] & 0x40) << 2) | b[12];
```

this is identical to `fineoffset_ws90.c`, variable names included, and it is the same in `wh51.cpp` and `wh45.cpp`. Your file headers reference those sources already. So I don't think WeatherStation2Meshtastic was in position to relicense them as MIT.

Nothing dramatic for us, rtl_433 is "either version 2 of the License, or (at your option) any later version", so it upgrades to GPLv3 and we can carry the code.

But could you remove the MIT LICENSE file and put the usual OpenMQTTGateway GPL header on the decoders instead, with the attribution to rtl_433 and WeatherStation2Meshtastic inline? We do the same in `gatewaySRFB.cpp`, which credits Xose Pérez directly in the header.

Thanks

Issue #2367, opened 2026-09-11
Buffer overflow in call to dtostrf()


mgerczuk:
**Describe the bug**
OMG crashes and reboots sporadically with logs like
```
2026-09-05 04:27:41 CORRUPT HEAP: Bad head at 0x3ffb9cf8. Expected 0xabba1234 got 0x3ffb9a34
2026-09-05 04:27:41 
2026-09-05 04:27:41 assert failed: multi_heap_free multi_heap_poisoning.c:253 (head != NULL)
```

**To Reproduce**
Happens when some device reports a temperature < -100°C. The ```char temp[5]``` in webUI.cpp in line 2078 is too small for ```dtostrf(-123.4, 3, 1, temp)``` which needs 7 bytes.

**Environment:**
* OpenMQTTGateway version used (V0.9.3, 0.8, development, e30b65ee6f4713d895a85690d222e036190da46b)

PR #2368, opened 2026-09-11
Replace dtostrf() with safer Arduino String constructor call

mgerczuk:
## Description:
Fix for #2367

## Checklist:

### DIFF
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -750,15 +750,11 @@ bool pubMQTT(String topic, String payload) {
 }
 
 bool pubMQTT(String topic, int payload) {
-  char val[12];
-  sprintf(val, "%d", payload);
-  return pubMQTT(topic.c_str(), val);
+  return pubMQTT(topic.c_str(), payload);
 }
 
 bool pubMQTT(String topic, unsigned long long payload) {
-  char val[21];
-  sprintf(val, "%llu", payload);
-  return pubMQTT(topic.c_str(), val);
+  return pubMQTT(topic.c_str(), payload);
 }
 
 bool pubMQTT(String topic, float payload) {
@@ -802,9 +798,7 @@ bool pubMQTT(const char* topic, double payload) {
 }
 
 bool pubMQTT(String topic, unsigned long payload) {
-  char val[11];
-  sprintf(val, "%lu", payload);
-  return pubMQTT(topic.c_str(), val);
+  return pubMQTT(topic.c_str(), payload);
 }
 
 void delayWithOTA(long waitMillis) {
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -762,15 +762,11 @@ bool pubMQTT(String topic, unsigned long long payload) {
 }
 
 bool pubMQTT(String topic, float payload) {
-  char val[12];
-  dtostrf(payload, 3, 1, val);
-  return pubMQTT(topic.c_str(), val);
+  return pubMQTT(topic.c_str(), String(payload, 1).c_str());
 }
 
 bool pubMQTT(const char* topic, float payload) {
-  char val[12];
-  dtostrf(payload, 3, 1, val);
-  return pubMQTT(topic, val);
+  return pubMQTT(topic, String(payload, 1).c_str());
 }
 
 bool pubMQTT(const char* topic, int payload) {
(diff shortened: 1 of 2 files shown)


1technophile:
Thanks for tracking this down—the undersized `dtostrf` buffers explain the failure reported in #2367.

I’d prefer bounded `snprintf` formatting here, mainly to avoid additional temporary allocations. The [ESP32 core](https://github.com/espressif/arduino-esp32/blob/3.3.9/cores/esp32/WString.cpp#L112) allocates `decimalPlaces + 42` bytes for `String(float, n)` and `decimalPlaces + 312` for `String(double, n)`. The [ESP8266 core](https://github.com/esp8266/Arduino/blob/3.1.2/cores/esp8266/WString.cpp#L105) uses a fixed 33-byte stack buffer with unbounded `dtostrf`, so the constructor can still overflow for very large values.

The numeric formatting in the `pubMQTT` overloads previously used stack buffers. We can retain that approach while bounding writes and checking for truncation so we don’t publish a misleading numeric value:

```cpp
bool pubMQTT(String topic, float payload) {
  return pubMQTT(topic.c_str(), payload);
}

bool pubMQTT(const char* topic, float payload) {
  char val[16];
  int n = snprintf(val, sizeof(val), "%.1f", payload);
  if (n < 0 || static_cast<size_t>(n) >= sizeof(val)) {
    return false;
  }
  return pubMQTT(topic, val);
}
```

The nearby `double` overload has the same issue (`sprintf("%f")` into 16 bytes overflows at positive `1e8`, and earlier for negative values). The `long` overload’s `char val[11]` is also one byte short for a 32-bit `LONG_MIN`. Could we fix those too?

```cpp
bool pubMQTT(const char* topic, long payload) {
  char val[12];
  int n = snprintf(val, sizeof(val), "%ld", payload);
  if (n < 0 || static_cast<size_t>(n) >= sizeof(val)) {
    return false;
  }
  return pubMQTT(topic, val);
}

bool pubMQTT(const char* topic, double payload) {
  char val[24];
  int n = snprintf(val, sizeof(val), "%f", payload);
  if (n < 0 || static_cast<size_t>(n) >= sizeof(val)) {
    return false;
  }
  return pubMQTT(topic, val);
}
```

For `webUI.cpp`, the same bounded formatting would work at each site. The `(String)temp` concatenation already existed, so this preserves the existing `String` usage without adding the numeric constructor’s temporary allocation. For example:

```cpp
            if (data.containsKey("temperature_C")) {
              float temperature_C = data["temperature_C"];
              char temp[16];

              snprintf(temp, sizeof(temp), "%.1f",
                       displayMetric ? temperature_C : convertTemp_CtoF(temperature_C));
              line3 = "temp: " + (String)temp + (displayMetric ? "°C " : "°F ");
            }

            float humidity = data["humidity"];
            if (data.containsKey("humidity") && humidity <= 100 && humidity >= 0) {
              char hum[16];
              snprintf(hum, sizeof(hum), "%.1f", humidity);
              line3 += "hum: " + (String)hum + "% ";
            }
```

The gravity site would use `"%.3f"`. These buffers cover normal sensor readings and prevent overflow for larger values, although out-of-range display values may be truncated.

The same function also has undersized `itoa()` buffers (`pres[4]`, `co2[4]`, `pm25[3]`, `pm10[3]`, `bpm[3]`, `imp[3]`, `lux[5]`, `steps[5]`, etc.). A pressure reading of `1013` alone requires five bytes including the terminator, so it already overflows `pres[4]`. Could you convert those in the same PR? For example:

```cpp
                int presint = data["pres"];
                char pres[12];
                snprintf(pres, sizeof(pres), "%d", presint);
                properties[property] = "pres: " + (String)pres + "hPa ";
```

This preserves the intended decimal precision for normal readings while bounding writes. I wouldn’t expect byte-for-byte equivalence for every value, since `dtostrf` and `printf` can differ at rounding boundaries and in special-value formatting.

1technophile:
Thanks, the updated approach looks good. A few integer conversions were missed in `webUIPubPrint()`:

- BODY: `steps[5]`, `actbpm[3]`, and `bpm[3]`.
- SCALE: `imp[3]`.
- UNIQ: the remaining `qual[3]` and `batt[5]` conversions.

For example, 10,000 steps needs six bytes including the terminator, and 100 bpm or impedance 100 needs four bytes, so those buffers can still overflow.

Could you switch these remaining sites to `intToString()` too? We can handle the broader property-array bounds checks and input validation in a separate follow-up PR.

1technophile:
Thanks

PR #2373, opened 2026-09-19
[BLE] Decode BM2 voltage and SoC from its encrypted advertisement

marklane-develop:
## Description:
The BM2 battery monitor broadcasts two advertisements: the iBeacon that Theengs Decoder already decodes for `batt`, and a second 16-byte manufacturer-data payload encrypted with AES-128-CBC using the same key as the existing `BM2_connect` handler (zero IV). Decrypted, it contains:

| bytes | content |
|---|---|
| 0–5 | device MAC, byte-reversed |
| 6–7 | voltage, big-endian, 1/100 V |
| 8 | state of charge, % |
| 9–15 | unknown |

This PR decrypts that advertisement in `process_bledata()` and publishes `volt` and `batt` in the normal `BTtoMQTT/<MAC>` message with the real RSSI, so BM2 voltage no longer needs a BLE connection. That matters on weak links, where adverts are received but a GATT connection fails. The existing BM2 HA discovery (`jsonVoltBM`, `jsonBatt`) works unchanged.

**Where the layout comes from:** the key is the one already in `BM2_connect` (from the community's BM2 reverse-engineering). The advert's plaintext layout was determined empirically - trial decryption of captured adverts, keeping only fields that matched independent ground truth on four devices: bytes 0-5 equal the advertiser's reversed MAC on every device; bytes 6-7 matched the voltage read over a BLE connection (exact on two devices, 0.01 V on a third); byte 8 is the same state of charge the plaintext iBeacon carries - over 53,521 adjacent iBeacon/encrypted packet pairs since, the two agree exactly 99.9 % of the time and otherwise differ by 1 as the value ticks over, across readings from 38 % to 100 %. Bytes 9-15 remain unknown (11, 13, 14 and 15 were constant across all four devices).

**Safety:** the advertisement has no plaintext marker, so a payload is accepted only if the decrypted bytes 0–5 equal the advertiser's own reversed MAC, plus a 5–20 V / ≤100 % sanity check. Any other device emitting 16 bytes of manufacturer data is rejected instead of producing a plausible false reading.

**Changes**
- `gatewayBLEConnect.cpp/.h`: BM2 key hoisted to `BM2_AES_KEY` (as for BM6) and the AES-CBC call factored into `BM2_decryptBlock()`, now shared by the connect handler and the advert decoder. No behaviour change to the connect path.
- `gatewayBT.cpp`: `decodeBM2Advert()` and its hook after the Theengs decode, next to the BLEDecryptor stage; `hexToBytes`/`reverseBytes` made available when `BLEDecoder` is set.
- `config_BT.h`: `BLEdevice.advData`. Once a device has produced a verified advertisement it is no longer flagged for periodic connection. A BM2 that never sends the encrypted advert keeps the existing connect behaviour.

**Behaviour change to note:** a BM2 with `advData` set has `connect == false`, so it becomes evictable from the device table when it is full, which it previously never was. It is re-learned on its next advertisement.

**Testing:** two ESP32 (`esp32dev-ble`) gateways and four BM2 devices, about 45 h on each gateway so far. One gateway is in range of the BM2s and has decoded 56,034 encrypted adverts with none failing; decoded values were compared message-for-message against an independent Node-RED AES implementation (505/505 identical). With default `bleconnect: true`, about 33 connect cycles over 8 h produced zero BM2 connections. Heap floor (~20 kB) and the task stack high-water marks are unchanged from the base commit. Builds clean with no warnings on `esp32dev-ble`, `esp32dev-ble-mqtt-undecoded` (`BLEDecoder=false`), `esp32c3-dev-m1-ble` and `lilygo-ble`; `clang-format` clean.

Each gateway has had one unexplained, non-graceful reboot on this build (after 12.5 h and 24.3 h - nothing published on `LOGtoMQTT`, so not an `ESPRestart()`). The unpatched base commit `4e49049b` did the same on the same hardware in the two days before (after 3 h and 32 h), whereas v1.8.1 had run 292 h, so this looks like the development base rather than this change - but I have no serial log to prove it.

**Not covered:** BM6 (different key and payload; its advert is not decoded here). RSSI is not yet exposed as a discovery sensor for BM2. That is a possible follow-up, as BM2 sits in the hand-written discovery group that has no RSSI entity.

## Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code)


marklane-develop:
Updated the description: the Testing section now has the full figures (56,034 encrypted adverts decoded over about 45 h, none failing) and notes one unexplained reboot per gateway on this build — the unpatched base commit did the same on the same hardware, v1.8.1 ran 292 h, and heap floor and stack high-water marks are unchanged by this patch. Also added how the advert layout was determined. The one failing check (`lilygo-ble`) is a PlatformIO package download that was reset mid-transfer, not a compile error; that environment builds cleanly locally.

